PR #6, opened 2020-10-29
EREGCSC-68 Display section name in header

whytheplatypus:
Display the name of the section currently being read in a place that remains visible throughout the user experience.

### DIFF
--- a/config/regulations-site/local_settings.py
+++ b/config/regulations-site/local_settings.py
@@ -43,6 +43,7 @@ TEMPLATES = [
         },
         "DIRS": [
             "%s/templates" % BASE_DIR,
+            "/var/lib/eregs/cmcs/templates",
             "/var/lib/eregs/cmcs/regulations/sidebar/content",
         ],
     },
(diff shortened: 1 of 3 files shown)


whytheplatypus:
Example of a change to the `regulations-site` javascript that would update the `active-title` with the full section title text.

```
diff --git a/regulations/static/regulations/js/source/views/header/sub-head-view.js b/regulations/static/regulations/js/source/views/header/sub-head-view.js
index fa50f124..249d7c0d 100644
--- a/regulations/static/regulations/js/source/views/header/sub-head-view.js
+++ b/regulations/static/regulations/js/source/views/header/sub-head-view.js
@@ -28,7 +28,11 @@ const SubHeadView = Backbone.View.extend({
     // populates subhead (.header-label) with new title depending on viewport location
     // ex: §478.1(a) to §478.1(b)
   changeTitle: function changeTitle(label) {
-    this.$activeTitle.html(label);
+    let title = $(".section-title").text();
+    if (title === "") {
+      title = label;
+    }
+    this.$activeTitle.html(title);
   },
 
   displayCount: function displayCount(resultCount) {
```

whytheplatypus:
This change fails to keep the title consistently up to date with the section title as navigation and rendering is handled by both the server and the client.

PR #17, opened 2020-12-07
Lambda function to rebuild search index

whytheplatypus:
The search functionality of eregs requires that the `rebuild_pgsql_index` be run when data is added.


whytheplatypus:
tested with a manual deploy

PR #27, opened 2021-01-08
EREGCSC-138

theabrad:


config/static-assets/static/regulations/js/source/views/drawer/drawer-view.js
```diff
@@ -13,6 +13,21 @@ Backbone.$ = $;
...
   el: '#menu',
 
+  events: {
+    'click .section-subpart': 'toggleSection',
+    'click .section-subjgrp': 'toggleSection',
+  },
+
+  toggleSection: function toggleSection(e) {
+    const $target = $(e.currentTarget);
+    const $sectionClass = `.${$target.attr('class').split(' ')[0]}`;
```
whytheplatypus, line 23: is there a way to remove the requirement of the right thing being in the first slot. Maybe using a custom `data-` attribute?

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #29, opened 2021-01-15
Allow for sections to be orphans in the toc

whytheplatypus:
Before this orphaned sections wouldn't display their names etc. this
fixes that edge case.

image


whytheplatypus:
With this change:
image

theabrad (approved):
Pulled it in locally and took a look. Looks good to me.

PR #39, opened 2021-03-01
EREGSCSC-240 guidance pipeline

theabrad:


guidance_pipeline/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+type Guidance struct {
+	header string
```
whytheplatypus, line 11: It looks like this isn't used and can be removed

guidance_pipeline/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"bytes"
+	"flag"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+)
+
+var update = flag.Bool("update", false, "update .golden files")
```
whytheplatypus, line 11: 👍

guidance_pipeline/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+func TestMain(t *testing.T) {
```
whytheplatypus, line 13: [TestMain](https://golang.org/pkg/testing/#hdr-Main) is sort of reserved. Not sure what would be better.

guidance_pipeline/to_json.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import (
+	"encoding/json"
+)
+
+type Link struct {
```
whytheplatypus, line 7: Can use the Guidance struct instead, if the json tags are moved to it.

guidance_pipeline/read_data.go
```diff
@@ -0,0 +1,31 @@
...
+	"encoding/csv"
+	"os"
+)
+
+func readData(file string) ([][]string, error) {
+
+	guidanceFile, err := os.Open(file)
+
+	if err != nil {
+		return [][]string{}, err
```
whytheplatypus, line 13: can be `return nil, err`

guidance_pipeline/read_data.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	defer guidanceFile.Close()
+
+	reader := csv.NewReader(guidanceFile)
+
+	// First line skip
+	if _, err := reader.Read(); err != nil {
+		return [][]string{}, err
```
whytheplatypus, line 21: can be `return nil, err`

guidance_pipeline/read_data.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	records, err := reader.ReadAll()
+
+	if err != nil {
+		return [][]string{}, err
```
whytheplatypus, line 27: can be `return nil, err`

guidance_pipeline/to_json.go
```diff
@@ -0,0 +1,57 @@
...
+	return regulation
+}
+
+func toJSON(file []byte, header string, guidances []Guidance) ([]byte, error) {
+	var regulations []Regulation
+
+	if len(file) > 0 {
+		err := json.Unmarshal(file, &regulations)
+		if err != nil {
+			return []byte{}, err
```
whytheplatypus, line 43: can be `return nil, err`

guidance_pipeline/to_json.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+
+	newReg := buildRegulation(header, guidances)
+	regulations = append(regulations, newReg)
+
+	regsJSON, err := json.MarshalIndent(regulations, "", " ")
+
+	if err != nil {
+		return []byte{}, err
```
whytheplatypus, line 53: can be `return nil, err`

guidance_pipeline/read_data.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	"encoding/csv"
+	"os"
+)
+
+func readData(file string) ([][]string, error) {
```
whytheplatypus, line 8: Update to use a `Reader` interface and pass the file in. That will let opening the file bubble up to main and make it clear what the resources being used are.

guidance_pipeline/read_data.go
```diff
@@ -0,0 +1,31 @@
...
+
+	if err != nil {
+		return [][]string{}, err
+	}
+	defer guidanceFile.Close()
+
+	reader := csv.NewReader(guidanceFile)
+
+	// First line skip
+	if _, err := reader.Read(); err != nil {
```
whytheplatypus, line 20: This is to skip the header line?

guidance_pipeline/write_data.go
```diff
@@ -0,0 +1,19 @@
...
+
+func writeData(filename string, data []byte) error {
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return err
+	}
+	if _, err := f.Write(data); err != nil {
+		return err
+	}
+	if err := f.Close(); err != nil {
```
whytheplatypus, line 15: Can use `defer f.Close()` right after checking for an error after `OpenFile`

guidance_pipeline/write_data.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+	"os"
+)
+
+func writeData(filename string, data []byte) error {
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_WRONLY, 0644)
```
whytheplatypus, line 8: Could be moved into main, if it was doing more I'd say leave the function but make it accept a `Writer` interface.

whytheplatypus (changes requested):
Really good, it clearly expresses the transform that's happening.

whytheplatypus (approved):

PR #46, opened 2021-03-11
EREGCSC-336 guidance download

theabrad:


guidance_pipeline/main.go
```diff
@@ -13,22 +16,84 @@ type Guidance struct {
...
+
 func main() {
-	if len(os.Args) != 2 {
-		fmt.Println("Usage:", "./guidance", "FILE")
+	flag.Usage = func(){
+		fmt.Fprintf(os.Stderr, "Usage: %s [options] [file]\n", os.Args[0])
+		flag.PrintDefaults()
+	}
+	flag.Parse()
+	if flag.NFlag() == 0 {
```
whytheplatypus, line 30: Needs a way to check explicitly for `file`, `directory`, or `url`. This passes if only `-o` is provided.

guidance_pipeline/main.go
```diff
@@ -65,7 +130,7 @@ func makeMapOfRegs(header string, records [][]string) map[string][]Guidance {
 
 func writeRegsToFile(header string, regs map[string][]Guidance) error {
 	for key, reg := range regs {
-		filename := formatFilename(key)
+		filename := formatFilename(*outputDirectory, key)
```
whytheplatypus, line 133: Give more context in the error states by adding things like this filename. Things like `fmt.Errorf("%s %s", filename, err)` work or a special Error type if it makes sense.

whytheplatypus (changes requested):
looking really good. I tried it out locally and the new guidance loaded and looks great. Couple ideas on making the errors more helpful.

theabrad:
check if valid URL in the link section and log if not.

guidance_pipeline/main.go
```diff
@@ -52,7 +133,7 @@ func makeMapOfRegs(header string, records [][]string) map[string][]Guidance {
 			for _, reg := range regs {
 				guidance := Guidance{
 					Name: record[0],
-					Link: record[1],
+					Link: validURL(header, record[1]),
```
whytheplatypus, line 136: To prevent the user seeing broken links this should return an error and not get added to `guidance` if it is an invalid url.

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #55, opened 2021-03-29
EREGCSC-353 Cypress

theabrad:


whytheplatypus (approved):

docker-compose.test.yml
```diff
@@ -0,0 +1,86 @@
...
+    links:
+      - regulations-core
+  static-assets:
+    image: nginx
+    ports:
+      - 8888:80
+    volumes:
+      - ./config/static-assets/nginx_fonts.conf:/etc/nginx/conf.d/default.conf
+      - ./regulations-site/regulations/static/regulations:/usr/share/nginx/html
+  e2e:
```
whytheplatypus, line 70: Another option might be to use `localhost:8000` for the `Cypress_baseUrl` in the original docker-compose.yml but change it's network_mode to host

```
e2e:
    network_mode: host
```

PR #76, opened 2021-05-25
EREGCSC-626 Remove references to v1 in regulations-site

cgodwin1:


whytheplatypus (changes requested):
it looks like there are some references to the `sidebars` folder that need to also be cleaned up https://github.com/CMSgov/cmcs-eregulations/blob/master/serverless/regulations-site/cmcs/regulations/sidebars

whytheplatypus (approved):

PR #91, opened 2021-07-12
Eregs consolidation

whytheplatypus:


regcore/search/models.py
```diff
@@ -1,11 +1,41 @@
...
+            .annotate(
+                headline=SearchHeadline(
+                    "content",
+                    SearchQuery(query),
+                    start_sel='<span class="search-highlight">',
+                    stop_sel='</span>',
+                ),
+            )\
+            .order_by('-rank')\
+            .values("type", "content", "headline", "label", "parent", "part__document__title", "part__title", "part__date")
```
whytheplatypus, line 32: Can get rid of the `values` call from this, it is useful specifically for the API endpoint and should be clear in it's use.

regcore/search/models.py
```diff
@@ -1,11 +1,41 @@
...
+    SearchQuery,
+    SearchRank,
+    SearchHeadline,
+)
 
 from regcore.models import Part
 
 
+class SearchIndexQuerySet(models.QuerySet):
+    def search(self, query):
```
whytheplatypus, line 18: There are two responsibilities in this function:
1. Query documents for a string in a consistent way
2. Limit that Query to only effective parts.

Separate the function into `effective()` and `search()` then we can say something like: `SearchIndex.objects.effective(date).search(query)`

regcore/models.py
```diff
@@ -5,6 +5,15 @@ class PartQuerySet(models.QuerySet):
     def effective(self, date):
         return self.filter(date__lte=date).order_by("name", "-date").distinct("name")
 
+    def effective_title(self, date, title):
+        return self.filter(title=title).filter(date__lte=date).order_by("name", "-date").distinct("name")
+
+    def full_part(self, date, title, part):
+        return self.filter(title=title).filter(date__lte=date).filter(name=part).latest("date")
```
whytheplatypus, line 12: These can be brought up for clarity: `Part.objects.effective(date).filter(title=title, part=part)`

regcore/models.py
```diff
@@ -5,6 +5,15 @@ class PartQuerySet(models.QuerySet):
...
+
+    def versions(self, title, part):
+        return self.filter(name=part).filter(title=title).order_by('-date')
```
whytheplatypus, line 15: Can be more specific here (abstractions should always hide something). And fetch the value specifically so it's not pulling down everything. I think `values('date')` will work but I'm not sure.

regcore/fields.py
```diff
@@ -0,0 +1,49 @@
+import base64
```
whytheplatypus, line 1: I thought we got rid of this?

cmcs_regulations/settings.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Build paths inside the project like this: BASE_DIR / 'subdir'.
+BASE_DIR = Path(__file__).resolve().parent.parent
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/3.2/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = os.environ.get("SECRET_KEY", 'django-insecure-u!&%t$qxa23zn1f*-+4pngd(p=nl_m3()+v839+fa=06y9(*)n')
```
whytheplatypus, line 24: lets remove the default.

regulations/apps.py
```diff
@@ -0,0 +1,46 @@
...
+        regulation tree. Rather than have the rendering code inspect which
+        templates are available at render time, we'll find the special cases
+        during the application startup"""
+        cls._find_templates('custom_nodes', cls.custom_tpls)
+
+    @classmethod
+    def precompute_node_type_templates(cls):
+        """Different node types may display differently. Scan the provided
+        templates to see if any node-type-specific templates are present."""
+        cls._find_templates('tree', cls.node_type_tpls)
```
whytheplatypus, line 46: can remove this

PR #95, opened 2021-07-19
EREGCSC-742 Add supplementary content read API

cgodwin1:


supplementary_content/views.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def make_category_tree(data):
+    tree = {}
+    for category in data:
+        stack = []
+        current = category
+        while current['parent'] is not None:
+            stack.append(current['parent'])
+            current = current['parent']
```
whytheplatypus, line 96: Pull this out into a function that returns the stack, this will help the scoping of the `current` variable.

whytheplatypus: Include `category` in the stack. `stack = [category]`. This allows you to leave off the if statement on line 103.

supplementary_content/views.py
```diff
@@ -0,0 +1,107 @@
...
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(make_category_tree(serializer.data))
+
+
+def add_category(category):
+    return {
+        'title': category['title'],
+        'description': category['description'],
+        'supplementary_content': [],
+        'sub_categories': {},
```
whytheplatypus, line 77: Use a `get` with a default to allow for existing values here, should let the for loop on lines 105-106 go away.

`'supplementary_content': category.get('supplementary_content', []),`

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #99, opened 2021-07-21
Tune regulations search behavior

whytheplatypus:
- correctly use rank
- rank by results from different parts of the content structure
- limit results to only those matching to a certain degree
- construct a match for citations of the style <part>.<section>


regulations/templates/regulations/search.html
```diff
@@ -41,6 +41,7 @@
                         {% for result in results %}
                             {% if result.label|length > 1 %}
                             <div class="result">
+                                <!--{{ result.rank }}-->
```
PhilR8, line 44: should this be here?

whytheplatypus: I was thinking it could be a nice way for us to debug / monitor the search behavior? It shows up if you view the source of the page

cgodwin1 (approved):

PR #108, opened 2021-07-22
EREGCSC-808 Fix active supbart collapse

whytheplatypus:


PhilR8 (approved):

regulations/static/regulations/css/scss/partials/_sidebar_left.scss
```diff
@@ -11,6 +11,10 @@ $toc_toggles_width: 64px;
 // Sidebar
 
 aside.left-sidebar {
+    .invisible {
+        height: 0px !important;
```
whytheplatypus, line 15: Check and see if the `!important` can be removed

whytheplatypus: nope, it's needed to override the inline styles used by the vue component.

PR #110, opened 2021-07-26
Eregcsc 818 tiny tasks

PhilR8:


whytheplatypus (approved):

regulations/static/regulations/css/scss/partials/_reg_text.scss
```diff
@@ -18,13 +16,44 @@
...
+    }
+
+    .part-title {
+        margin: $spacer-5 0 $spacer-3;
     }
 
     h2 {
-       color: $black;
-       font-size: $base-font-size * 1.2;
+        color: $black;
```
inkandsound, line 55: We're trying to avoid calling colors by names directly - in `application_settings.scss` there should be an abstracted variable we can call (i.e. `$color_primary`or similar rather than `$teal`) or you can add a new variable there and call it.

PhilR8: `$black` has been changed to `$title_color` and updated where used.

PhilR8 (commented):
Comment resolved.

PR #111, opened 2021-07-27
Eregcsc 796 left sidebar tabbing

PhilR8:


regulations/static/components/CollapseButton.vue
```diff
@@ -1,17 +1,19 @@
...
-        </button>
-    </div>
+    <button
+        v-bind:class="{ visible: visible }"
+        v-on:click="click"
+        aria-label="expand or collapse a subpart"
+    >
+        <slot name="expanded" v-if="visible">Hide</slot>
+        <slot name="collapsed" v-if="!visible">Show</slot>
+    </button>
```
PhilR8, line 9: removed containing div because it was being focused using keyboard navigation.  Button is also now full height and can be tapped/clicked anywhere in that height to toggle collapse

regulations/static/components/Collapsible.vue
```diff
@@ -16,6 +16,7 @@ export default {
         requestAnimationFrame(() => {
             this.visible = this.state === "expanded";
             this.isVertical = this.direction === "vertical";
+            this.setTabIndex(this.childtag, this.visible);
         });
```
PhilR8, line 20: Using JS to set tabindex to 0/-1 based on `this.visible`.  I know this component is supposed to be as generic as possible but thought this would be an issue for any content that is a child of this component.  My semi-generic solution was to add an optional prop named `childtag` that can be passed in to component and used to determine which child elements will have tabindex toggled on expand/collapse.

regulations/static/components/CollapseButton.vue
PhilR8: Also added `data-test` data attribute here to better target elements for e2e tests; `data-test` is one of the options recommended in Cypress Best Practices here: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

regulations/static/components/Collapsible.vue
PhilR8: Per @whytheplatypus 's review, refactored implementation to add/remove `display: none` to collapse content instead of digging into content and adding/removing tabindexes.  Much cleaner.

whytheplatypus (approved):

PR #119, opened 2021-08-02
lambda to create a new superuser

theabrad:


serverless/createsu.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+import os
+from django.contrib.auth.models import User
+
+
+def handler(self, *args, **options):
+    if not User.objects.filter(username='root').exists():
```
whytheplatypus, line 7: Just for extra paranoia would you make both username and password configurable by env var and remove the default for the password.

serverless/serverless-experimental.yml
```diff
@@ -48,6 +48,9 @@ functions:
   reg_core_migrate:
     handler: migrate.handler
     timeout: 15
+  create_su:
```
whytheplatypus, line 51: Need to add the env vars for the username / password. Can probably reuse the same ssm vars used for the `HTTP_AUTH` env vars `${ssm:/eregulatinos/http/user~true}` `${ssm:/eregulations/http/password~true}`

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #126, opened 2021-08-05
Patch part view for supp content

whytheplatypus:
Before this the part view wouldn't render if the `sections` value was blank.


regulations/templates/regulations/partials/sidebar_right.html
```diff
@@ -9,11 +9,6 @@ <h2>Part {{ reg_part }} Final Rules</h2>
 
     <h2>Subpart Resources</h2>
 
-    <supplementary-content api_url="{% url 'regcore' %}" title="{{ title }}" part="{{ reg_part }}" :sections="{{ sections }}" />
+    <supplementary-content api_url="{% url 'regcore' %}" title="{{ title }}" part="{{ reg_part }}" {% if sections %}:sections="{{ sections }}"{% endif %}></supplementary-content>
```
cgodwin1, line 12: This causes the component to crash because sections is a required property. Also, the header "Subpart Resources" still appears although it is blank. Maybe `{% if sections %}` before the `h2`?

cgodwin1 (changes requested):

regulations/templates/regulations/partials/sidebar_right.html
cgodwin1: Additionally, if there are no sections, the limit of 2 for the final rules component makes less sense since we aren't as height-constrained.

whytheplatypus: good catch, i'd forgotten to include the updates to the component. But the conditional around the whole h2 makes more sense.

I'll leave changing the limit for a later ticket as we see what new styles and decisions need to be made based on things being collapsible etc.

cgodwin1 (approved):

PR #130, opened 2021-08-06
Add favicons and HTML

inkandsound:


inkandsound:
@PhilR8 I have everything set up for the favicon, but it'd be good to point to it in `root` as well (i.e. `regulations-pilot.cms.gov/favicon.ico`), which looks like it requires an update to `cmcs_regulations.urls`. Could you take a look? There are two `settings.py` files that seem to reference this so I'm not sure where the update should go.

The file is in static: `/images/favicon/favicon.ico`.

Thanks!

theabrad:
> @PhilR8 I have everything set up for the favicon, but it'd be good to point to it in `root` as well (i.e. `regulations-pilot.cms.gov/favicon.ico`), which looks like it requires an update to `cmcs_regulations.urls`. Could you take a look? There are two `settings.py` files that seem to reference this so I'm not sure where the update should go.
> 
> The file is in static: `/images/favicon/favicon.ico`.
> 
> Thanks!

do you mean something like this https://stackoverflow.com/a/21938270 @inkandsound ?

whytheplatypus (approved):

theabrad (approved):

cgodwin1 (approved):

PR #134, opened 2021-08-11
Eregcsc 830 jaws arrival anchor

PhilR8:


regulations/templates/regulations/partials/node_types/SECTION.html
```diff
@@ -2,10 +2,13 @@
     Used once per section. Contained by subparts at the top level.
 {% endcomment %}
 
-<section id="{{node.label | join:'-'}}" class="reg-section">
-    <h2 class="section-title">
-    <a class="reference" href="#{{node.label | join:'-'}}"><i class="fa fa-link"></i></a>
-    {{node.title}} </h2>
+<section aria-labelledby="section-title" tabindex="-1" id="{{node.label | join:'-'}}" class="reg-section">
+    <h2 class="section-title" id="section-title">
```
whytheplatypus, line 6: will this result in multiple elements having the id `section-title`?

whytheplatypus (changes requested):

regulations/templates/regulations/partials/node_types/SECTION.html
PhilR8: the section would have an `aria-label` that matched the `id` of the child `h2`.  I'm emulating example 2 here:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

whytheplatypus: the problem I'm seeing is that there will be multiple `h2`s on the page with `id="section-title"`. Should be able to change it to something like `id="section-title-{{node.label | join:'-'}}"`

PhilR8: Ah, I see what you're saying.  I'll make that change.

whytheplatypus (approved):

PR #135, opened 2021-08-11
EREGCSC-743 load supplementary content

theabrad:


whytheplatypus:
TODO:
- [ ] add a approved boolean to the model, default true, api should filter for only approved supp content (allows these to be loaded without them showing up just yet)

.github/workflows/load-supp-content.yml
```diff
@@ -0,0 +1,29 @@
...
+      name: ${{ matrix.environment }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: false
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.16' # The Go version to download (if necessary) and use.
+      - run: make supplementary_content
```
whytheplatypus, line 22: This won't populate the `guidance` folder in a place where the serverless function can reach it. It will build it in the github action, not in aws.

Makefile
```diff
@@ -68,6 +68,15 @@ data.local: export EREGS_PASSWORD=UkOAsfkItN
 data.%: tools/ecfr-parser/build/ecfr-parser
 	./tools/ecfr-parser/build/ecfr-parser -title 42 -subchapter IV-C -parts 400,457,460 -eregs-url $(CORE_URL)
 
+tools/guidance_pipeline/build/guidance_pipeline: tools/guidance_pipeline/*.go
+	cd tools/guidance_pipeline; go build -o build/ .
+
+tools/guidance_pipeline/guidance/*.json: tools/guidance_pipeline/build/guidance_pipeline
```
whytheplatypus, line 74: the tool wont work if the folder doesn't exist yet. The makefile could take care of creating the directory if it doesn't exist.

supplementary_content/views.py
```diff
@@ -46,19 +47,36 @@ class Meta:
...
+                    Prefetch(
+                        'sections',
+                        queryset=RegulationSection.objects.filter(
+                            title=title,
+                            part=part,
+                            section__in=section_list,
+                        )
+                    )
+                ).distinct(),
+            )
```
whytheplatypus, line 67: contain with something like `with_supplementary_content_for`.. its long but better than this.

supplementary_content/views.py
```diff
@@ -46,19 +47,36 @@ class Meta:
...
-        self.part = self.kwargs.get("part")
-        query = Category.objects.filter(
-            supplementary_content__sections__title=self.title,
-            supplementary_content__sections__part=self.part,
-            supplementary_content__sections__section__in=self.section_list,
-        ).distinct()
+        section_list = self.request.GET.getlist("sections")
+        title = self.kwargs.get("title")
+        part = self.kwargs.get("part")
+        query = Category.objects.prefetch_related(
```
whytheplatypus, line 54: I'm questioning the decision to start with `Category` at the top, given that the category tree has to be built anyway.

supplementary_content/views.py
```diff
@@ -107,20 +123,6 @@ def _get_category(tree, id):
...
-    return False
-
-
-def _filter_content(tree, title, part, sections):
-    for category in tree:
-        new_content = [item for item in category['supplementary_content'] if _matches_sections(item, title, part, sections)]
-        category['supplementary_content'] = new_content
-        _filter_content(category['sub_categories'], title, part, sections)
-
-
```
whytheplatypus, line 123: The structure required these duplicate checks. The hope is that the cognitive complexity can be reduced by only checking for the filters in one way.

whytheplatypus (dismissed):

PR #148, opened 2021-08-19
EREGCSC-640&609 Implement formulas, tables, images, and appendices

cgodwin1:


regulations/templatetags/paragraphs.py
```diff
@@ -6,7 +6,10 @@
 
 @register.filter(name='pdepth')
 def pdepth(value):
-    depth = len(value.get("label", []) or []) - section_depth
-    if len(value.get("marker", []) or []) > 1:
-        depth = depth - (len(value.get("marker", []) or []) - 1)
-    return depth
+    if value.get("parent_type", "") == "appendix":
+        return 1
```
whytheplatypus, line 10: This probably makes more sense as a default. What can be cleaned up in this function to basically say if there's no label etc.. return 1? I expect there's a bunch of unnecessary complexity present here.

whytheplatypus (changes requested):

tools/ecfr-parser/parseXML/parseXML.go
```diff
@@ -286,9 +300,91 @@ func (sl *SectionCitation) UnmarshalText(data []byte) error {
...
 }
 
+type Appendix struct {
+	Type     string          `xml:"TYPE,attr" json:"node_type"`
+	Citation AppendixCitation `xml:"N,attr" json:"label"`
+	Header   string          `xml:"HEAD" json:"title"`
+	Children AppendixChildren `xml:",any" json:"children"`
+}
+
+func (s *Appendix) PostProcess() error {
```
whytheplatypus, line 310: If other tools aren't citing appendix paragraphs, remove this.

regulations/templatetags/paragraphs.py
```diff
@@ -6,7 +6,14 @@
...
-    depth = len(value.get("label", []) or []) - section_depth
-    if len(value.get("marker", []) or []) > 1:
-        depth = depth - (len(value.get("marker", []) or []) - 1)
-    return depth
+    parent_type = value.get("parent_type", "")
+    label_len = len(value.get("label", []))
+    marker_len = len(value.get("marker", []) or [])
+    depth = label_len - section_depth
+
+    if parent_type == "appendix" or label_len == 0:
```
whytheplatypus, line 14: Should be able to remove the reference to `appendix`. may be able to go even further and simply say at the end that `if depth < 1, return 1, otherwise return depth` (basically return a depth of at least one.

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #159, opened 2021-08-27
feat: pass TESTING flag into app during cypress deploy action

PhilR8:
first pass at attempting to pass in ENV variable to app during cypress e2e automated tests that are triggered during the deploy github action

### DIFF
--- a/cmcs_regulations/settings.py
+++ b/cmcs_regulations/settings.py
@@ -159,4 +159,3 @@ HTTP_AUTH_USER = os.environ.get("HTTP_AUTH_USER")
 HTTP_AUTH_PASSWORD = os.environ.get("HTTP_AUTH_PASSWORD")
 
 GA_ID = os.environ.get("GA_ID")
-AUTOMATED_TEST = os.environ.get("AUTOMATED_TEST")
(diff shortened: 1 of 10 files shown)


cmcs_regulations/settings.py
```diff
@@ -158,3 +159,4 @@
 HTTP_AUTH_PASSWORD = os.environ.get("HTTP_AUTH_PASSWORD")
 
 GA_ID = os.environ.get("GA_ID")
+AUTOMATED_TEST = os.environ.get("AUTOMATED_TEST")
```
whytheplatypus, line 162: can get rid of this with the new pattern

regulations/templates/regulations/base.html
```diff
@@ -11,10 +11,12 @@
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <script async src="https://www.googletagmanager.com/gtag/js?id={{ GA_ID }}"></script>
     <script>
+        if ( "{{ AUTOMATED_TEST }}" === "true" ) {
+            window["ga-disable-{{ GA_ID }}"] = true;
```
whytheplatypus, line 15: Should be able to simplify this to `window["ga-disable-{{ GA_ID }}"] = {{ AUTOMATED_TEST }}`

PhilR8: yeah good call - I ended up with this: 

`window["ga-disable-{{ GA_ID }}"] = {{ AUTOMATED_TEST | yesno:"true,false" }};`

whytheplatypus (approved):

PR #164, opened 2021-08-30
fix: use env variable for custom cypress config item

PhilR8:


.github/workflows/deploy.yml
```diff
@@ -67,4 +67,6 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           working-directory: e2e
-          config: baseUrl=${{ steps.deploy-regulations-site-server.outputs.url }},DEPLOYING_TO_PROD=${{ matrix.environment == 'prod' }}
+          config: baseUrl=${{ steps.deploy-regulations-site-server.outputs.url }}
+        env:
+          CYPRESS_DEPLOYING: true
```
PhilR8, line 72: https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_

passing OS-level environment variables with the CYPRESS_ prefix will automatically be added to Cypress' environment variables and made available to the test.

The `config` parameter in the `with` directive seems to allow you to use any of the available configuration options - and my interpretation of the following page makes it seem like you can only use provided options and cannot pass in any custom options:

https://docs.cypress.io/guides/references/configuration

PhilR8: also to note is that I'm not just targeting prod, but I'm just gonna have it be true for all environments when running automated tests.

whytheplatypus (approved):

PR #167, opened 2021-09-01
EREGCSC-850 Footer adjustment

inkandsound:


regulations/static/regulations/css/scss/partials/_footer.scss
```diff
@@ -112,7 +132,7 @@ footer {
 
     // divide to strip unit (px)
     // https://css-tricks.com/snippets/sass/strip-unit-function/
-    @include custom-max($mobile-max / 1px) {
+    @include custom-max($tablet-max / 1px) {
```
PhilR8, line 135: The tablet comp shows the footer being three columns at widths 768px and above:

https://www.figma.com/file/wEaWv7FFtYluyMaJ4qTRsp/eRegs-Full-Responsive-Experience?node-id=429%3A163

regulations/static/regulations/css/scss/partials/_footer.scss
```diff
@@ -70,29 +76,43 @@ footer {
...
             padding: $spacer-4 0;
         }
     }
 
     .footer-right {
+
+        @include custom-max($tablet-max / 1px) {
+            flex: none;
+        }
+
```
PhilR8, line 104: the negative margin is pulling elements outside of their parent: 
Screen Shot 2021-09-01 at 11 10 09

I think we can get the same exact effect and keep everything inside of their containing elements by removing the negative margin left for all children and adding this to `.footer-right`:

```suggestion
    padding: 0 16px;
    align-items: start;
    justify-content: start;
```

PhilR8 (changes requested):

regulations/static/regulations/css/scss/partials/_footer.scss
inkandsound: @PhilR8 I adjusted to break it sooner because It was a bit awkward for smaller tablets once live. iPad Pros will have the 3-column still.

regulations/static/regulations/css/scss/partials/_footer.scss
inkandsound: @PhilR8 I know it pulls outside of the parent, but the issue is really that the middle element is too wide. The footer columns need to align with the content columns above them so they don't look off. Pulling the right side outside of its parent seemed less awkward than adjusting the middle column when I played with it, but open to suggestions.

PhilR8: Adding the padding to footer-right will make the middle narrower.  Right now the footer is set up with flexbox so that `.footer-left` and `.footer-right` have a defined flex-basis value (which is like a min-width).  `.footer-middle` doesn't - it will expand or contract to fill the remaining width.  Adding padding to `.footer-right` will make the middle narrower, while also having the right side of the footer match the width of the right sidebar in `.homepage-main-content`.

If you remove the negative margin and add the padding, everything looks exactly the same, with the benefit of having everything stay within their containing elements.

regulations/static/regulations/css/scss/partials/_footer.scss
PhilR8: Sounds good!

regulations/static/regulations/css/scss/partials/_footer.scss
PhilR8: I've made the tweaks - let's see if this .mov file below works:

https://user-images.githubusercontent.com/1651907/132902810-0375fd6e-9d98-4892-b053-5f07a6c4830b.mov

The footer should work basically the same as you had it, but `footer-right` keeps everything in its container, the middle element is narrower, and everything lines up nicely with the right sidebar as well.  

I've also done some tweaks to the external link icon - I had added padding for its use elsewhere but it introduced a side effect here, so I've just removed the padding.

PhilR8 (approved):

regulations/static/regulations/css/scss/partials/_reg_text.scss
```diff
@@ -5,6 +5,9 @@
 @import "../application_settings";
 
 .reg-text {
+    overflow-wrap: break-word;
+    overflow: auto;
+
```
PhilR8, line 10: This was to correct long, unbroken strings that were making the reg text too wide.  Example:

https://regulations-pilot.cms.gov/42/460/Subpart-E/2021-05-12/#460-72-e

has a big URL in it that was causing issues

PR #191, opened 2021-09-29
Added a Command to the Makefile to create a local superuser

theabrad:


knollfear (approved):
Tested locally and it works as expected.

PR #196, opened 2021-10-12
Eregcsc 701 copy btn

PhilR8:


regulations/static/components/tooltips/CopyBtn.vue
```diff
@@ -0,0 +1,259 @@
...
+
+    return elType === "labeled-icon"
+        ? el.offsetWidth / 2
+        : el.offsetWidth * 0.7;
+};
+
+const getAnchorX = (el, elType) => {
+    if (!el) return 0;
+
+    return parseInt(window.getComputedStyle(el).fontSize) + 20;
```
knollfear, line 83: idk if this will reliably return an integer.  I looked at these docs and it is clearly returning 32px.  https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle

regulations/static/components/tooltips/CopyBtn.vue
```diff
@@ -0,0 +1,259 @@
+<template>
+    <div class="copy-btn-container">
+        <button
+            class="copy-btn text-btn"
+            :class="buttonClasses"
+            :aria-label="btn_type === 'icon' ? ariaLabel : false"
```
knollfear, line 6: this may be my inexperience with vue, but this seems odd to me.  If the btn_type is not "icon" the aria label is "false"?  also, why spread out the logic into 2 places?  

```javascript
        :aria-label="ariaLabel"
        ...
        ariaLabel() {
            return btn_type === 'icon' ? `${this.label} for ${this.title}`: '';
        }

```

regulations/static/components/tooltips/CopyBtn.vue
```diff
@@ -0,0 +1,259 @@
...
+
+    watch: {
+        copyStatus: async function (newStatus, oldStatus) {
+            if (
+                newStatus === "pending" &&
+                (oldStatus === "idle" || oldStatus === "success")
+            ) {
+                try {
+                    // async write to clipboard
+                    await navigator.clipboard.writeText(this.getCopyText());
```
knollfear, line 161: Do we need to support IE?  It does not support this api  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard

knollfear (commented):
This looks good. I made some style comments.  Feel free to fix them or ignore them as you see fit.

regulations/static/components/tooltips/CopyBtn.vue
PhilR8: Fixed.

regulations/static/components/tooltips/CopyBtn.vue
```diff
@@ -0,0 +1,259 @@
...
+                    "click",
+                    el.__clickawayEventHandler__
+                );
+            },
+        },
+    },
+
+    methods: {
+        handleEnter(e) {
+            if (!this.entered && !this.clicked) this.entered = true;
```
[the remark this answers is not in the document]
PhilR8: updated.

regulations/static/components/tooltips/CopyBtn.vue
```diff
@@ -0,0 +1,259 @@
...
+        : el.offsetWidth * 0.7;
+};
+
+const getAnchorX = (el, elType) => {
+    if (!el) return 0;
+
+    return parseInt(window.getComputedStyle(el).fontSize) + 20;
+};
+
+const appendPxSuffix = (int) => `${int}px`;
```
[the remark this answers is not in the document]
PhilR8: although on second thought I'm going to wait until perhaps we need to use it again - I'm looking at some Vue-specific ways to implement utility functions like this (mixin, plugin, etc) and I would want to give it more attention to do one of those ways right.  So for now I'm going to leave it here.

PR #201, opened 2021-10-18
EREGCSC-995 - enable JS/VueJS sourcemaps

PhilR8:
Passing `needMap: false` in rollup-plugin-vue options, as well as using `--sourcemap` flag when running rollup, seems to solve this issue.

See this conversation for conversation about how this issue was reported and solved:

https://github.com/vuejs/rollup-plugin-vue/issues/272


knollfear (approved):
This is a big improvement.  LGTM!

PR #215, opened 2021-10-28
EREGCSC-1020 Admin panel fixups

cgodwin1:


supplemental_content/admin.py
```diff
@@ -47,6 +50,11 @@ class SectionAdmin(BaseAdmin):
     search_fields = ["title", "part", "section_id"]
     ordering = ("title", "part", "section_id", "parent")
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
```
knollfear, line 53: Does this belong in BaseAdmin?  I feel like in some cases this won't be the right thing, even if it is the right thing today.

knollfear (approved):
LGTM!

PR #225, opened 2021-11-04
EREGCSC-1022 Revise eCFR API fetch strategy

cgodwin1:
Utilizing thread pool to limit active connections to 3 at a time. User-configurable via command-line argument "-threads".

Also improved some logging messages. E.g.: changed "[MAIN]" to "[main]" for easier reading.


cgodwin1:
Note: still need to determine how to handle retries for individual parts. Global retry (i.e. 3 tries total and the whole program stops), or 3 tries for each part? (easier to implement)

knollfear (approved):
Looks good.  Tested locally and found more threads to decrease run time.  3 workers is appx twice as fast as 1.

PR #228, opened 2021-11-08
Knollfear/1012 escape label text

knollfear:
This PR adds in a CURL command outside of the ECFR run and inside it to ensure that general network connectivity is intact.  It also moves the creation of the HTTP client a bit so that it is recreated very time instead of recycled.  THis seems to affect how we respond to errors in fetching the data.  It also decodes the incoming HTML so that &amp is represented correctly as just &.

Test run was successful:
https://github.com/CMSgov/cmcs-eregulations/actions/runs/1435963882

Some other super small fixes are in here too such as removing a - from the home page banner and fixing the carets on the supplemental content.


cgodwin1 (approved):
Some minor spacing issues in ecfr.go (50, 67, 76), versions.go (43), and main.go (102-104, 150). Otherwise LGTM.

I would like to do a cleanup PR once we get everything fixed and working so these aren't important right now anyway.

PR #234, opened 2021-11-11
EREGCSC-922: "+ Show More" right sidebar button animation

PhilR8:


knollfear (approved):
looks good to me.  Works as expected.

PR #240, opened 2021-11-17
EREGCSC-1034: Standardize header height around application

PhilR8:
Take header styles previously applied only to landing/homepage and apply them to all pages of the site.  Specifically:

- Regulation content pages
- Search result pages
- About page


knollfear (approved):
LGTM

PR #243, opened 2021-11-18
EREGCSC-1037 Make reader view request supp content for subparts

cgodwin1:
Reader view now retrieves a list of subpart(s) on the current page and passes that to the supplemental content vue component.

If supplemental content is assigned to an entire section, it will now be correctly retrieved and displayed.


knollfear (approved):
This looks good and works locally as expected.

PR #245, opened 2021-11-19
Implement Django Report Admin

knollfear:
THis installs and exposes to the admin panel the django report admin module that allows the creation of customized reports by anyone with admin access.


cmcs_regulations/urls.py
```diff
@@ -23,5 +23,6 @@
     path("", include('regulations.urls')),
     path('favicon.ico', RedirectView.as_view(url=settings.STATIC_URL + 'images/favicon/favicon.ico')),
     path('admin/', admin.site.urls),
+    path(r'report_builder/', include('report_builder.urls')),
```
cgodwin1, line 26: Shouldn't need regex prefix here

cgodwin1 (approved):

cmcs_regulations/urls.py
knollfear: removed

PR #247, opened 2021-11-23
EREGCSC-985 Load Sections and Subparts to Supplemental Content

theabrad:
Added functionality to the parser to get a list of sections and subparts and post them to the supplemental content API to create sections and subparts automatically.


supplemental_content/views.py
```diff
@@ -42,3 +45,38 @@ def get_queryset(self):
...
+    def create(self, request, *args, **kwargs):
+        for section in request.data["sections"]:
+            try:
+                new_section, created = Section.objects.get_or_create(
+                            title=section["title"],
+                            part=section["part"],
+                            section_id=section["section"]
+                        )
+            except Exception:
+                return Response({'error': True, 'content': 'Exception!'}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
cgodwin1, line 60: Since all exceptions are the same, and they all cause the function to return early, would it be suitable to wrap the entire function body in a big `try` `except` block instead of having 3 separate blocks? Not applicable if you're going to be specializing each one obviously.

Makefile
```diff
@@ -53,23 +53,28 @@ tools/ecfr-parser/build/ecfr-parser: tools/ecfr-parser/*.go tools/ecfr-parser/**
 
 data.prod: ## Load a Part of Title 42. e.g. make data.prod.435 will load Part 435 into prod
 data.prod: CORE_URL = https://3iok6sq3ui.execute-api.us-east-1.amazonaws.com/prod/v2/
+data.prod: SUPPLEMENTAL_URL = https://3iok6sq3ui.execute-api.us-east-1.amazonaws.com/prod/v2/supplemental_content
```
knollfear, line 56:
```suggestion
data.prod: SUPPLEMENTAL_URL = $(CORE_URL)supplemental_content
```

Makefile
```diff
@@ -53,23 +53,28 @@ tools/ecfr-parser/build/ecfr-parser: tools/ecfr-parser/*.go tools/ecfr-parser/**
...
 data.val: CORE_URL = https://qavc1ytrff.execute-api.us-east-1.amazonaws.com/val/v2/
+data.val: SUPPLEMENTAL_URL = https://qavc1ytrff.execute-api.us-east-1.amazonaws.com/val/v2/supplemental_content
 
 data.dev: ## Load a Part of Title 42. e.g. make dev.data.435 will load Part 435 into dev
 data.dev: CORE_URL = https://hittwbzqah.execute-api.us-east-1.amazonaws.com/dev/v2/
+data.dev: SUPPLEMENTAL_URL = https://hittwbzqah.execute-api.us-east-1.amazonaws.com/dev/v2/supplemental_content
 
 data.experimental: ## Load a Part of Title 42. e.g. make data.experimental URL=[experimental lambda URL] into dev-experimental
 data.experimental: CORE_URL = $(URL)/v2/
+data.experimental: SUPPLEMENTAL_URL = $(URL)/v2/supplemental_content
```
knollfear, line 68: this doesn't seem quite right.  /v2/ would be duplicated

Makefile
```diff
@@ -53,23 +53,28 @@ tools/ecfr-parser/build/ecfr-parser: tools/ecfr-parser/*.go tools/ecfr-parser/**
 
 data.prod: ## Load a Part of Title 42. e.g. make data.prod.435 will load Part 435 into prod
 data.prod: CORE_URL = https://3iok6sq3ui.execute-api.us-east-1.amazonaws.com/prod/v2/
+data.prod: SUPPLEMENTAL_URL = https://3iok6sq3ui.execute-api.us-east-1.amazonaws.com/prod/v2/supplemental_content
 
 data.val: ## Load a Part of Title 42. e.g. make data.val.435 will load Part 435 into val
 data.val: CORE_URL = https://qavc1ytrff.execute-api.us-east-1.amazonaws.com/val/v2/
+data.val: SUPPLEMENTAL_URL = https://qavc1ytrff.execute-api.us-east-1.amazonaws.com/val/v2/supplemental_content
```
knollfear, line 60: see above change

Makefile
```diff
@@ -53,23 +53,28 @@ tools/ecfr-parser/build/ecfr-parser: tools/ecfr-parser/*.go tools/ecfr-parser/**
...
 
 data.dev: ## Load a Part of Title 42. e.g. make dev.data.435 will load Part 435 into dev
 data.dev: CORE_URL = https://hittwbzqah.execute-api.us-east-1.amazonaws.com/dev/v2/
+data.dev: SUPPLEMENTAL_URL = https://hittwbzqah.execute-api.us-east-1.amazonaws.com/dev/v2/supplemental_content
```
knollfear, line 64: see above change

Makefile
```diff
@@ -53,23 +53,28 @@ tools/ecfr-parser/build/ecfr-parser: tools/ecfr-parser/*.go tools/ecfr-parser/**
...
 
 data.experimental: ## Load a Part of Title 42. e.g. make data.experimental URL=[experimental lambda URL] into dev-experimental
 data.experimental: CORE_URL = $(URL)/v2/
+data.experimental: SUPPLEMENTAL_URL = $(URL)/v2/supplemental_content
 
 data.local: ## Load a Part of Title 42. e.g. make data.local.435 will load Part 435
 data.local: CORE_URL = http://localhost:8000/v2/
+data.local: SUPPLEMENTAL_URL = http://localhost:8000/v2/supplemental_content
```
knollfear, line 72: see above change

Makefile
```diff
@@ -53,23 +53,28 @@ tools/ecfr-parser/build/ecfr-parser: tools/ecfr-parser/*.go tools/ecfr-parser/**
...
 data.local: export EREGS_USERNAME=RpSS01rhbx
 data.local: export EREGS_PASSWORD=UkOAsfkItN
 
 data.%: tools/ecfr-parser/build/ecfr-parser
-	./tools/ecfr-parser/build/ecfr-parser -loglevel trace -log-parse-errors=false -attempts 3 -title 42 -subchapter IV-C -parts 400,457,460 -eregs-url $(CORE_URL)
+	./tools/ecfr-parser/build/ecfr-parser -loglevel trace -log-parse-errors=false -attempts 3 -title 42 -subchapter IV-C -parts 400,457,460 -eregs-url $(CORE_URL) -eregs-supplemental-url $(SUPPLEMENTAL_URL)
```
knollfear, line 77: do we really need to pass this in?  Can't we just construct it in the app itself from the CORE_URL?  It seems relatively static.

supplemental_content/views.py
```diff
@@ -42,3 +45,38 @@ def get_queryset(self):
                 )
             ).distinct()
         return query
+
+
+class SupplementalContentSectionsView(generics.CreateAPIView):
+    def create(self, request, *args, **kwargs):
+        for section in request.data["sections"]:
+            try:
+                new_section, created = Section.objects.get_or_create(
```
knollfear, line 54: We don't need to capture the return variables if we aren't using them.  also, can/should we validate the values in any way.  Is a part of "KNOLL" acceptable?  It not, we should figure out what is acceptable and check it somehow.

supplemental_content/views.py
```diff
@@ -42,3 +45,38 @@ def get_queryset(self):
                 )
             ).distinct()
         return query
+
+
+class SupplementalContentSectionsView(generics.CreateAPIView):
```
knollfear, line 50: What kind of permissions does this endpoint have?  We can't have just anyone posting these to our URL.

supplemental_content/views.py
```diff
@@ -42,3 +45,38 @@ def get_queryset(self):
...
+    def create(self, request, *args, **kwargs):
+        for section in request.data["sections"]:
+            try:
+                new_section, created = Section.objects.get_or_create(
+                            title=section["title"],
+                            part=section["part"],
+                            section_id=section["section"]
+                        )
+            except Exception:
```
knollfear, line 59: can we also add a bit of info about what specific thing errored out?  We are going to be troubleshooting this, we should make our lives easier.

tools/ecfr-parser/ecfr/sections.go
```diff
@@ -0,0 +1,78 @@
...
+type Subpart struct {
+	Title    string    `json:"title"`
+	Part     string    `json:"part"`
+	Subpart  string    `json:"subpart"`
+	Sections []Section `json:"sections"`
+}
+
+func ExtractStructure(s Structure) (Part, error) {
+	title := s.Identifier[0]
+	structurePart := s.Children[0].Children[0].Children
```
knollfear, line 25: This doesn't feel great.   Why is it the third child?  Do we know the rest don't have multiple children?

Makefile
cgodwin1: Wasn't the purpose of passing it in to make uploading this data optional? Could use a bool flag though

tools/ecfr-parser/ecfr/sections.go
cgodwin1: This is due to the structure pulled from eCFR. It is indeed a static thing because the top level is title, then chapter, then subchapter. See https://regulations-pilot.cms.gov/v2/2021-01-01/title/42

theabrad: ecfr.gov structure will always have the "part" be the third child. i.e https://www.ecfr.gov/api/versioner/v1/structure/2021-09-20/title-42.json?part=456
I could add in a check to make sure the 3rd child's type == "part"

supplemental_content/views.py
theabrad: There was no auth. I just added it in.

PR #248, opened 2021-11-23
EREGSCSC-1022 - Make ECFR parser more reliable

knollfear:
This seems to be choking strictly on versions from 2016 so in the case of a 2016 version the parser will adjust the date of the version to 2017-01-01.  This appears to be relatively consistent across the board, but I will also be following up with the ECFR team.


cgodwin1 (approved):
Should follow up on this in future with eCFR and see if we can get rid of this workaround

PR #252, opened 2021-12-02
EREGCSC-1074 - getRules method errors when improperly handling undefined

PhilR8:


regulations/static/components/RelatedRules.vue
```diff
@@ -114,7 +114,7 @@ export default {
           while(url){
             const response = await fetch(url);
             const rules = await response.json();
-            results = results.concat(rules.results)
+            results = results.concat(rules.results ?? []);
```
PhilR8, line 117: when there are no results to the query, the returned `rules` object from the `fetch_rules` method does not have a `results` property. Attempting to access `rules.results` returns `undefined`.  This `undefined` value was being added to the returned array, so that array was no longer empty - it contained an `undefined` primitive (like so: `[undefined]`).  

The `getRules` method downstream assumes either an empty array for no results or an array containing properly formed rule objects.  When `getRules` finds the `undefined` value in the array, it throws an unhandled error.

This fix uses the nullish coalescing operator to return an empty array instead of `undefined` if no results are returned.

knollfear (approved):

PR #253, opened 2021-12-02
EREGCSC-1058 Skip reg parts that have already been imported

knollfear:
THis Pr creates a public endpoint that lists all of the dates and parts that are currently in the database.  The ECFR parser pulls in this file and if a date/part combo has been pulled in already it skips enqueuing that part.  This will lead to the parser running in only a few seconds when there is no new data to pull.


regcore/urls.py
```diff
@@ -18,5 +19,6 @@
         path("<date>/title/<title>", EffectivePartsView.as_view()),
         path("<date>/title/<title>/part/<name>", EffectivePartView.as_view()),
         path("<date>/title/<title>/part/<name>/toc", EffectivePartTocView.as_view()),
+        path("title/<title>/existing", ExistingPartsView.as_view())
```
cgodwin1, line 22: Recommend comma after this line for cleaner future diffs

regcore/views.py
```diff
@@ -32,6 +33,15 @@ class Meta:
         }
 
 
+class ExistingPartSerializer(serializers.BaseSerializer):
+
+    def to_representation(self, instance):
+        return {
+            'date': instance.get("date"),
+            'partName': instance.get("partName").split(",")
```
cgodwin1, line 41: Same as previous

tools/ecfr-parser/eregs/eregs.go
```diff
@@ -70,3 +78,55 @@ func PostPart(ctx context.Context, p *Part) error {
 	log.Trace("[eregs] Posted ", length, " bytes for part ", p.Name, " version ", p.Date, " in ", time.Since(start))
 	return nil
 }
+
```
cgodwin1, line 81: Noticing some strange spacing errors in this file, on lines 85-101, 104-122, 130-131

tools/ecfr-parser/main.go
```diff
@@ -136,6 +136,10 @@ func run() error {
```
cgodwin1, line 136: More spacing issues in this file as well

cgodwin1 (commented):
Looks great! Just some minor formatting issues

regcore/urls.py
knollfear: added

regcore/views.py
knollfear: added

tools/ecfr-parser/eregs/eregs.go
knollfear: I think I improved it.  If you have any suggestions, let's sync up and talk them out.

tools/ecfr-parser/main.go
knollfear: same as above.

cgodwin1 (approved):
LGTM

theabrad (approved):
Pulled it down. Ran with empty DB and then ran again and saw skipped versioned parts.

PR #277, opened 2021-12-03
EREGCSC-1071 Get eCFR parser running in AWS

cgodwin1:
Parser to run as a Lambda on a daily schedule instead of Github Actions.

Currently, configuration (e.g. title, part, etc) is stored as environment variables retrieved from AWS SSM.

Also reverts the workaround for Github Actions that creates a new client for each eCFR request. This made it work in GHA, but leaked the file descriptor for every connection opened and caused failures on some systems.


.github/workflows/load-regulations.yml
```diff
@@ -19,14 +17,17 @@ jobs:
...
-          EREGS_PASSWORD: ${{ secrets.EREGS_PASSWORD }}
-          NAME: ${{ matrix.environment }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        run: |
+          pushd serverless
+          npm install serverless -g
+          npm install
+          AWS_CLIENT_TIMEOUT=360000 serverless invoke --config ./serverless-parser.yml --function ecfr_parser --stage ${{ matrix.environment }}
+          popd
```
knollfear, line 33: do we need this?

knollfear (approved):
Good work.  LGTM!

PR #283, opened 2021-12-11
Remove "Using this Tool" section from about page

brittag:
As part of EREGCSC-950 I want to take a different approach here (demo video with description), but it'll take a certain amount of time, and for now it's best to remove the outdated material to avoid confusion. I could not find anywhere that this section is currently referenced from, so I don't think I'm breaking any links.


knollfear (approved):
LGTM

PR #291, opened 2021-12-16
fixup: make homepage spec look for correct id

PhilR8:


PhilR8:
Correcting the ID for which the spec is looking makes the test pass again:

Screen Shot 2021-12-16 at 13 21 07

theabrad (approved):

PR #307, opened 2022-01-10
EREGCSC-1072 Make parser request runtime arguments from eRegs API (+ easy multi-title support)

cgodwin1:
Resolves #1072

**Description-**
After moving the parser to AWS, runtime arguments are specified using AWS parameter store values. This means that (for example) if someone wants to add a part to the site, they will need to gain AWS access, locate the correct variable in the parameter store, modify it, then redeploy the Lambda for the parser. This is not user-friendly at all!

Instead, parser will query an endpoint on the eRegs API to determine how it should run and what data should be parsed. Users will then use the admin site on eRegs to easily configure the parser in a user-friendly way.

This also has the benefit of making multi-title support for the parser trivial. The endpoint on eRegs will return a JSON object containing parameters and an arbitrary number of titles to parse.

1. Create local admin user and login
2. Visit "Parser Configuration" page and observe/modify as needed
3. Visit /v2/parser_config and verify correct parameters returned
4. Run parser and verify correct behavior

Note: Parser now has multi-title support and this can be verified by adding a new title in the admin panel and re-running the parser.


knollfear (approved):
Installed and ran locally as expected.

PR #325, opened 2022-01-28
Made row clickable django admin

thwalker6:
Made whole row clickable  for supplemental content table in django admin.  Added a value of "NONE" if there is no value assigned.  No impact on database just in place of "-"


knollfear (approved):
LGTM.  Works locally as expected.

PR #331, opened 2022-02-04
EREGCSC-1174 Prototype Data Access Layer

PhilR8:
Resolves [EREGCSC-1174 [Prototype] Create Data Access Layer](https://jiraent.cms.gov/browse/EREGCSC-1174)

**Description**

Create API layer containing methods to fetch data from server and transform to desired shape for use in Vue components.

- Adds an api.js file with methods to pull down part names, most recent date of update, and homepage ToC information
- Displays most recent date of update in footer
- Homepage table of contents is rendered and Part Link takes the user to a part page via Vue Dynamic Route Matching
- Enables JumpTo Vue component to navigate to part page via Vue Dynamic Route Matching

1. Visit [deployed prototype](https://d2hgxe2x7z11zt.cloudfront.net/) and ensure all above items are present.


solution/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:3-alpine
 
-COPY ["static-assets/requirements/requirements.txt", "/app/src/"]
```
knollfear, line 3: Thank you.  I didn't realize I had messed this up.

solution/ui/prototype/src/utilities/api.js
```diff
@@ -0,0 +1,371 @@
...
+        method: "DELETE",
+        headers: authHeader(token),
+        params,
+        body: JSON.stringify(data),
+    });
+}
+
+// ---------- api calls ---------------
+
+const login = (username, password) => {
```
knollfear, line 222: This and the next method can probably get deleted.

solution/ui/prototype/src/utilities/api.js
```diff
@@ -0,0 +1,371 @@
...
+    return niceDate(_get(result.reduce(reducer), "date"));
+};
+
+const getPartNames = async (title = "42") => {
+    const result = await httpApiGet(`title/${title}/existing`);
+
+    return _sortedUniq(result.flatMap((part) => part.partName).sort());
+};
+
+const getHomepageStructure = async () => {
```
knollfear, line 250: Can this reducer be shelled out into a utils class?  It does some post processing of the result to make it into a more usable format right?  Should we do more of this in Django or in the parser anyways?

solution/ui/regulations/.eslintrc.json
```diff
@@ -1,15 +1,15 @@
 {
     "env": {
         "browser": true,
-        "es2021": true
+        "es2020": true
```
knollfear, line 4: moving backwards?

PhilR8: I can probably change this back.  I'm having inconsistent ESLint behavior after we reorganized the repository and this was a suggestion from a stackoverflow article that worked for me.  But I just switched it back this morning and things are looking okay... so I'll revert this to es2021.

solution/ui/prototype/src/utilities/api.js
PhilR8: we're totally outside of Django here so this would have to be something the parser does of the front end does.  

Caleb and I talked about using the parser to create a bunch of new data structures every time it runs, structures that would be specific to certain endpoints or parts of the prototype.  So yes, the parser would be a better place to do this transformation.  Caleb and I talked about some future stories that could/would be written to do exactly that.

As far as where to put this - I think this could be put into a utils class for sure, broken up into parts that can be composed together.  Although I don't think now is the right time to do that because we don't know what parts of this reducer would be helpful elsewhere.

However, I've started work on EREGCSC-1182 - building out the JS part view - and I'm seeing likely areas of overlap.  So this would be work better suited for that story.

solution/ui/prototype/src/utilities/api.js
PhilR8: deleted

knollfear (approved):
LGTM!

PR #333, opened 2022-02-08
Eregcmsc 1073 make categories show up

knollfear:
Resolves #1093

**Description-**
Delivers list of categories and sub categories to the frontend on page load and then integrates the API response into that list.

- Adds categories and sub categories to the page load data
- integrates API response to that list on API call success

1. Navigate to a page with expected supplemental content and click around to make sure it is still there as expected and extra categories show up.  
2. ****"SHow if Empty" does NOT work at this moment.  I can make it work with minimal effort.

### DIFF
--- a/solution/ui/regulations/css/scss/partials/_sidebar_right.scss
+++ b/solution/ui/regulations/css/scss/partials/_sidebar_right.scss
@@ -29,6 +29,10 @@ aside.right-sidebar {
         }
     }
 
+    #subpart-resources-heading{
+        margin-bottom:0px;
+    }
+
     #view-button {
         font-size: $base-font-size;
         color: $primary_link_color;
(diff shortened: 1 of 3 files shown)


solution/backend/regulations/templates/regulations/partials/sidebar_right.html
```diff
@@ -23,6 +25,7 @@ <h2 id="subpart-resources-heading">
         part="{{ reg_part }}"
         :sections="{{ node_list.sections }}"
         :subparts="{{ node_list.subparts }}"
+
```
thwalker6, line 28: Super nitpicky but empty line should probably be removed

knollfear: fixed. nitpick away.

cgodwin1 (approved):
LGTM

PhilR8 (approved):

PR #336, opened 2022-02-11
Fix supplemental content migration to properly generate display_name

cgodwin1:
Resolves #1187

**Description-**
Previous migration did not successfully generate display_name field for children of AbstractLocation. This is because we utilized get_model (as recommended) which reconstructs models on the fly from previous migrations. These reconstructed models do not include any additional code such as save hooks, so the custom save hook did not run.

Another migration now exists which directly imports supplemental_content.models.AbstractLocation and resaves all instances. It silently skips running if the model is changed, renamed, or deleted in the future.

1. Ensure you have a fully up-to-date main and all migrations are run
2. Checkout this branch and run migrations again
3. Verify display_name field is properly filled out in Django admin


knollfear (approved):
LGTM

PR #341, opened 2022-02-17
EREGCSC-1194 Optimize supplemental content models

cgodwin1:
Resolves #1194

**Description-**
This is an extension of https://github.com/CMSgov/cmcs-eregulations/pull/335

Generating string representations of inherited models is expensive due to the need for a recursive algorithm to walk through parent and child classes. It was previously assumed that this was done in-memory, but in fact it generates an excessive number of database queries.

This applies the display_name field method [done here](https://github.com/CMSgov/cmcs-eregulations/pull/335) to the rest of the supplemental content models to significantly reduce the number of queries.

- All supplemental content models now inherit from AbstractModel
- AbstractModel has a display_name field and methods for filling it on save
- Recursive name generation algorithm only runs on save instead of on each page load

1. Make main up to date and have some active supplemental content models (sections/subparts will do).
2. Run `docker-compose exec regulations python manage.py migrate`. This will take a few seconds as all the names are computed for the first time.
3. Visit Django admin and verify a sane number of database queries using Django Debug Toolbar.


cgodwin1:
PR closed to re-deploy experimental branch

PR #343, opened 2022-02-18
Eregcmsc 1184 resources modal

knollfear:
Resolves #1184

**Description-**
Creates a resource view that shows supplemental content for a single section

- Adds a "View Section Resources" button to the Part page
- added 10 fake "sections" to each page.  These are needed to ensure that the button changes the resources correctly.
- adds a caching process to data pulled from the API.  Go to .../cache to manually edit, delete, or add cache values.

1. go to any part page and click on the "View Resources" button
2. Page should come up with resources shown.  You can manually add data for this endpoint to the cache if you want to see "real" values.
note that the first section won't open the resource drawer because it is 0 indexed.  This is not a legit bug as there are no section 0's in our data.


cgodwin1 (approved):
LGTM

PR #349, opened 2022-02-28
existingpart

thwalker6:
Resolves #
 CREGCSC-1162

**Description-**
Add swagger documentation to the API

Adds an api view for our backend.

Slight modification to the api for 
V2/title/{title}/part/{part}/supplmentalcontnent.   Since the swagger view does not work with get_queryset it had to be changed to a get and modifying how the response is serialized.  End result is the same though.
 
Go to /api/Swwagger.  This will take you to the swagger view.  Test out any of the apis.  Some require authentication

To test the other change  add this to the end of the url in the normal environment and test environment:
v2/title/42/part/433/supplemental_content?&sections=8&sections=10&sections=11&sections=15&sections=32&sections=34&sections=35&sections=36&sections=37&sections=38&sections=40&subparts=A

Results should be the same.

### DIFF
--- a/solution/backend/cmcs_regulations/settings.py
+++ b/solution/backend/cmcs_regulations/settings.py
@@ -46,6 +46,7 @@ INSTALLED_APPS = [
     'solo',
     'django_opensearch_dsl',
     'corsheaders',
+    'drf_spectacular'
 ]
 
 MIDDLEWARE = [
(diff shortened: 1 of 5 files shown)


PhilR8:
Did we talk about only displaying GET endpoints - i.e. hiding all DELETE etc endpoints?

cgodwin1:
> Did we talk about only displaying GET endpoints - i.e. hiding all DELETE etc endpoints?

We did. I think we decided from a security perspective it doesn't matter but may raise questions.

knollfear:
> > Did we talk about only displaying GET endpoints - i.e. hiding all DELETE etc endpoints?
> 
> We did. I think we decided from a security perspective it doesn't matter but may raise questions.

I put in an override for this to just return a 204 status code.  Assuming our password is secured there was no real security risk, but at the same time, I was able to actually delete a record when I tested it.

knollfear (approved):
Approved.  LGTM!

PR #355, opened 2022-03-08
Eregs 1193 splitscreen reader resource view

knollfear:
Resolves #1193

**Description-**
Creates a split pane view and implements a reader view in the left pane

- adds a new set of views:
- .../PD/<part>
- .../PD/<part>/subPart-<subPart Letter>
- .../PD/<part>/subPart-<subPart Letter>/<section>
each of these should display just the regulation section you expect.  Navigation and breadcrumbs should work as well
- Added endpoint (.../v2/supplemental_content_count_by_part?part=<part>)to return list of display names per part and the count of the number of pieces of supplemental content tied to each of those parts.

1. Navigate around using the previous/next buttons
2. Use the breadcrumbs
3. ensure that the supplemental content counts seems reasonable


solution/ui/prototype/src/components/node_types/Subpart.vue
```diff
@@ -42,12 +51,33 @@ export default {
...
 
     computed: {
         kebabTitle() {
             return getKebabTitle(this.node.label);
         },
+        numSupplementalContent(){
+            if (!this.supplementalContentCount){
+              return 0
+            }
+            const total = this.node.children.reduce((count, node) => {
```
PhilR8, line 75: getting a JS error when I visit https://d1eyo42ow73een.cloudfront.net/PD/42/431

```
TypeError: Cannot read properties of null (reading 'reduce')
    at a.numSupplementalContent (Subpart.vue:75:1)
```

solution/ui/prototype/src/components/part/PartContent.vue
```diff
@@ -47,6 +49,16 @@ export default {
             type: String,
             required: true,
         },
+        showResourceButtons: {
+            type: Boolean,
+            required: false,
+            default: true
+        },
+        supplementalContentCount:{
+          type: Object,
```
PhilR8, line 58: inconsistent spacing

PhilR8 (approved):

PR #362, opened 2022-03-10
Eregcmsc 1227 optimize serializer

knollfear:
Resolves #1227

**Description-**
Optimizes the serialization process that builds the category tree for the response.

- Reduces usage of the polymorphic serializer
- installs library to make working with inherited models easier
- builds category tree completely and then trims unused nodes instead of adding nodes form selected content and deduping.

1. go to part 440 subpart A and the page should load the supplemental content, currently this times out.


cgodwin1 (approved):
LGTM!

PR #367, opened 2022-03-17
copied in deployment for prototype site.

knollfear:
Resolves #1239

**Description-**

Adds deployment of the prototype site to the main deployment process.  This will allow for a static, non changing URL for the prototype site that connects to the production database.

- deployment action adds a section for deploying the prototype.

1. This can not be verified until it the PR is approved and the deployment action is run.
2. When the deploy script runs it should print the cloudfront ID to the console.  Otherwise the URL should be available on the AWS console.  That URL should show the prototype site.


PhilR8 (approved):

PhilR8:
LGTAS! (Let's give this a shot)

PR #368, opened 2022-03-17
EREGCSC-1226 hide show resource buttons

knollfear:
Resolves #EREGCSC-1226 and #EREGCSC-1229

**Description-**

Cleans up some interactions between the PD and JS views. Ensures that buttons for "Show Resources" only show if there are resources to show.

- Show Resources button only shows if that subpart or section resource has resources.  For subpart this means supplemental content directly on the subpart.  For the PD view the blue button will show content for all sections in that subpart.  

1. on 42/400 none of the subparts have supplemental content so those buttons should not show.
2.on 42/441 subpart B has content so should show up.  
3. 400.202 has no supplemental content so no show resources button should show.

### DIFF
--- a/solution/ui/prototype/src/components/PDPart/SectionCards.vue
+++ b/solution/ui/prototype/src/components/PDPart/SectionCards.vue
@@ -50,13 +50,11 @@
 </template>
 
 <script>
-import { getSupplementalContentNew } from "@/utilities/api";
 
 export default {
     name: "SectionCards",
     data: () => ({
         show: false,
-        supList: null
     }),
     props: {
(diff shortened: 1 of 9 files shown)


PhilR8:
This looks good.  It looks like the Section node issue has been corrected.

The view resources buttons are only being rendered if the subpart/section is included in the `supplemental_content_count_by_part` response, and the counts are matching up to the number of resource items shown in the drawer/sidebar. 

The console errors have also been cleaned up.

LGTM!

PhilR8 (approved):

PR #371, opened 2022-03-21
Eresgcsc 1212 resources

thwalker6:
Resolves #

Eregcsc- 1212 - Partially

Creating the resources page for Pams prototype.  The main reason i am checking this in is to introduce the utility for creating a treeselect drop down in vue.    

This is creating the api call in v2 to pull in all of the categories.  After that it is pulled into the vue app normally.  A utility was created to organize the categories in a format acceptable for Tree select.   Treeselect requires that you have the data inserted in a particular way.

[{ id:id, label:abel, children:children[children is optional}]

I broke this ticket out so that you all can use it since I think other components need it similarly.  There is still work to be done on this story but since others might need this piece I kind of wanted to break it off from here.

Go to /resources and the first drop down should have the resources with checkboxes for categories and subcategories.  Sub categories will be in a tree view of other categories.  If not subcategories no drop down will appear for them.


solution/ui/prototype/src/views/ResourcePage.vue
```diff
@@ -0,0 +1,45 @@
...
+        FlashBanner,
+        Header,
+        Footer,
+        Splitpanes,
+        Pane,
+        ResourceFilters,
+
+    },
+    data: () => ({
+        items: ["foo", "bar", "fizz", "buzz", "fizzbuzz", "foobar"],
```
knollfear, line 38: This also seems maybe not quite ready for merge into master.

solution/ui/prototype/src/components/ResourcesPage/ResourceFilters.vue
```diff
@@ -0,0 +1,78 @@
...
+        selectedTitles: [],
+        supList: [],
+        categories: [],
+        catOptions:[],
+        partandsubparts: []
+    }),
+ 
+    async created() {
+        try {
+            this.supList = await getSupplementalContentNew(42, 441, [], ["B"]);
```
knollfear, line 61: This doesn't feel quite right.  Shouldn't this be variable?  Is this what you were talking about at standup?

knollfear: I don't think this component needs any supplemental content in scope.  I assumed this component would emit up to a higher order component when a selection was made.

solution/ui/prototype/src/views/ResourcePage.vue
thwalker6: I just was pushing my code to a remote, not ready for a code review.  I might have accidentally hit request PR

solution/ui/prototype/src/utilities/api.js
```diff
@@ -396,7 +396,7 @@ const getHomepageStructure = async () => {
 
     return transformedResult;
 };
-
+getCategories
```
PhilR8, line 399: this is just hanging out here

solution/ui/prototype/src/components/ResourcesPage/ResourceFilters.vue
thwalker6: Removed.  This was just for testing purposes

PhilR8 (approved):

PR #381, opened 2022-03-24
EREGCSC-1243 [Prototype] Adjust homepage copy

cgodwin1:
Resolves #1243

**Description-**

In order to help us get good feedback testers who end up on the eRegs prototype homepage, we should have some reasonably realistic copy that they can react to, rather than just a small bit of text.

- Homepage hero now has text "eRegulations organizes together regulations, subregulatory guidance, and other related policy materials."

1. Visit prototype homepage and verify hero looks correct with new header text.


knollfear (approved):
LGTM

PR #383, opened 2022-03-25
EREGCSC-1224 - JS Resource Page "Option B"

PhilR8:
Resolves [EREGCSC-1224](https://jiraent.cms.gov/browse/EREGCSC-1224)

**Description**

This PR implements a NON-FUNCTIONAL resources sidebar (JS Resources Option B).  It is accessed either from the Parts Sidebar's "View All Resources" button, or from the `/resources-sidebar` route.

NOTE: if you visit without query params (eg title=42&part=433) then the Subpart and Section expansion panels are empty.

**This pull request changes**

- Adds new route for `/resources-sidebar` to show a LITTLE BIT of the JS Resources Option B comp.

1. Go to the prototype linked below, go to a part and add `sidebar` as the third URL param (ex: 42/431/sidebar) and then click the View All Resources button in the right sidebar.

Link straight to new route with query params:
https://d3bgpoi8tarxza.cloudfront.net/resources-sidebar?title=42&part=433&section=10


solution/ui/prototype/src/components/resources/ResourcesSidebarFilters.vue
```diff
@@ -0,0 +1,105 @@
...
+
+    computed: {
+        computedProp() {
+            return this.dataProp.toUpperCase();
+        },
+    },
+
+    methods: {
+        methodName() {
+            console.log("method has been invoked");
```
thwalker6, line 72: Does this need to be here?

solution/ui/prototype/src/views/Resources.vue
```diff
@@ -335,6 +357,7 @@ export default {
     beforeCreate() {},
 
     async created() {
+        console.log("route", this.$route);
```
thwalker6, line 360: Take it out, leave it not super particular, but probably doesnt need to be in there unless you think we might need it in the future for debugging

cgodwin1 (approved):

PR #388, opened 2022-03-30
EREGCSC-1257 Add depth field computation to parser

cgodwin1:
Resolves #1257
**Important!**

"Skip versions" must be disabled prior to deploying this to dev, val, and prod. Failure to do so risks breaking any page that makes use of `Part.toc`.

**Description-**

As a user I want eRegs to load as fast as possible, and as a developer I want code to be as bug-free as possible.
Currently, when traversing a structure or document to get to the part level, we use "structure.children[0].children[0].children[0]" to go from title to chapter to subchapter to part. This has the possibility of failing if a regulation structure e.g. includes a subtitle, or omits a subchapter.

- Depth field exists in the Part model of regcore
- TOC computed property of the Part model utilizes this depth field replacing children[0].children[0].children[0]
- Parser computes the depth of each part before sending it to eRegs

1. Re-run parser with "skip versions" disabled
2. Visit /v2/2021-03-01/title/42/part/433/toc or similar and verify the table of contents starts its listing at the Part level (e.g. `type = part` and `name = "Part 433 - State Fiscal Administration"` or similar)


knollfear (approved):
LGTM!

PR #390, opened 2022-03-31
EREGCSC-1259 Add middleware to render API responses as HTML so Django Debug Toolbar works

cgodwin1:
Resolves #1259

**Description-**

As a developer I need to be able to debug API endpoints quickly and effectively. Although we have the ability to use Django Debug Toolbar for frontend pages, it won't work on backend pages because those responses are in JSON.

- Adding `?html_api` to the end of any API call returns the JSON response as basic HTML. (Changed from `?debug` to be more precise.)

1. Visit any API page, e.g. `/v3/title/42/part/433/version/2021-03-01/toc` and append `?html_api` (`/v3/title/42/part/433/version/2021-03-01/toc?html_api`)
2. Verify that the JSON content returned is correct, and view page source to verify that it is wrapped in HTML tags
3. Verify that Django Debug Toolbar appears as expected


knollfear (approved):
LGTM!

PR #392, opened 2022-04-04
Explicitly setting the search config to english and allowing quoted search terms

knollfear:
Resolves #1094

**Description-**

Explicitly adds an english config to the search queries.

- Stemming, search vector creation and highlighting should all be different now.

1. Searching for "State Medicaid Manual" with and without the quotes.  With quotes should be more specific and return fewer results
2. Search for:    
    Application, applications
    Demonstration, demonstrations
    Authority, authorities
    Disability, disabilities 
3. you should see similar results for both terms.  Search results should be more inclusive of the highlights.  For example Disability should include "Disabled" as a highlighted word in the results indicating it was in fact searched for.


cgodwin1 (approved):
LGTM!

PR #393, opened 2022-04-05
Adjusted Javascript API to remove garbage from sectionlist

thwalker6:
Resolves #

**Description-**

There was a bug because of some nuances of the json returned that sections without subparts would not appear and that additional classifications will block them from appearing properly as well. This will fix that.

- expected change 1

The drop down on /pdresources for section should display all sections  and subparts.  No junky looking selections as wel.

2. 
Go to /pdresources and look at the section drop down.  Everything should look pretty clean.


knollfear (approved):
LGTM.  I hope the depth indicator on the V3 API helps here too.

PR #396, opened 2022-04-06
EREGCSC-1253 Create additional table of contents endpoints

cgodwin1:
Resolves #1253

**Description-**

As a developer it is necessary to retrieve a table of contents for titles, parts, and more. Ideally we want to retrieve only the data necessary to facilitate this.

The following new V3 endpoints exist:
- `/v3/title/<title>/toc`
  Retrieve table of contents for one title only (extension of /v3/title/<title>)
- `/v3/title/<title>/part/<part>/version/<date>/subpart/<subpart>/toc`
  Retrieve table of contents for a specific subpart down to the section level

**This pull request does NOT change...**

Originally a `/v3/title/<title>/part/<part>/version/<date>/section/<section>/toc` endpoint was planned, but this was done under the incorrect assumption that the `structure` field contains data down to the paragraph level. It doesn't, so to implement this would require a change in the parser to populate this information, or a complex algorithm to fill in this data from the `document` field.

Given that this level of granular detail is probably not necessary for a table of contents endpoint, I'm leaving it out for now. If we decide we want paragraphs in the tables of contents, I suggest we make another ticket to account for the extra complexity of modifying the parser.

1. Visit each of the endpoints listed above, substituting correct variables (e.g. `title=42`, `part=433`, `date=2021-03-01`, and `subpart=A` and verify the correct data is returned.

### DIFF
--- a/solution/backend/regcore/models.py
+++ b/solution/backend/regcore/models.py
@@ -46,14 +46,12 @@ class Part(models.Model):
         # TODO: add once /v2/title/X/existing is removed, the following line breaks it for some reason
         # ordering = ("title", "name", "-date")
 
-    def get_part_level(self, data):
-        for _ in range(self.depth):
-            data = data['children'][0]
-        return data
-
     @property
     def toc(self):
-        return self.get_part_level(self.structure)
+        structure = self.structure
+        for _ in range(self.depth):
+            structure = structure["children"][0]
+        return structure
 
 
 class ParserConfiguration(SingletonModel):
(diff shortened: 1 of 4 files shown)


PhilR8 (approved):
I took a look at these endpoints.  They look good - the Title ToC endpoint returns everything we need for the homepage ToC - I feel like this would almost be a drop-in replacement for `getAllParts` with a little tweaking to the `getHomepageStructure` method in the prototype's api file.  And it reduces the response from half a MB to 7.4Kb.  The Subpart ToC endpoint looks pretty good as well, although we haven't yet implemented that in the prototype.  

LGTM!

PR #397, opened 2022-04-07
EREGCSC126

thwalker6:
Resolves # [EREGCSC-1261](https://jiraent.cms.gov/browse/EREGCSC-1261)

**Description-**
A different way of pulling in supplemental content.  This will do it in two phases.  First you get a list of all supplemental content tied to each section/subpart.  Then you  pull all supplemental content by a part listed by its ID.  You take the IDS from the frist pull based on your slection and pull the values of the supplemental content by part.

- The way supplemental content is pulled for the /PDResources page

Go to /pdresources.  Select either a part, subpart or selection.  It will then pull the supplemental content associated with those.  If only a part is selected all supplemental content in the part will be pulled if subpart or section is sleeted it will go on that filtered of a level.


PhilR8 (approved):
LGTM!

PR #399, opened 2022-04-11
EREGCSC-1272 [Spike] Enable SME testing of supplemental content search

cgodwin1:
Resolves #1272

**Description-**

In order to enable testing of supplemental content search we need a page which returns search results for supplemental content.

- The prototype `/PDresources` page is hooked into basic supplemental content search.
- Using the search bar will return search results, with highlighted text, and a simple indicator as to what the user searched for.

1. Visit prototype page `/PDresources`.
2. Search for "medicaid" and "test", and verify the correct results are returned.
3. Use the filter options normally to make sure those still work properly.


solution/ui/prototype/src/views/ResourcePage.vue
```diff
@@ -4,6 +4,9 @@
             <Header />
             <div class="searchpane">
                 <v-text-field
+                    v-on:click:append="search"
+                    v-on:keydown="search"
```
knollfear, line 8: `                <v-text-field
                    v-on:click:append="search"
                    v-on:keydown.enter="search"
                    v-model="searchQuery"
`

solution/ui/prototype/src/views/ResourcePage.vue
```diff
@@ -58,13 +62,12 @@ export default {
         singleSupList: [],
         sortedSupList: [],
         preSelectedSections:[],
-        preSelectedParts: []
-
-
+        preSelectedParts: [],
+        searchQuery: "",
+        usingSearch: false,
```
knollfear, line 67: I don't like extra fields like this. if(searchQuery) is enough to know if search is being used and should always be the same as usingSearch without having to juggle another variable..

solution/ui/prototype/src/views/ResourcePage.vue
```diff
@@ -99,6 +102,24 @@ export default {
         }
     },
     methods: {
+        async search(event) {
+            if ((event.type === "keydown" && event.key === "Enter") || event.type === "click") {
```
knollfear, line 106: Unnecessary with changes suggested above.

knollfear (commented):
Looks good.  needs a bit of cleanup as suggested.  Also please merge in main as I think it has conflicts.

knollfear (approved):
LGTM!

PR #408, opened 2022-04-19
EREGCSC-1275 Test feasibility of parsing FR documents to enable granularity

cgodwin1:
Resolves #1275

**Description-**

Implementation of a parser in Go that fetches Federal Register documents related to a given list of parts, formats them, and sends them to eRegs for handling in our supplemental content system.

- New parser exists under `/solution/parser/fr-parser`
- Parser uses existing Parser Configuration manageable from within Django admin
- Parser extracts list of FR docs and sends them to `/v3/supplemental_content`
    - Note that this is a slow process right now and needs batching and optimizing (currently 1 FR doc per request)
- Parser attaches lists of relevant sections to each FR doc
- At least 90% test coverage

**The following improvements have been made since the proof of concept stage...**

- Parser does not attempt to fetch the full XML text if it doesn't exist (reduces network calls significantly)
- Unrecognized categories are created and populated on the Django side
- Parser skips documents that have already been fetched if "Skip versions" is set to true in the parser configuration (future work: should this field name be changed to something like "Skip processed"?)
- Django frontend pulls FR content from our supplemental content system instead of directly from the Federal Register API
- V3 rewrite for supplemental content (now called "resources"), implements the following changes:
    - New V3 endpoints (see `/solution/backend/resources/v3urls.py`), designed to be flexible and optimized with as few DB queries as possible per request.
    - Models are optimized for inheritance (see `/solution/backend/resources/models.py`)
    - DB queries optimized using `select_subclasses()`, `select_related()`, and `prefetch_related()`. This extends to all API calls and Django Admin pages.
    - New endpoints are fully documented on Swagger (see `/api/swagger` in your browser of choice.)

1. Migrate in this specific order:   (Our GHA deploys will take care of this automatically, this is only needed __one time__ locally!)
    - `docker-compose exec regulations python manage.py migrate regulations`    (renames database tables)
    - `docker-compose exec regulations python manage.py migrate resources`    (applies new V3 model migrations)
    - `docker-compose exec regulations python manage.py migrate`    (to make sure all other migrations are fully applied)
2. Verify you have some parts and/or subchapter(s) specified in Parser Configuration
3. From `/solution/`, run `make frdocs.local` (this will take some time, I suggest lowering the number of parts/subchapters processed if it's taking too long for you)
4. Check "Federal Register Documents" in Django admin to see newly uploaded FR docs
5. Review the new V3 resource endpoints in `/api/swagger`, and experiment with them to make sure their behavior lines up with the documentation.
6. Review V2 endpoints to make sure that they are still working. They had to be modified because of the changes to the models.


solution/parser/fr-parser/main.go
```diff
@@ -0,0 +1,183 @@
...
+	log.Debug("[main] Retrieving list of content for title ", title, " part ", part)
+	content, err := fetchContentFunc(ctx, title, part)
+	if err != nil {
+		return fmt.Errorf("Fetch content failed: %+v", err)
+	}
+
+	log.Debug("[main] Processing content for title ", title, " part ", part)
+	for _, c := range content {
+		if err := processDocumentFunc(ctx, title, part, c); err != nil {
+			log.Error("[main] Failed to process title ", title, " part ", part, " doc ID ", c.DocumentNumber, ": ", err)
```
knollfear, line 148: If this isn't raised as an error, should it be log.Warn?

knollfear (commented):
This works great.  Most comments are just about deployment details, not the functional code.  I am not approving just to keep it from merging into main.

solution/parser/fr-parser/main.go
cgodwin1: Only log.Fatal changes the return code. I see Warn as "something bad happened but it's not a big deal" and Error is "something failed but I can continue anyway".

solution/parser/fr-parser/main.go
```diff
@@ -0,0 +1,183 @@
...
+	ecfrEregs "github.com/cmsgov/cmcs-eregulations/ecfr-parser/eregs"
+	"github.com/cmsgov/cmcs-eregulations/ecfr-parser/ecfr"
+
+	"github.com/aws/aws-lambda-go/lambda"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// TIMELIMIT is the total amount of time the process has to run before being cancelled
+const TIMELIMIT = 5000 * time.Second
```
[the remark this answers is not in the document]
cgodwin1: I didn't pick this value. It was part of the eCFR parser and I just kept it. It's meant to be arbitrarily large but not infinite.

PR #413, opened 2022-04-28
1319 create unknown categories

knollfear:
Resolves #EREGCSC-1319

**Description-**
Creates a category when a piece of federal register content is uploaded that has a type that is not recognized.
 
- expected change 1

1. Check the number of categories you have (should be 10)
2. Run make frdata.local 
3. You should have 12 categories instead of the previous 10.


cgodwin1 (approved):
LGTM!

PR #415, opened 2022-04-29
updating UI to show users when quoted search might help.

knollfear:
Resolves #EREGCSC-1279

**Description-**

Showing users how they might be able to improve their results with quoted search

- added 2 variables to the search template context
- updated template to show desired words about quoted search

1. if a search has 2 or more words and is NOT quoted the page should show "Didn't find ..."  
2. If there are synonyms it should say "Or search..."
3. All other searches should be unchanged.


thwalker6 (approved):
Looks good

PR #420, opened 2022-05-16
remove sections and subparts related to part

thwalker6:
Resolves #
 EREGCSC-1350
**Description-**

Fixes logic to remove sparts and sections related to part when the part chip is removed on the resourcces page

- expected change 1

Select a part on the resources page.
Select a section or subpart.
Remove the part chip.
All filters related to the part should be cleared.


PhilR8:
If I perform the following flow:
1. Select part 431
2. Select part 431 subpart A
3. Select Part 433

The subparts dropdown updates to show subparts from part 433 but the sections dropdown does not update to show sections from part 433 - it still shows sections from part 431.

If I copy the URL and paste it into a new browser, the sections dropdown remains disabled: https://dcaci9wwks316.cloudfront.net/resources?title=42&part=431,433&subpart=431-A&q=

If I refresh and go through the flow again (432 > 432 Subpart A > 433) and then remove part 431, the sections dropdown still shows sections from part 431 even though that section has been removed.

thwalker6:
> If I perform the following flow:
> 
> 1. Select part 431
> 2. Select part 431 subpart A
> 3. Select Part 433
> 
> The subparts dropdown updates to show subparts from part 433 but the sections dropdown does not update to show sections from part 433 - it still shows sections from part 431.
> 
> If I copy the URL and paste it into a new browser, the sections dropdown remains disabled: https://dcaci9wwks316.cloudfront.net/resources?title=42&part=431,433&subpart=431-A&q=
> 
> If I refresh and go through the flow again (432 > 432 Subpart A > 433) and then remove part 431, the sections dropdown still shows sections from part 431 even though that section has been removed.

All part of a process in fixing all the bugs here.  To keep the PR's separate the only thing being fixed here really is the chips itself.    The two other stories should impact this.  The one for showing all sections and subparts on the filter, and the other to fix subheaders.

knollfear (approved):
LGTM as far as it goes.  I think there are some surrounding issues that make this harder to test, but I think this is in a good direction.

PR #421, opened 2022-05-17
Eregcsc 1339 change resource sidebar filter

knollfear:
Resolves #1339 and #1340

**Description-**

Adds a link at the top of each section and a button at the bottom that causes the right sidebar to filter content down to just that selected section.

Adds a link and button to sections to update the right sidebar
A link to revert to the subpart if a section has been chosen

Go to any subpart
click the link or button that says "View $ 123.45 resources (5)" and the right sidebar should update to the new results.  THis is an API call so it might take a second to finish up.
click on the new link above the content that says: "View All Subpart B Resources  (48)" and the state should return to the way it was before you clicked anything.


solution/ui/regulations/js/src/components/ViewResourcesLink.vue
```diff
@@ -0,0 +1,58 @@
...
+};
+</script>
+
+<style>
+  .btn{
+    background-color: #046791;
+    color: #fff;
+    border-radius: 3px;
+    padding: 9px 10px 10px;
+    border: none;
```
PhilR8, line 50: cursor: pointer;

solution/ui/regulations/js/src/components/ViewResourcesLink.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+    <div v-if="isLink()" style="padding-left: 5px; font-size: 12px; margin-top: -10px;">
+      <a v-if="this.count !== '0'" v-on:click="clickHandler">View {{this.section}} resources ({{ this.count }})</a>
+      <div v-else>No resources for {{this.section}}.</div>
```
PhilR8, line 4: for both the link and the button, the "View § 432.30 resources" part of the button label should be `font-weight: 600`.  The count in parentheses (`(1)`) can say at the default font weight.

Something like:
```
      <button class="btn" v-if="this.count !== '0'" v-on:click="clickHandler">
          <span class="bold">View {{this.section}} resources</span> ({{ this.count }})
      </button>
```

solution/ui/regulations/js/src/components/SupplementalContent.vue
```diff
@@ -1,4 +1,9 @@
 <template>
+  <div>
+    <a v-if="selectedPart" v-on:click="clearSection" style="color:#5B616B">View All Subpart {{subparts[0]}} Resources  ({{resourceCount}})</a>
```
PhilR8, line 3: This link should be `font-size: 12px`. Also, the "View All Subpart resources" part of the button label should be `font-weight: 600`. Can wrap in a span.

PhilR8:
LGTM!

PhilR8 (approved):

PR #422, opened 2022-05-18
Eregcsc-1331-1327

thwalker6:
Resolves #

1331, 1327
**Description-**

Fixed a bug where if multiple parts are selected on the subparts and sections will display for teh latest chronologically.  
Fixed a bug where sections with subheaders dont display properly as well as orphan sections.

- expected change 1

Select 2 seperate parts.

Check the drop down for subparts and sections.  Subparts and sections for both parts should be there


solution/ui/prototype/src/views/Resources.vue
```diff
@@ -257,11 +262,44 @@ export default {
...
+            return queryParams;
+        },
+        async getSectionsBySubpart(subpart) {
+            const splitSubpart = subpart.split("-");
+            const allSections = await getAllSections();
+            const sectionList = allSections
+                .filter((sec) => {
+                    return (
+                        sec.part === splitSubpart[0] &&
+                        sec.subpart == splitSubpart[1]
```
PhilR8, line 296: mixing equality comparison (`==`) with strict equality comparison (`===`) -- is this intentional?

PR #425, opened 2022-05-20
Fix: Resources deploy properly to dev/val environments on Prod deployment

PhilR8:
Resolves issue where Vue Router and Related Resources links do not work with dev/val URLs on deployment, e.g. https://qavc1ytrff.execute-api.us-east-1.amazonaws.com/dev and ./val

**Description**
When deploying eRegs to production, the production-ready code is also deployed to the `dev` and `val` environments.  The URLs for these deployments have "dev" and "val" subdirectories, which is not handled by the code and the Vue router does not work.  This means that the Resources page and the Cache Explorer do not load when visiting the `dev` and `val` URLs.

Mirroring the work down for deploying the experimental site, a `VITE_ENV` variable will be passed in to the Vue/Vite app and used where necessary for the router to direct the user to the proper route.

**This pull request changes**

- `VITE_PR` is renamed to `VITE_ENV` to reflect broader use than just to pass in a formatted string based on Pull Request number
- `VITE_ENV` will be used as the Vue router base and for the Related Resources links.  This will work for Experimental Deployments (where `VITE_ENV` will be `dev-XXX`) and for prod deployments will either be `dev` or `val` depending on the environment
- The `prod` environment variable will be ignored for now; the [CMS domain URL](https://regulations-pilot.cms.gov/) does not have subdirectories and `/` will be used as the Vue Router base.

1. Visit the Deploy Experimental site below, visit `/cache` and `/resources` to ensure they still work as expected
2. Merge this branch into main to trigger the prod deployment script and then test the `dev` and `val` URLs to see if `/resources` and `/cache` work as expected

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           STATIC_URL: http://localhost:8888/
           STATIC_ROOT: ../static-assets/regulations
-          VITE_PR: dev-${{ steps.findPr.outputs.pr }}
+          VITE_ENV: dev-${{ steps.findPr.outputs.pr }}
         run: |
           pushd solution
           make regulations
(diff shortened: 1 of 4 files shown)


knollfear (approved):
LGTM!

PR #427, opened 2022-05-25
Eregcsc 1339 change resource sidebar filter

knollfear:
Resolves #1340

**Description-**

Switches the endpoint for pulling resources into the right column to use the V3 API which is simply more performant.

- resources are pulled in via the v3 api

Go to any subpart or orphan page
resources should load, even if there are hundreds of them
click on View § XX.YYY resources (Z)  and verify the correct number of resources show.


PhilR8 (approved):
LGTM!

PR #434, opened 2022-06-02
EREGCSC-1367 -- Search and Resources Empty States

PhilR8:
Resolves [EREGCSC-1367](https://jiraent.cms.gov/browse/EREGCSC-1367)

**Description**

Search pages for Regulations Search and Resources Search will now have an intentionally designed empty state if there are no results.

**This pull request changes**

- Creates new SearchEmptyState vue component
- Builds SearchEmptyState component in both the `eregs-vite` pipeline and the rollup pipeline so it can be used in both Django templates (like the regulations search template) and in Vue SPAs (like the resources page)
   - This component does not use the Vuetify library; updating pipelines for creating a component to be used in both Django and Vue SPAs that uses Vuetify is still TBD
- Creates new storybook story for SearchEmptyState component
- Implements SearchEmptyState component in Regulations Search and Resources Search
- Fixes bug where Regulations search page would not load if no query parameter was provided (e.g. `localhost:8000/search/`)
- Updates deploy-experimental github action to include static assets like fonts and images to the deployed instance of Storybook
   - OpenSans fonts still need some work here but outside scope of this story
- Updates prod deploy github action to include static assets like fonts and images

The JIRA story has a very detailed Acceptance Criteria.  Work through the AC (reprinted below) and then compare each state to the mockups in the [Figma comp](https://www.figma.com/proto/TQS2hDz1GZEg1Z4ZoMg7fs/eRegs-Search-UI-Updates?page-id=197%3A546&node-id=197%3A659&viewport=-1709%2C912%2C0.31&scaling=min-zoom&starting-point-node-id=197%3A659&show-proto-sidebar=1).

> Regulations search results page in general (whether or not the results are 0):
> - Updated language for scope info above search box ("This site searches Title 42, Parts 400 and 430-460")

> Regulations search results, empty state only:
> - Modified language for results summary: "x results" becomes "x results in Medicaid & CHIP Regulations"
> - Addition of "Expand your search:" block, including links populated with query searched (as described in Figma annotations)
>    - If there is no search term query, links are not populated with query
>    - If there is a search term query, links are populated with query

> Resources search results in general (whether or not the results are 0):
> - No changes

> Resources search results, empty state only:
> - Modified language for resources summary: "x results" becomes "x Results in Resources"
> - Addition of "Expand your search:" block, including links (follows patterns described in Figma annotations for regulations search results page)
>    - If there is no search term query, links are not populated with query
>    - If there is a search term query, links are populated with query


solution/ui/regulations/eregs-vite/src/components/SearchEmptyState.stories.js
```diff
@@ -0,0 +1,21 @@
...
+
+const Template = (args, { argTypes }) => ({
+    props: Object.keys(argTypes),
+    components: { SearchEmptyState },
+    template: '<SearchEmptyState v-bind="$props" />',
+});
+
+export const Basic = Template.bind({});
+Basic.args = {
+    eregs_url: "https://www.google.com",
```
thwalker6, line 16: Should this be google?  Do we not have the proper url?

thwalker6 (commented):
Is google fine for the eregs_url?  I know its a .stories file so I'm just really asking.  Other than that it looks fine

solution/ui/regulations/eregs-vite/src/components/SearchEmptyState.stories.js
PhilR8: This URL can either be for the resources page or the regulations search.  What I should probably do is use a custom control to let storybook users simply pick between those two options.

knollfear (approved):
LGTM!

PR #443, opened 2022-06-10
Search by location and add chips

thwalker6:
Resolves # EREGSCSC-1321

**Description-**

We want to make it so that if you search a location ex. 400.200 you will update the filters with the appropriate chips

- expected change 1

Type in 400.200 into the search bar.  It should add the chips for part 400 and section 400.200

Clear the results.  Type in 400.200 hello.  No chips should be added but it will search on that as a normal query.

Type in 400.99.  No chips should be added since its an invalid section.  Will just query on 400.99

Type in 500.9.  Same behavior s above

It will work for any valid part/section combination.  It will check the part first to see if the part number is valid, then check the section number to see if it belongs to that part.

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/views/Resources.vue
+++ b/solution/ui/regulations/eregs-vite/src/views/Resources.vue
@@ -15,8 +15,9 @@
                         type="text"
                         class="search-field"
                         append-icon="mdi-magnify"
-                        hide-details
                         dense
+                        hint="Search resources using keywords or citations."
+                        persistent-hint
                         v-model="searchInputValue"
                         @click:append="executeSearch"
                         @click:clear="clearSearchQuery"
@@ -181,7 +182,7 @@ export default {
                     selectedIdentifier: this.searchInputValue.replace(".", "-"),
                     searchSection: true,
                 };
-                this.updateFilters(payload);
+                return this.updateFilters(payload);
             }
             this.$router.push({
                 name: "resources",
@@ -366,8 +367,8 @@ export default {
             const parts = dataQueryParams.part.split(",");
             const newPartDict = {};
 
-            parts.forEach((x) => {
-                newPartDict[x] = {
+            parts.forEach((part) => {
+                newPartDict[part] = {
                     title: "42",
                     sections: [],
                     subparts: [],
--- a/solution/ui/regulations/eregs-vite/src/views/Resources.vue
+++ b/solution/ui/regulations/eregs-vite/src/views/Resources.vue
@@ -21,7 +21,6 @@
                         @click:append="executeSearch"
                         @click:clear="clearSearchQuery"
                     />
-
                 </form>
             </ResourcesNav>
             <div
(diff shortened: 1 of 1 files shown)


solution/ui/regulations/eregs-vite/src/views/Resources.vue
```diff
@@ -142,7 +138,7 @@ export default {
             },
             supplementalContent: [],
             searchInputValue: "",
-            URL: import.meta.env.VITE_API_URL
+            URL: import.meta.env.VITE_API_URL,
```
knollfear, line 141: I think this is actually no longer used.

solution/ui/regulations/eregs-vite/src/views/Resources.vue
```diff
@@ -251,27 +258,66 @@ export default {
...
+                });
             }
-            this.$router.push({
-                name: "resources",
-                query: newQueryParams,
-            });
+        },
+
+        async checkPart(payload, scope) {
+            const partExist = this.filters.part.listItems.filter((part) => part.name == payload[0]).length > 0;
```
knollfear, line 314: .find seems apprpriate here.  you can just replace filter and then lose .length > 0.

solution/ui/regulations/eregs-vite/src/views/Resources.vue
```diff
@@ -251,27 +258,66 @@ export default {
...
+
+            if (partExist && scope === "section") {
+                const sectionList = await getSectionsForPart(42, payload[0]);
+                return (sectionList.filter((section) => section.identifier[1] == payload[1]).length > 0);
```
knollfear, line 318: find might work here too.

solution/ui/regulations/eregs-vite/src/views/Resources.vue
```diff
@@ -311,71 +363,65 @@ export default {
...
+            this.partDict = newPartDict;
 
             if (dataQueryParams.section) {
-
                 const sections = dataQueryParams.section
                     .split(",")
-                    .filter(x =>
-                        x.match(/^\d+/) && x.match(/\d+$/)
-                    )
+                    .filter((x) => x.match(/^\d+/) && x.match(/\d+$/))
```
knollfear, line 382: echoing Phil's request for a descriptive variable name.   What is this supposed to be here?

solution/ui/regulations/eregs-vite/src/views/Resources.vue
```diff
@@ -311,71 +363,65 @@ export default {
...
-                })
+                });
             }
             if (dataQueryParams.subpart) {
                 const subparts = dataQueryParams.subpart
                     .split(",")
-                    .filter(x =>
-                        x.match(/^\d+/) && x.match(/\w+$/)
-                    )
+                    .filter((x) => x.match(/^\d+/) && x.match(/\w+$/))
```
knollfear, line 396: another generic variable name.

knollfear (commented):
This looks good.  Just a few small code cleanups and we can get this merged in.

solution/ui/regulations/eregs-vite/src/views/Resources.vue
thwalker6: changed

solution/ui/regulations/eregs-vite/src/views/Resources.vue
thwalker6: undid the change but merging didnt remove this

PR #445, opened 2022-06-15
Prototype v3 sections

knollfear:
Resolves #

**Description-**

updates Part.vue in prototype site to use the V3 api

- removes calls that use old getAllParts endpoint because it is inconsistent.

Part page should still have subparts and sections populated as expected.


PhilR8 (approved):
LGTM!

PR #447, opened 2022-06-17
Feedback from page

thwalker6:
Resolves #

1374 Feedback

**Description-**

Adjusting text for subparts count and when an orphan section is selected.

Orpah sections should not link to subpart A resources and the subpart count was incorrect..

- expected change 1

From the homepage click Part 433.  Click section 433.1.
 No View All Subpart A Resources Linkn should appear in the right sidebar.

Go back.  Click a different section.  The correct count for subpart resources should be in the "View All Subpart {X} Resources" link in the right sidebar..

### DIFF
--- a/solution/ui/regulations/js/src/components/SupplementalContent.vue
+++ b/solution/ui/regulations/js/src/components/SupplementalContent.vue
@@ -1,7 +1,7 @@
 <template>
     <div>
         <a
-            v-if="selectedPart"
+            v-if="selectedPart && subparts.length ===1"
             v-on:click="clearSection"
             class="show-subpart-resources"
         >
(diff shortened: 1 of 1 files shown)


solution/ui/regulations/js/src/components/SupplementalContent.vue
```diff
@@ -37,8 +37,10 @@ import SupplementalContentCategory from "./SupplementalContentCategory.vue";
 import {
     getSupplementalContentByCategory,
     v3GetSupplementalContent,
+    getSubPartsForPart
 } from "../../api";
 import { EventCodes, formatResourceCategories } from "../../utils";
+import { getSubpartTOC } from '../../../eregs-vite/src/utilities/api';
```
knollfear, line 43: This is the wrong API file.  If it isn't in the API file referenced in line 41 above, you need to implement it there.  It is probably a small function and not much to copy over if it isn't there.  This is why the build is breaking though

knollfear:
closing to clean up the experimental deploy before changing experimental deploy process

PR #449, opened 2022-06-21
Eregcsc 1374, Update sidebar based on what they click

thwalker6:
Resolves #
1374 feedback
**Description-**

Hides "Show all subpart A resources link if an orphan section is selected.

Fixes the count of supplemental content in a subpart for "Show all subpart {X} resources" number.  Before number was incorrect.

- expected change 1

### DIFF
--- a/solution/ui/regulations/js/api.js
+++ b/solution/ui/regulations/js/api.js
@@ -397,6 +397,7 @@ const v3GetSupplementalContent = async (apiURL, {locations, locationDetails=fals
 
 }
 
+const getSubpartTOC = async (title, part, subPart) => httpApiGetV3(`title/${title}/part/${part}/version/latest/subpart/${subPart}/toc`)
 // API Functions Insertion Point (do not change this text, it is being used by hygen cli)
 
 export {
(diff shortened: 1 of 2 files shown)


solution/ui/regulations/js/api.js
```diff
@@ -397,7 +397,12 @@ const v3GetSupplementalContent = async (apiURL, {locations, locationDetails=fals
 
 }
 
-// API Functions Insertion Point (do not change this text, it is being used by hygen cli)
+const getSubpartTOC = async (apiURL, title, part, subPart) => 
+{
+    const url = apiURL.replace('/v2/', '/v3/')
+
+    return httpApiGetLegacy(`${url}title/${title}/part/${part}/version/latest/subpart/${subPart}/toc`)
+}// API Functions Insertion Point (do not change this text, it is being used by hygen cli)
```
knollfear, line 405: Can we please get a new line added here?

knollfear (commented):
I had an issue when I went to a subpart via a fragment link.  When I go to the subpart with a more normal path it works, but https:// . . . #A it doesn't work right.

solution/ui/regulations/js/api.js
thwalker6: done

thwalker6:
Fixed the #A not working.  The logic is that if it is not a number than will take the subpart and pull those resources for the subpart

knollfear (approved):
LGTM

PR #456, opened 2022-06-29
Category only Filtering bug

knollfear:
Resolves #

**Description-**
On the resources page if you select only a category the filtering does not happen.  A part is currently required to make category filtering work, which isn't right.  Someone should be able to filter for a category without choosing a part first.

- a small check for part Dict is skipped if not parts are selected.

1. Go to resources page, change the category.  On Prod nothing will happen, on this branch it should search for all resources that have that category.


thwalker6 (commented):
resources age wont load on server

thwalker6 (approved):

PhilR8 (approved):

PR #467, opened 2022-07-12
Left drawer navigation

knollfear:
Resolves #1394

**Description-**

Tightens up the margins between the title tabs and divider on the left nave bar.

- just a style tag

View the left nav bar in the expanded state and check the spacing between the tabs and separator.

suggest a couple minor changes: breadcrumbs should be non-bold and looks like they should be a touch bigger - Reduced font weight to 400 and increased size to 14px from 12px.

### DIFF
--- a/solution/ui/prototype/src/components/node_types/Section.vue
+++ b/solution/ui/prototype/src/components/node_types/Section.vue
@@ -1,37 +1,32 @@
 <template>
-    <section
-        :aria-labelledby="kebabTitle"
-        tabindex="-1"
-        :id="kebabTitle"
-        class="reg-section"
-    >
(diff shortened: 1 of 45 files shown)


PhilR8 (approved):
LGTM!

solution/ui/prototype/src/components/node_types/Section.vue
```diff
@@ -114,7 +114,8 @@ export default {
 };
 </script>
 
-<style>
+<style lang="scss">
```
PhilR8, line 117: I think this PR looks just fine.  I don't think we need `lang=scss` if we're not using any Sass functionality (e.g. nested selectors, Sass $variables), though.

PR #483, opened 2022-07-21
EREGCSC-1412 Modify API to serialize grouped Federal Register Documents

cgodwin1:
Resolves #1412

**Description-**

Updates the API to allow related Federal Register docs to be shown.

- Endpoints `/v3/resources` and `/v3/resources/federal_register_docs` show FR doc groups
- Related resources are managed automatically using post save hooks (see the bottom of resources/models.py)

1. Visit the above endpoints and verify that resources of type "federal_register_doc" have a `related_docs` field, and it is not empty for at least some. It will be empty for many as well.
2. If verifying locally, use `?html_api` parameter and verify that the number of database queries is static (i.e. not increasing with the number of results on the page)


knollfear (approved):
LGTM.  Tested locally and works as expected too.  Nice job!

PR #486, opened 2022-07-21
EREGCSC-1376 Adding in sub categories when a parent category is chosen.

knollfear:
Resolves #EREGCSC-1376

**Description-**

When selecting a parent category the subcategories should be selected too.  This is like how subpart selection works.

- Adds subcategory chips when parent category is selected

Add a parent category, see subcateogry chips get added
select a duplicate category to add, confirm it doesn't duplicate
Add another top level category and see many subcategories get added
Any other things you feel like testing

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/views/Resources.vue
+++ b/solution/ui/regulations/eregs-vite/src/views/Resources.vue
@@ -575,7 +575,7 @@ export default {
                 .filter((item) => item.type === "category")
                 .reduce((acc, item) => {
                     acc[item.name] = item;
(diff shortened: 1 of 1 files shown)


thwalker6:
So not that it is necessarily an issue but it is weird behavior( a bit by design).  If I select a parent category and get rid of all the children I can have zero results with a parent category.  Should proper behavior be that if 1 child is removed the chip for the parent also goes away?  The reason we dont do this with the subpart/section chips is because of how supplemental content is located but with categories its just completely different

thwalker6 (approved):
g2g

PR #489, opened 2022-07-22
EREGCSC-1329 -- Sort resources by date

PhilR8:
Resolves [EREGCSC-1329](https://jiraent.cms.gov/browse/EREGCSC-1329)

**Description**

Adds ability for users to sort results of a QUERY STRING search by "Relevance".  Default sorting of results is "Date (Newest)".

**This pull request changes**

- Updates Django `ResourceExplorerViewSetMixin` mixin to accept `sort` method.  Returns resources sorted by `relevance` if sort method of `relevance` is passed in with request along with a text `search_query`.
- Adds "Sort by" select to front end of resources page
- Updates URL query params when "Sort by" select changes
- "Sort by" select only active when a search query is used to get results; the select is disabled and locked to "Date (Newest)" if search query is not used

1. Go to Swagger page and view `/v3/resources/`.  Note that there is now a `sort` parameter with a description of what strings can be used.  Try it out with and without a `q` parameter.  Without a `q` parameter, results should always be returned sorted with newest first.  With a `q` parameter and with `sort` set to `relevance`, response order should be different than if `sort` is set to `newest`.
2. Visit Resources page.  Note that "Sort by" select is disabled to begin.
3. Enter a search query and execute search.  Note that the "Sort by" select should become active and that it is set to "Date (Newest)". Note that the URL has a query param for "sort."
4. Change the "Sort by" select to "Relevance" and make sure the response order changes.  Compare to swagger response above if desired.
5. Clear search query and note that the "Sort by" select should become disabled and return to "Date (Newest)".


solution/backend/resources/v3views.py
```diff
@@ -231,6 +231,8 @@ class ResourceExplorerViewSetMixin(OptionalPaginationMixin, LocationFiltererMixi
                               "search with quotes.", str, False),
         OpenApiQueryParameter("categories", "Limit results to only resources found within these categories. Use "
                               "\"&categories=X&categories=Y\" for multiple.", int, False),
+        OpenApiQueryParameter("sort", "Sort results by this field. Valid values are \"newest\", \"oldest\", and \"relevance\". "
```
cgodwin1, line 234: Seeing "newest" and "oldest" in a couple of places, but the queryset only has `if X sort by relevance, else sort by date_annotated`

cgodwin1 (changes requested):

solution/backend/resources/v3views.py
PhilR8: reference to "oldest" has been removed until it is implemented in a future story.

PR #497, opened 2022-08-04
EREGCSC-1447 -- Enable "sort by oldest" for resources search page

PhilR8:
Resolves [EREGCSC-1447](https://jiraent.cms.gov/browse/EREGCSC-1447)

**Description**
- null dates go to the end of the list
- newest first is the default when sorting by date, can choose oldest
- copy would be "Date (newest)" vs "Date (oldest)"

**This pull request changes**
- Updates `resources/v3views` to accept new `oldest` sort method
- Sort select on Resources page is always active now, as any results can be sorted by newest/oldest
- If there is no text `searchQuery`, the "Relevance" search option is disabled.  It becomes enabled when a search query is also used as a search paramter.

1. On the resources page, I can search for a term like COVID and sort by oldest, and I see documents from 2020 at the top


cgodwin1 (approved):
LGTM

PR #507, opened 2022-08-12
Homepage fix

knollfear:
Resolves  a problem when deploying an experimental site.

**Description-**

For a while when deploying an experimental site the first run of the parser would fail and then on subsequent runs succeed.  This was caused by an error in the seed data that led to 2 duplicate locations.  These have been removed and experimental deploys should look much better.  

- removes 2 duplicate sections from the seed data
- links 2 FR docs to the appropriate resources
- swaps out a structure built from parts to using a TOC value from the Title

1. Verify that the go parser ran on the first try.  In this PR deploying the reg site failed, but the parser has never failed
2. Sections 438 and 441 should exist on the home page
3. The homepage should be organized in the same way as production and not broken up into multiple sections
4. Section 438 and 441 should look normal in the reader view


solution/backend/regulations/views/homepage.py
```diff
@@ -24,7 +23,7 @@ def get_context_data(self, **kwargs):
         if not parts:
             return context
 
-        full_structure = get_structure(parts)
+        full_structure = Title.objects.all().values_list("toc", flat=True)
```
thwalker6, line 26: Not a problem I see we are using teh get_structure only in one other location.  Are we essentially just using this instead of what you did for the TOC?  Can we just do the same there with no issue.  Just a code clean up thing.  Wouldnt include it here but Just a small story for next sprint

thwalker6 (approved):

thwalker6:
Looks good.  LGTM

PR #512, opened 2022-08-16
EREGCSC-1359 Correct FR parser handling of non-Title-42 SECTNOs

cgodwin1:
Resolves #1359

**Description-**

When the FR parser encounters a doc that exists for more than one title, it marks all extracted sections (from SECTNO tags in the XML) as belonging to the first title it processes that contains that doc. Thus the sections once uploaded to eRegs are invalid, e.g. 42 CFR 123.45 actually belongs to title 45.

- The FR parser processes `<CFR>` tags in the XML to map sections listed in `<SECTNO>` tags to the proper title.
- Sections that belong to titles outside the scope of our parsing are ignored.
- Unit tests are updated accordingly.

1. From a clean database, run the FR parser with the default title options.
2. In the Django Admin panel, go to "Federal Register Documents" and search for "84 FR 7610".
3. Navigate to the Locations field for that doc, and verify that there is exactly one section that is marked as belonging to title 45, and the rest are title 42. (Before this fix, all locations would belong to title 42.)
4. Optional: check other FR docs to make sure their locations are correct.

### DIFF
--- a/solution/parser/ecfr-parser/main.go
+++ b/solution/parser/ecfr-parser/main.go
@@ -60,7 +60,7 @@ func getLogLevel(l string) log.Level {
 	case "trace":
 		return log.TraceLevel
 	default:
-		log.Warn("[main] \"", config.LogLevel, "\" is an invalid log level, defaulting to \"warn\".")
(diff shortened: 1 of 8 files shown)


knollfear (approved):
LGTM!

PR #513, opened 2022-08-17
EREGCSC-1452 -- Remove View Resources buttons from Print View and Mobile View

PhilR8:
Resolves [EREGCSC-1452](https://jiraent.cms.gov/browse/EREGCSC-1452)

**Description:**
The View Section Resources button has been added at the start and end of each section.  However, these buttons were still visible in the print view and they should instead be hidden in the print view.

Additionally, these buttons should also be hidden in mobile widths, which for our purposes is 767px and narrower.

- Adds a class to the `view-resources-link` btns so they can be styled and queried in e2e tests
- adds that class to the style rule to be hidden in the print view in `_print.scss`
- adds that class to the style rule to be hidden in narrow, mobile widths in `_buttons_and_tooltip.scss`
- Updates e2e tests to ensure buttons are shown in normal, desktop widths and hidden in narrow widths or in the print view

1. Visit a subpart on the experimental deployment (ex: https://5te8lq0ox9.execute-api.us-east-1.amazonaws.com/dev513/42/433/Subpart-A/2021-03-01/)
2. Click the "print" button at the top of the right sidebar
3. the print preview should not show the "View XXX Resources" buttons at the top and bottom of each section
   - compare to print preview on prod and notice that those buttons DO appear on the print preview
4. Click out of the print preview and make your browser window narrower than 767px
5. the "View XXX Resources" buttons should no longer be visible
6. the entire right resources sidebar should no longer be visible


solution/ui/regulations/js/src/components/ViewResourcesLink.vue
```diff
@@ -1,19 +1,36 @@
 <template>
-    <div v-if="isLink()" style="padding-left: 5px; font-size: 12px; margin-top: -10px;">
-      <a v-if="this.count !== '0'" v-on:click="clickHandler">
-        <span class="bold">View {{this.section}} resources</span> ({{ this.count }})
-      </a>
-      <div v-else class="bold disabled" >No resources for {{this.section}}.</div>
+    <div
+        v-if="isLink()"
+        class="view-resources-link"
+        style="padding-left: 5px; font-size: 12px; margin-top: -10px"
```
thwalker6, line 5: The style for the v-if and the v-else are the same minus the margin-top: -10 px.  It would probably be better to add this to the same class and then add the style of just margin-top: -10px in case we want to change something.

thwalker6 (changes requested):

thwalker6 (approved):

thwalker6:
LGTM

PR #517, opened 2022-08-19
EREGCSC-1487 -- Resources with keyword search query parameter do not load when link pasted into new tab

PhilR8:
Resolves [EREGCSC-1487](https://jiraent.cms.gov/browse/EREGCSC-1487)

**Description**

If you copy and paste a Resources link with a keyword search query parameter into a new tab, resources do not load and the user sees the empty state.

Example: https://regulations-pilot.cms.gov/resources/?q=clinic&sort=newest 

The above page will return zero results if opened in a new tab.  It should return 24 results.

- Removes computed prop `multiWordQuery` being used as a method

1. Visit resources page in experimental deployment here: https://nlhvn5mtui.execute-api.us-east-1.amazonaws.com/dev517/resources
2. Enter "clinic" into search input and press enter or click the magnifying glass
3. Copy URL and open in new tab
4. You should see results on the page


solution/ui/regulations/eregs-vite/src/views/Resources.vue
```diff
@@ -729,7 +731,6 @@ export default {
         if (this.queryParams.q) {
             this.searchQuery = this.queryParams.q;
             this.synonyms = await this.retrieveSynonyms(this.queryParams.q)
-            this.multiWordQuery()
```
PhilR8, line 732: this is a computed prop, not a method.  It should update reactively as its dependency (`this.searchQuery`) changes.

thwalker6 (approved):
LGTM

PR #518, opened 2022-08-19
EREGCSC-1479 Improve anti-indexing methods

cgodwin1:
Resolves #1479

**Description-**

Further helps to prevent eRegs from showing up in search engine results.

- When search indexing is disabled via the admin panel:
    - Google & Bing are allowed to crawl the site, but all pages contain "noindex" tags so any existing data will be dropped.
    - All other robots are disallowed from crawling.
- Verified all pages are correctly serving noindex tags.

1. With indexing disabled via the admin panel, visit /robots.txt and note that Google and Bing are allowed to crawl, but other robots are not.
2. View the source of pages on the site and note the `<meta name="robots" content="noindex" />` tag within the `<head>` of the page.

### DIFF
--- a/solution/backend/regcore/templates/add_synonym.html
+++ b/solution/backend/regcore/templates/add_synonym.html
@@ -3,6 +3,7 @@
 <head>
     <meta charset="UTF-8">
     <title>Bulk Synonym Import</title>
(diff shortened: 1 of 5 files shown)


PhilR8 (approved):
LGTM!

PR #519, opened 2022-08-22
Disable resources test suite

PhilR8:
**Description**

Resources test suite is temperamental and does not pass reliably.

This PR skips resources test suite until it can be looked at closer and debugged.

**This pull request changes**

- adds `skip` keyword to resources test suite so the test runner skips it.

1. Look in deploy action for this PR and see if the resources tests were skipped
2. Pull down branch locally and run `make test` and see if resources tests were skipped


cgodwin1 (approved):
LGTM

PR #520, opened 2022-08-22
EREGCSC-1479-B Add middleware for managing X-Robots-Tag

cgodwin1:
Resolves #1479 again

**Description-**

In addition to the other indexing control methods, there is a HTTP response header we can add to make sure "noindex" is present on all pages and other files when requested.

- "NoIndex" middleware exists which, if indexing is disallowed, inserts `X-Robots-Tag = noindex` as an HTTP response header.
- Consolidated middleware into one file per app (middleware.py) for simplicity

1. If eregs is already running, run `docker-compose restart regulations` to reload settings.py
2. Locally, disable indexing from the "Site Configuration" admin panel.
3. Visit any eRegs page locally.
4. Click "Headers" on Django Debug Toolbar, and verify that under "Response headers", `X-Robots-Tag` is set to `noindex`.
5. Allow indexing from the admin panel and verify that the response header is now missing.


PhilR8 (approved):
LGTM!

PR #523, opened 2022-08-26
EREGCSC-1467 eCFR parser retry functionality causes duplicate data in a part

cgodwin1:
Resolves #1467

**Description-**

This PR fixes several parser issues:

- Original parser retry functionality causes duplicate data in a part object.
    - For example, the name of a part could be "438.1.438.1.438.1".
    - Solved by replacing the original retry functionality.
- Fixes data inconsistency on the eRegs side when one or more network calls fail, e.g. if the parser succeeds in posting a Part object, but fails to upload the Title object, it won't show up on the TOC even though it exists on the site. (See #534 for more details.)

- Old eCFR parser retry functionality is removed.
    - Replaced by new retry functionality that uses the existing versions endpoint to simplify the logic.
    - See #543 for more details.
- Parser posts part content/structure along with resource locations and TOC info all in one request.
- All POST/PUT requests are now _atomic_, so if any portion of a request fails to upload or fails to be processed server-side, the database is rolled back automatically and no part of the database is left in an inconsistent or out-of-sync state. (All or nothing approach.)
- V3 views and serializers in regcore are now split into smaller, more manageable files.
- The Title model in regcore created during previous V3 work is now removed in favor of manipulating Part instances instead, to reduce system complexity and interdependencies within the data.
    - The `/v3/title/<title>` endpoint is removed as well, since we no longer have or need that data.
- The `name` and `title` fields in the Part model are now `IntegerField`s instead of `CharField`s. This enables simpler sorting and is more consistent with the rest of our models.
- New V3 endpoint, `/v3/part` for uploading part + location data together. (Location data is optional.)
- In V3 views and serializers, "Contents" is renamed to "TOC" for clarity.
- Existing versions endpoint migrated to `/v3/title/<title>/versions`

1. After applying migrations, run the parser as normal and verify that it works properly
2. Run tests (`cd /dir/to/ecfr-parser/; go test ./...`)
3. Visit the homepage and verify that the table of contents is showing all parts and the structure is correct
4. Optionally, to verify new retry functionality: add `raise Exception("test")` before line 47 of `solution/backend/regcore/v3views/parser.py` and run the parser again.


PhilR8 (approved):
Verified manually and works as described.  LGTM!

peggles2 (approved):
LGTM 👍

PR #526, opened 2022-08-31
EREGCSC-1470 -- Only show last updated dates when that date has meaning

PhilR8:
Resolves [EREGCSC-1470](https://jiraent.cms.gov/browse/EREGCSC-1470)

**Description**

eCFR did not track individual changes to parts prior to January 1st, 2017.  

If a part has not been updated since eCFR began tracking individual changes, it shows a last amended date of January 1st, 2017, even though it was not amended on that date.  The actual last amended date is unknown.

Therefore, if a part has a last updated date of January 1st, 2017, it is not a meaningful date and that date should not be shown on the landing or reader view pages.

**This pull request changes**

- adds `has_meaningful_latest_version_date` field to context for `reader` templates and `regulation_landing` template
- Only shows `latest-version` span in those templates if `has_meaningful_latest_version_date` is true

1. Visit Parts 434 and 436 in two places:
   - [the "landing" view](https://2wkkx2bko5.execute-api.us-east-1.amazonaws.com/dev526/42/434/#434): the subpart table of context, which is where you are sent when you use the "Jump To" functionality
   - [the "reader" view](https://2wkkx2bko5.execute-api.us-east-1.amazonaws.com/dev526/42/434/Subpart-A/2017-01-01/), which is the actual regulations at the supart level
2. Ensure that these part pages do not have a "last amended date" at the top of the page.  You should only see "Part 436 up to date from eCFR as of Aug 31, 2022."
3. Visit any other part ([Part 435](https://2wkkx2bko5.execute-api.us-east-1.amazonaws.com/dev526/42/435/#435), for example) in both landing and reader view
4. Ensure that it has an "up to date" date and a "last amended" date, since it has been amended since January 1st, 2017


cgodwin1 (approved):
LGTM!

PR #527, opened 2022-09-01
Only seed experimental deploy databases once

cgodwin1:
Resolves #n/a

**Description-**

Currently every time you commit and push to an open PR, the experimental deploy attempts to reload the fixtures.

To more closely match the behavior of local environments and production, and fix fixture import bugs (when IDs differ causing unique constraints to be violated), we should only load fixtures automatically once per deploy, on first creation.

The `populate_content` function checks the count of each model with a fixture. If the count is 0, we can safely load the fixture. Otherwise it is skipped.

1. Check that this PR deploys correctly after the first and second push.

### DIFF
--- a/solution/backend/populate_content.py
+++ b/solution/backend/populate_content.py
@@ -3,20 +3,6 @@ import os
 from django.core.management import call_command
 
 
-from resources.models import (
-    Category,
-    SubCategory,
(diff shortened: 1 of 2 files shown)


thwalker6 (approved):
LGTM

PR #534, opened 2022-09-08
EREGCSC-1493 Network failures sometimes cause eCFR data to become out-of-sync when parser runs

cgodwin1:
Resolves #1493

**Description-**

The eCFR parser currently performs several types of requests to the eRegs server, namely: requesting configuration, enumerating existing versions, posting part content/structure, posting resource locations, and finally, posting an updated table of contents. Because these network calls are independent, if one fails, some critical data may be missing, until the next parser run or forever until a run from scratch.

This PR makes it so that all data relating to a reg part is contained in one network call per version of a part.

- Parser posts part content/structure along with resource locations and TOC info all in one request.
- All POST/PUT requests are now _atomic_, so if any portion of a request fails to upload or fails to be processed server-side, the database is rolled back automatically and no part of the database is left in an inconsistent or out-of-sync state. (All or nothing approach.)
- V3 views and serializers in regcore are now split into smaller, more manageable files.
- The Title model in regcore created during previous V3 work is now removed in favor of manipulating Part instances instead, to reduce system complexity and interdependencies within the data.
    - The `/v3/title/<title>` endpoint is removed as well, since we no longer have or need that data.
- The `name` and `title` fields in the Part model are now `IntegerField`s instead of `CharField`s. This enables simpler sorting and is more consistent with the rest of our models.
- New V3 endpoint, `/v3/part` for uploading part + location data together. (Location data is optional.)
- In V3 views and serializers, "Contents" is renamed to "TOC" for clarity.
- Existing versions endpoint migrated to `/v3/title/<title>/versions`

To verify this PR, I recommend starting with an existing database with regulation content loaded already. This is optional, but it will test the migrations as well.

1. Run migrations
2. Run the eCFR parser and make sure that it completes successfully
3. Visit the homepage and verify that the table of contents is showing all parts and the structure is correct
4. Visit the admin panel and make sure that sections are uploading as expected

The bulk of this work will not be able to be validated until some time has passed. Network calls do occasionally fail, so we will have to see this occur and verify that no unintended side effects of this are caused.


thwalker6 (approved):
LGTM

PR #535, opened 2022-09-09
EREGCSC-1458 Resource Download Button

thwalker6:
Resolves #

1458
**Description-**

Creates a download all resources button for the resources page

- expected change 1
A button wil appear on the resources page to download all resources in a csv

Go to the search resources page.

Search for resources with filter and or a query or neither

Hit the button.

All resources will be downloaded in a csv that are searched for.


solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -0,0 +1,121 @@
...
+                responseContent = responseContent.concat(content.results)
+
+            }
+            return responseContent
+        },
+    },
+}
+</script>
+<style scoped>
+.csv-button {
```
PhilR8, line 104: Is this `csv-button` class being used anywhere?

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -0,0 +1,121 @@
+<template>
+    <div class="resource_btn_container">
+        <button class="default-btn action-btn search_resource_btn" @click="createCSV">
```
PhilR8, line 3: the button needs to be the same height as the sort selectbox.  Right now it's too tall: 
Screen Shot 2022-09-09 at 16 05 48

changing the `action-btn` top and bottom padding to `8px` should do the trick.

PhilR8: we also need to handle narrow screen widths.  During planning we'd talked about just changing the button label to be "CSV" like this:

Screen Shot 2022-09-09 at 16 09 56

We should also add an `aria-label` to this button with something like "Download Spreadsheet as CSV" when the label is shortened.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -0,0 +1,121 @@
...
+        categories: {
+            type: Array,
+            default:() => []
+        },
+        supCount: {
+            type: Number,
+            default: 0
+        }
+    },
+    data: {
```
PhilR8, line 42: data needs to be a function with an explicit return.  You can use the method shorthand and write it like 

```
    data() {
        return {
            supplementalContent: []
        }
    },
```

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -0,0 +1,121 @@
...
+            }
+            return locString
+        },
+        async getSupplementalContent() {
+            let supNum = 0;
+            let page = 1;
+            let content = []
+            let responseContent = []
+            while (supNum < this.supCount) {
+                content = getSupplementalContentV3({
```
PhilR8, line 86: this is not currently working locally or on the experimental deployment

PhilR8 (changes requested):
The download functionality isn't working locally or on the experimental deployment.

I've also added some minor styling and breakpoint issues to be addressed.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
PhilR8: We should also add an `aria-label` to this button with something like "Download Spreadsheet as CSV" when the label is shortened.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
thwalker6: removed

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
thwalker6: changed

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
thwalker6: fixed

PhilR8 (approved):
These changes look good and the download functionality appears to work well!  Nice work.  LGTM!

PR #536, opened 2022-09-09
EREGCSC-1482 -- Synthetic part

PhilR8:
Resolves [EREGCSC-1482](https://jiraent.cms.gov/browse/EREGCSC-1482)

**Description**
Creates JS file of responses to mock a synthetic part of the regs that includes all edge cases.  Adds these JSON responses as mocks to be used with storybook stories.

- Adds mock service worker (MSW) to storybook 
   - this fixes issues with Supplemental Content stories not working
- creates JS file of synthetic part that matches a response from V3 of the API
- Adds Synthetic Part (as Part 499) to Storybook under the "PROTOTYPE > Regulation Text" heading

> **Note**
> The MSW works locally but does not yet work on a hosted instance of Storybook. That is TODO with another story

1. Go to [hosted version of Storybook](https://d3qvw8g3pq7moe.cloudfront.net/storybook/index.html) for this PR
2. In the left sidebar, go to PROTOTYPE > Regulation Text
3. There should be two entries:
   - Nodes
   - Synthetic Part
4. Select Synthetic Part > Part 499 to see the entire Synthetic Part styled as expected (i.e. eRegs styling)

If you'd like to see the Mock Service Worker in action:
1. Pull down this branch locally and run `make storybook`.  Make sure the local Django dev server is not running.
2. Visit `localhost:6006` and open the DevTools console
3. Look through the noise to find `[MSW] Mocking enabled.` in red text
4. In the left sidebar, go to SUPPLEMENTAL RESOURCES > Supplemental Content > Basic
5. You should see all of the categories become active and filled with supplemental content.
6. Inspect the Network tab and notice that a call to `localhost:8000` was successful and returned 66 results even though `localhost:8000` is not currently active


solution/ui/regulations/js/src/components/SupplementalContent.vue
```diff
@@ -229,6 +229,7 @@ export default {
                         );
                 } else {
                     await this.get_location_string()
+                    console.log("this.joined_locations", this.joined_locations);
```
thwalker6, line 232: Do we want this console.log in here?  Doesnt matter either way to me, just wondering if you forgot it or something.

thwalker6 (approved):
Looks good.  Small question though.  You put the mock responses in a .js file but would it make more sense to be a json file.  I'm not sure how the mocks work as in if they need to be a .js file or not.?  It's really not a big deal at all, just more of a general question.

PhilR8:
> Looks good. Small question though. You put the mock responses in a .js file but would it make more sense to be a json file. I'm not sure how the mocks work as in if they need to be a .js file or not.? It's really not a big deal at all, just more of a general question.

I think it's a matter of opinion but I like exporting from a JS file.  I've chopped up the `fullPart` response a bit so that things like `citations` and `tables` are broken out into their own objects that can be used in the `fullPart` response, which also gives us the ability to just import `citations` into a `Citations` storybook story without having to parse an entire JSON file.

And I'm looking ahead a little bit for eventual front end unit/snapshot tests with Jest - these mock data will be easy to re-use for those tests.  And also for the eventual cypress integration tests that will be needed for the reader view if we ever convert it to a Single Page App like the resources page.

TL;DR: I think using a JS file gives us much more flexibility.

solution/ui/regulations/js/src/components/SupplementalContent.vue
PhilR8: whoops, lemme get rid of this.

PR #537, opened 2022-09-13
EREGCSC-1501 Fix issue with resource page

thwalker6:
Resolves #
EREGCSC-101
**Description-**

Fixes issue where FRDOCS got grouped together and would replace the linked documents with the most recent one.

- expected change 1

When you sort supplemental content by oldest it should all be in order again

Go to the resources page.  Sort by oldest.  Within the first 20 they should all be in order.

Search for "36443".  This should pull up the FR doc realated to 36443.  Before it would pull up 33209


PhilR8 (approved):
This works well! LGTM!

PR #542, opened 2022-09-14
EREGCSC-1500 Add view resource button to pages

thwalker6:
Resolves #
EREGCSC-1500

**Description-**

Adds the view resources button to all pages except for  the resources ppage

- expected change 1

Go the following pages and see if the "View Resources" Button appears in the toolbar
 "/" home page
"/about"
"/search"

Go to "/resources" and verify that the button is not there

### DIFF
--- a/solution/backend/regulations/templates/regulations/partials/header/reg-title.html
+++ b/solution/backend/regulations/templates/regulations/partials/header/reg-title.html
@@ -1,11 +1,9 @@
 {% load static %}
 
 
-    <div class="reg-title">
+    <div class="reg-title {% if part %}part{% endif %}">
         <div class="flex-container">
             <div class="flex-header">
-
-                {{static}}
                 {% if part %}
                 <a href="{% url 'regulation_landing_view' title reg_part %}">
                     <strong>{{title}} CFR Part {{ part }}</strong>
--- a/solution/backend/regulations/templates/regulations/partials/header/reg-title.html
+++ b/solution/backend/regulations/templates/regulations/partials/header/reg-title.html
@@ -1,9 +1,12 @@
 {% load static %}
 
-{% if part %}
+
     <div class="reg-title">
         <div class="flex-container">
             <div class="flex-header">
+
+                {{static}}
+                {% if part %}
                 <a href="{% url 'regulation_landing_view' title reg_part %}">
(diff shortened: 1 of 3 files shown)


solution/backend/regulations/templates/regulations/partials/header/reg-title.html
```diff
@@ -1,9 +1,12 @@
 {% load static %}
 
-{% if part %}
+
     <div class="reg-title">
```
PhilR8, line 4: Having `reg-title` on every page is a fine solution to this, but it adds a gray left border to the header:

Screen Shot 2022-09-15 at 09 12 47

That border is generally meant to separate the site title ("Medicaid & CHIP eRegulations") and the reg title on the Part TOC/Reader view pages (ex: "42 CFR Part 433").  It's there's no reg title, the border should be hidden.

My vote would be to hide it on the homepage/about page/search results page, but you might ask the designers if it's okay to leave it here.

solution/ui/regulations/css/scss/partials/_header.scss
```diff
@@ -308,7 +308,6 @@ header {
         }
 
         .reg-title:after {
-            content: "Search Results";
```
PhilR8, line 311: was this left over from a long forgotten version of the header?

PhilR8 (changes requested):
There are some python linter issues that should be corrected, as well as a small styling issue that could go either way.

solution/ui/regulations/css/scss/partials/_header.scss
thwalker6: Thats what I was thinking.  If you go to that page it actually doesnt appear anywhere but with my changes it did appear.  Befoer when we did the if part on the django template, the .reg-title was hidden.

PhilR8: okay nice.  deleting stale code is always good!

solution/backend/regulations/templates/regulations/partials/header/reg-title.html
thwalker6: Added some django/css logic to make it appear if part appears

PhilR8 (approved):
LGTM!

PR #546, opened 2022-09-19
Eregcsc 1507

thwalker6:
Resolves #

EREGCSC-1507  Missing FR Docs in sidebar

**Description-**

Some FR-DOCS are missing on the sidebar.  This will allow them to appear if not there.

If there are more FRDOCS than the limit it will show a show more button allowing more federal documetns to appear.  Before there was a hardcoded 10 in there with a limit of 5.  So if it was less than 10 FR docs but more than5 you had no way of showing them.  Threw in the bonus of it not working in supplemental content as well :D.

Go to title 42 part 400 section 200.  Before you were missing a bunch of them such as Medicare Program; Hospital Insurance Entitlement and Benefits.  you should see a show more button now with the display and find the fr group in there.

### DIFF
--- a/solution/ui/regulations/js/src/components/RelatedRuleList.vue
+++ b/solution/ui/regulations/js/src/components/RelatedRuleList.vue
@@ -27,7 +27,7 @@
             </template>
         </template>
         <collapse-button
-            v-if="showMoreNeeded && rulesCount > limit"
+            v-if="showMoreNeeded"
             :name="innerName"
             state="collapsed"
             class="category-title"
@@ -81,7 +81,7 @@
                 </template>
             </template>
             <collapse-button
-                v-if="showMoreNeeded && rulesCount > 0"
+                v-if="showMoreNeeded && rulesCount > 10"
                 :name="innerName"
                 state="collapsed"
                 class="category-title"
--- a/solution/ui/regulations/js/src/components/RelatedRuleList.vue
+++ b/solution/ui/regulations/js/src/components/RelatedRuleList.vue
@@ -27,7 +27,7 @@
             </template>
         </template>
         <collapse-button
-            v-if="showMoreNeeded && rulesCount > 10"
+            v-if="showMoreNeeded && rulesCount > limit"
             :name="innerName"
             state="collapsed"
             class="category-title"
--- a/solution/ui/regulations/js/src/components/SupplementalContentList.vue
+++ b/solution/ui/regulations/js/src/components/SupplementalContentList.vue
@@ -47,7 +47,7 @@
             >
             </supplemental-content-object>
             <collapse-button
-                v-if="showMoreNeeded && contentCount > 10"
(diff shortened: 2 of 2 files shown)


solution/ui/regulations/js/src/components/SupplementalContentList.vue
```diff
@@ -47,7 +47,7 @@
             >
             </supplemental-content-object>
             <collapse-button
-                v-if="showMoreNeeded && contentCount > 10"
+                v-if="showMoreNeeded && contentCount > limit"
```
PhilR8, line 50: It shouldn't be changed here because it is working as designed.  See Subregulatory Guidance > Frequently Asked Questions (FAQs) here:

https://regulations-pilot.cms.gov/42/435/Subpart-B/2019-05-06/

If there are more than `limit` items (in this case five) but fewer than 10, then this button should be hidden.  This was discussed in a long validation thread you can read here:

https://fearless.slack.com/archives/C02DL6UU076/p1636723469000900

The same location on your branch [here](https://dd7ssdpi8b.execute-api.us-east-1.amazonaws.com/dev546/42/435/Subpart-B/2019-05-06/) shows two "show less" buttons (one after the fifth item and one at the end of the list) when there should only be one (after the fifth item).

solution/ui/regulations/js/src/components/RelatedRuleList.vue
```diff
@@ -27,7 +27,7 @@
             </template>
         </template>
         <collapse-button
-            v-if="showMoreNeeded && rulesCount > 10"
```
PhilR8, line 30: There definitely is a bug in this component though.  It should mirror the behavior of the collapse buttons in `SupplementalContentList` below.

So that would mean having the `v-if` directive condition in this first `collapse-button` to be only `showMoreNeeded`:

`v-if="showMoreNeeded"`

and the `v-if` directive of the second `collapse-button` below on line 84 be:

`v-if="showMoreNeeded && rulesCount > 10"`

This will put both components in sync where they only show the "middle" Show More button if there are between five and ten items, and will show both the "middle" Show More button and the Show More button at the bottom of the list of items if there are ten or more items.

PhilR8 (changes requested):
There's definitely a bug here but this isn't quite the right fix.  There should be a "Show More/Show Less" button visible -- the logic in the `RelatedRuleList` is wrong -- but the logic in `SupplementalContentList` is correct and shouldn't be changed.  

I've linked to a long Slack Validation thread in the comments below where it was decided that:

- if there are five or fewer items, then they're all shown
- if there are between five and ten items, a single Show More/Show Less button is shown in between the five items and the rest of the items
- and if there are ten or more items, there is a Show More/Show Less button shown in between the five items and the rest of the items, AND there is a Show More/Show Less button at the end of the list.

PhilR8 (approved):

PhilR8:
LGTM!

PR #547, opened 2022-09-19
EREGCSC-1499 Put Rule Groupings onto Homepage

PhilR8:
Resolves [EREGCSC-1499](https://jiraent.cms.gov/browse/EREGCSC-1499)

**Description**

Recent FR Rules are making their return to the homepage after a brief absence.  They will appear in the right sidebar of the homepage as before, but now they will be displayed as grouped FR docs, just like the are displayed in the right sidebar of the Reader View.

- Updates homepage sidebar to reintroduce Recent FR Docs
- Limits number of FR document groups to 3
- Limits number of lines of title text to 3 for homepage FR docs

1. Visit [experimental deployment homepage](https://x76onhzk4g.execute-api.us-east-1.amazonaws.com/dev547) and look at right sidebar.  Make sure the "Recent Rules" section exists above the "All Resources" section.  
2. Make sure everything matches what is seen in the [Figma Comp here](https://www.figma.com/proto/IoCTRZi12sUFYzGgRgJN5S/eRegs-Homepage---Sidebar-Update?page-id=2410%3A671&node-id=2410%3A2665&viewport=-2100%2C3199%2C0.11&scaling=scale-down&starting-point-node-id=2410%3A2665).
   - Note: per conversation with Betsy, both "View More Changes" button and the "View All Resources" button will have a height of 36px.
3. Follow the "View More Changes" link and it should take you to the resources page with the FR Docs category filter pre-selected.
4. Ensure the three FR doc groups shown are the most recent three docs (compare against Resources page or API response, etc).


solution/backend/regulations/templates/regulations/partials/recent_changes.html
```diff
@@ -2,17 +2,21 @@
...
-    {% comment %}<h3>Recent Documents for 42 CFR 430-460</h3>{% endcomment %}
+    <h3>Recent Rules</h3>
 
-    {% comment %}<supplemental-content class="related-rules" api_url="{% url 'regcore' %}" requested_categories="26"></supplemental-content>{% endcomment %}
+    <p class="recent-rules-descriptive-text">Includes 42 CFR 400, 430-460</p>
 
-    {% comment %}<hr class="latest-changes-horiz-rule"/>{% endcomment %}
+    <recent-changes-container api-url={% url "regcore" %} ></recent-changes-container>
+
+    <a href="{% url 'resources' %}?title=42&resourceCategory={{ fr_docs_category_name }}" class="action-btn default-btn link-btn">View More Changes</a>
```
peggles2, line 11: Tiny nit pic but can we keep the quotations consistant? I see {% url 'resources' %} in this line but line 9 its {% url "regcore" %} ?

solution/ui/regulations/js/src/components/RelatedRule.vue
```diff
@@ -11,7 +11,8 @@
                     v-if="type !== ''"
                     class="recent-flag indicator"
                     :class="indicatorClasses"
-                >{{type}}</span>
+                    >{{ type }}</span
+                >
```
peggles2, line 15: Is this the closing tag for </span placed on the next line?

peggles2 (commented):
Looks good, just some minor cosmetic tweaks. Also do we want to add some tests for this?

solution/ui/regulations/js/src/components/RelatedRule.vue
PhilR8: indeed.  I use prettier to format code via a hotkey and it has some interesting opinions on how to format tags like this.  The real reason is that it ensures that no empty spaces are included on either side of the `type` variable, which can happen when you have the span tags on different lines.  I liked it how it was before this change, but I also don't mind prettier's change.  This is something we can add to the list of things we'd like to adjust with our linters and formatters.

solution/backend/regulations/templates/regulations/partials/recent_changes.html
PhilR8: we need `'resources'` to be in single quotes because it is quoted within a double-quoted `href` attribute.

PhilR8:
> Looks good, just some minor cosmetic tweaks. Also do we want to add some tests for this?

I think adding a test to the homepage suite is a good idea.  I'll add one this afternoon.

solution/ui/regulations/js/src/components/RelatedRule.vue
peggles2: Agreed!

peggles2 (approved):
Nice work with the cypress tests!

PR #550, opened 2022-09-21
EREGCSC-1522 Switch all eCFR parser endpoints to v3

cgodwin1:
Resolves #1522

**Description-**

To facilitate the move to the V3 API, the eCFR and FR parsers need to exclusively use it.

- eCFR and FR parsers use only V3 API
- New environment variable, EREGS_API_URL_V3 exists pointing to V3, instead of modifying existing variable
    - Existing variable should instead be removed when transition is complete
- New V3 endpoint, `/v3/parser_config`, exists and functions the same as the V2 counterpart
- Packages are split out from the parser apps. New packages are: `fr-parser` and `ecfr-parser` for the parsers themselves, and `lib/ecfr`, `lib/eregs`, `lib/fedreg`, `lib/network`, and `lib/parsexml` for library packages.
    - This reduces complexity with regards to importing dependencies.
    - Consistent integration with V3 API - no more importing `eregs` package from the other parser and changing the URL

1. Visit `/v3/parser_config` and verify that parser configuration shows up and is the same as `/v2/parser_config`
2. Visit `/api/swagger` and make sure docs for the above endpoint are correct
3. Run eCFR and Federal Register parsers to ensure they function


thwalker6 (approved):
LGTM

PR #555, opened 2022-09-23
ERECSC-1508 - clicking on left section should update section content on…

peggles2:
Resolves #

Clicking on the left nav sub section should update section content on the right

- expected change 2

1. On local server visit localhost:8000/42/430/
2. Click on Subpart A 
3. Click on a subsection
4. You should see on the right hand side the details of the sub section. 

Screen Shot 2022-09-23 at 5 23 46 PM

### DIFF
--- a/solution/ui/e2e/cypress/integration/left-sidebar.spec.js
+++ b/solution/ui/e2e/cypress/integration/left-sidebar.spec.js
@@ -13,7 +13,7 @@ describe("Left sidebar", () => {
 
     })
 
-    it("opens the section when the right nav subsection is clicked", () => {
+    it("opens the section when the left nav subsection is clicked", () => {
         cy.viewport("macbook-15");
         cy.visit(destination);
         cy.get(`h3#nav-Subpart-A`).click({ force: true })
--- a/solution/ui/e2e/cypress/integration/left-sidebar.spec.js
+++ b/solution/ui/e2e/cypress/integration/left-sidebar.spec.js
@@ -12,6 +12,7 @@ describe("Left sidebar", () => {
         });
 
     })
+
     it("opens the section when the right nav subsection is clicked", () => {
         cy.viewport("macbook-15");
         cy.visit(destination);
--- a/solution/ui/e2e/cypress/integration/left-sidebar.spec.js
+++ b/solution/ui/e2e/cypress/integration/left-sidebar.spec.js
@@ -17,6 +17,10 @@ describe("Left sidebar", () => {
         cy.viewport("macbook-15");
         cy.visit(destination);
         cy.get(`h3#nav-Subpart-A`).click({ force: true })
+        cy.get(`h2#subpart-resources-heading`).should(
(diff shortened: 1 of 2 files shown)


solution/ui/regulations/js/src/components/SupplementalContent.vue
```diff
@@ -192,32 +192,41 @@ export default {
...
-                this.selectedPart = `§ ${section}`;
-        }
-        } else {
+          location = this.parseHash(window.location.hash)
+         } else {
             this.fetch_content(this.title, this.part);
         }
         this.fetch_content(this.title, this.part, location)
+        window.addEventListener('hashchange', (event) => {
+          location = this.parseHash(window.location.hash)
```
PhilR8, line 201: the second argument of `addEventListener` needs to be a named function instance or method so the listener can be removed in the `destroyed` lifecycle.  See explanation here: 

https://deepscan.io/docs/rules/bad-remove-event-listener

It's likely that we'll have to create another method that calls `this.parseHash` and then calls `this.fetch_content` and then use that exact method instance here, as well as in the `removeEventListener` method.

solution/ui/regulations/js/src/components/SupplementalContent.vue
```diff
@@ -192,32 +192,41 @@ export default {
...
     },
-
     mounted() {
         this.$root.$on(EventCodes.SetSection, (args) => {
             this.selectedPart = args.section;
         });
         this.categories = getDefaultCategories();
     },
+    destroyed() {
+      window.removeEventListener("hashchange", this.hashchange);
```
PhilR8, line 212: `this.hashchange` is undefined because it's not a property defined anywhere in the component.  

The second argument here needs to be the exact same function instance that was passed into the `addEventListener` method above.  See description and examples here:

https://deepscan.io/docs/rules/bad-remove-event-listener

solution/ui/e2e/cypress/integration/left-sidebar.spec.js
```diff
@@ -13,6 +13,17 @@ describe("Left sidebar", () => {
 
     })
 
+    it("opens the section when the left nav subsection is clicked", () => {
+        cy.viewport("macbook-15");
+        cy.visit(destination);
+        cy.get(`h3#nav-Subpart-A`).click({ force: true })
```
PhilR8, line 19: it might be nice here to have an assertion that checks that the `h2#subpart-resources-heading` has the expected "Subpart A Resources" text before the section click so we ensure it changes.  That's just a personal opinion though.

PhilR8 (changes requested):

solution/ui/e2e/cypress/integration/left-sidebar.spec.js
peggles2: Good suggestion.

PhilR8 (approved):
LGTM!  Nice work!

PR #556, opened 2022-09-23
EREGCSC-1506 Subpart letters not in admin panel for part 600

cgodwin1:
Resolves #1506

**Description-**

In order to test if the previous fixes to the eCFR parser resolve the referenced issue, we need to be able to allow the eCFR parser to reparse all regulation parts. Before, disabling "Skip versions" in the Django Admin panel would also permit the Federal Register parser to reparse all FR docs, which is currently not acceptable.

- Two options in the admin panel, `skip_reg_versions` and `skip_fr_documents`, replaces single option `skip_versions`.
- Both options default to "true", and `skip_fr_documents` is currently uneditable.
- eCFR parser uses `skip_reg_versions`, FR parser uses `skip_fr_documents`.

1. Run Go tests and linting (`go test ./...` and `golint ./...` respectively)
2. Visit `/v2/parser_config` and verify the new options show up and are correctly displayed
3. Run the FR parser and the eCFR parser, at least for a few seconds, to make sure they correctly begin


solution/backend/regcore/migrations/0012_auto_20220923_1156.py
```diff
@@ -0,0 +1,48 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='parserconfiguration',
+            name='skip_versions',
+        ),
+        migrations.AddField(
+            model_name='parserconfiguration',
+            name='skip_fr_documents',
+            field=models.BooleanField(default=True, help_text='Should the Federal Register parser skip processing documents that have been previously processed?'),
```
peggles2, line 21: would it be easier to read if we don't have all these in one long line?

solution/parser/fr-parser/main.go
```diff
@@ -140,7 +140,7 @@ func start() error {
 		log.Info("[main] retrieving content for title ", title.Title)
 		parts := getPartsListFunc(ctx, title)
 		for _, part := range parts {
-			if err := processPartFunc(ctx, title.Title, part, existingDocs, config.SkipVersions, titles); err != nil {
+			if err := processPartFunc(ctx, title.Title, part, existingDocs, config.SkipFRDocuments, titles); err != nil {
```
peggles2, line 143: This method processPartFunc takes a lot of arguments. Generally best practice to have max of 3 arguments.  https://medium.com/@meeusdylan/go-reduce-function-parameters-19b785a87a59

peggles2 (changes requested):
Nice job! looks good! Just some minor tweaks.

solution/backend/regcore/migrations/0012_auto_20220923_1156.py
cgodwin1: Migrations are automatically generated and we try not to edit them unless writing a custom one

solution/parser/fr-parser/main.go
cgodwin1: Happy to oblige on this! But I think it should be another tech debt ticket honestly, it's out of the scope of this fix and would take more significant refactoring to make all code compliant.

peggles2: Sounds good, please go ahead and create a ticket for this. And I will approve this pull request.

peggles2 (approved):

cgodwin1:
Follow up ticket for reducing function parameter complexity: EREGCSC-1539 in JIRA

PR #562, opened 2022-09-28
[Hotfix] Get correct category for FR docs

PhilR8:
Resolves production error caused by [EREGCSC-1499](https://jiraent.cms.gov/browse/EREGCSC-1499)

**Description**

Was incorrectly (and inefficiently) looking up the name of the FR Docs Category in a way that was not working on dev/val/prod.  Worked with @cgodwin1 to use a better, more efficient way.

**This pull request changes**

- Replaces looping over categories in favor of accessing the category directly and getting the name property

1. Ensure that the "View More Changes" button in the "Recent Changes" section of the homepage (right sidebar) has the proper FR Docs category name (whatever it may be in each environment) in the `href` attribute, and that the button takes you to the resources page with the FR Docs category pre-selected


cgodwin1 (approved):
LGTM

PR #563, opened 2022-09-28
EREGCSC-1523

thwalker6:
Resolves # EREGCSC-1523

**Description-**

Set up intial unit test for django.  First test are for some of the api in regcore

- expected change 1

Added unit test for django.  No changes to app, only changes was adding test

 You can run the make command, "make python.unittest".  This will run the first 10 test for regcore.  A couple of model test and then some testing out the api.

### DIFF
--- a/solution/backend/regcore/tests/test_api_view.py
+++ b/solution/backend/regcore/tests/test_api_view.py
@@ -1,0 +1,115 @@
+from datetime import date
+from regcore.models import Part
+from regcore.search.models import Synonym
+from rest_framework import status
+from collections import OrderedDict
+from rest_framework.test import APITestCase
+
+
+class RegcoreSerializerTestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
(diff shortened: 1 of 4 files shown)


solution/backend/regcore/tests/test_api_view.py
```diff
@@ -0,0 +1,115 @@
...
+                                                     ('label_level', 'Part 432'),
+                                                     ('parent_type', 'subchapter'),
+                                                     ('descendant_range', ['432.1', '432.55']),
+                                                     ('label_description', 'State Personnel Administration'),
+                                                     ('children', [])])])])])])]}
+
+        response = self.client.get("/v3/title/42/toc")
+
+        self.assertEqual(response.data, toc)
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
```
peggles2, line 80: Should we add one where the response fails? What error would you expect?

peggles2 (approved):
LGTM 👍   Just one suggestion for a test.

PR #565, opened 2022-09-29
EREGCSC-1460 Convert polymorphic serializers into fields for Swagger docs

cgodwin1:
Resolves #1460

**Description-**

As a developer integrating with eRegs' API, I want accurate OpenAPI documentation available to me. Currently, when a field may return more than one data type, this is not accurately reflected in Swagger. Using drf-spectacular's "PolymorphicProxySerializer", we can include this information.

_For V3 only_:
- Fixes serializer name conflicts between regcore and resources apps
    - Serializers in ALL apps must be named differently to not confused Swagger.
- Resources viewsets use PolymorphicProxySerializer to properly serialize endpoints. These changed:
    - /v3/resources
    - /v3/resources/locations
    - /v3/resources/categories
- Fixed Swagger conflicts (i.e. duplicate component names, etc.)
- Fixes bugs caused by drf-spectacular updates
- Pinned drf-spectacular package to version `0.24.2` as recommended by https://drf-spectacular.readthedocs.io/en/latest/readme.html#release-management: "For production we strongly recommend pinning the version and inspecting a schema diff on update."
    - Prevents future bugs from suddenly popping up due to schema breaking changes to drf-spectacular.
- Incidental: fixed exception on homepage if no FR Doc Category is set in the configuration

1. Verify Cypress tests pass (no API errors)
2. Visit `/api/swagger` and read through response examples for v3 endpoints for accuracy

**Future work...**

Debugging Swagger revealed a large number of warnings or errors in the logs. Many are caused by outdated V2 endpoints and these will be resolved naturally when V2 support is dropped. However, there are some that should be addressed related to field hints and authentication in V3. Only in-scope warnings are fixed by this ticket.

### DIFF
--- a/solution/backend/cmcs_regulations/settings.py
+++ b/solution/backend/cmcs_regulations/settings.py
@@ -90,8 +90,7 @@ STATIC_URL = os.environ.get("STATIC_URL", None)
 STATIC_ROOT = os.environ.get("STATIC_ROOT", None)
 
 # SECURITY WARNING: don't run with debug turned on in production!
-#DEBUG = os.environ.get("DEBUG", False)
-DEBUG = True
+DEBUG = os.environ.get("DEBUG", False)
 
 WORKING_DIR = os.environ.get("WORKING_DIR", "/var/lib/eregs")
 TEMPLATES = [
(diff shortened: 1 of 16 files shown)


thwalker6 (approved):
lgtm

PR #566, opened 2022-10-03
1532 - Use OIDC for key management

peggles2:
Resolves #

**Description-**

- expected change 1

### DIFF
--- a/.github/workflows/deploy-prototype.yml
+++ b/.github/workflows/deploy-prototype.yml
@@ -2,6 +2,17 @@ name: "Deploy Prototype"
 
 on: [fork]
 
+configure_aws_github: &configure_aws_github
+  uses: aws-actions/configure-aws-credentials@v1
+  with:
+    role-to-assume: ${{ env.AWS_OIDC_ROLE_TO_ASSUME }}
+    aws-region: ${{ env.AWS_DEFAULT_REGION }}
+
+permissions:
+  id-token: write
+  contents: read
+  actions: read
+
 jobs:
   deploy:
     environment:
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy-prototype.yml
```diff
@@ -21,18 +26,21 @@ jobs:
...
       - name: Setup Node
         uses: actions/setup-node@v1
         with:
           node-version: '14.x'
+      # Configure AWS credentials for GitHub Actions
+      - name: Configure AWS credentials for GitHub Actions
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_OIDC_ROLE_TO_ASSUME }}
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
```
PhilR8, line 39: is there a reason why we're using `AWS_DEFAULT_REGION` here but explicitly using `us-east-1` elsewhere?

peggles2: Oh good question! During deployment experimental, the notify action generates the Django Site, Prototype Site and Storybook urls. The way it does this is from outputting the url like this           `echo "::set-output name=url::$(cat output.log | grep -m1 'ANY -' | cut -c 9-)"` but when this is a secret,  then github is returning it as **** which messes up the url and made the deployment failed. To fix that had to hardcode the region. This does not impact deploy-prototype.yml as we do not have notify action. If you like we can do a quick zoom call so I can explain it better. 
Here is what the URL looks correctly, and when its set as a secret. 

```
endpoints:
  ANY - https://bstpsibijc.execute-api.us-east-1.amazonaws.com/dev564
  ANY - https://bstpsibijc.execute-api.us-east-1.amazonaws.com/dev564/{proxy+}

```
```
endpoints:
  ANY - https://3s5macya05.execute-api.***.amazonaws.com/dev566
  ANY - https://3s5macya05.execute-api.***.amazonaws.com/dev566/{proxy+}  
```
  
I'll just change that to make it consistent even if it works fine for the other deployment.yml files.

PhilR8: makes sense to me!  I think us-east is the cheapest region anyway.

PR #577, opened 2022-10-14
EREGCSC-1559 fix of not matching csv

thwalker6:
Resolves #570 

**Description-**
CSV was not matchin search results do to fr-grouping and title 45 being included

- expected change 1

Go to the resources page and search for anything like "covid".

Download the csv.

The same amount of results should appear.  You can also verify by looking at the api calls.  It should be the same, provided it is less than 100 results.

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
+++ b/solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
@@ -91,10 +91,9 @@ export default {
             while (supNum < this.supCount) {
                 content.push(getSupplementalContentV3({
                     page,
-                    partDict: this.partDict,
+                    partDict: Object.keys(this.partDict).length > 0 ? this.partDict : "all",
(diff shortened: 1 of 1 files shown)


solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -93,7 +93,9 @@ export default {
                     page,
                     partDict: this.partDict,
                     categories: this.categories,
-                    q: this.query
+                    q: this.query,
+                    partDict:"all",
```
PhilR8, line 97: there are two `partDict` keys, with the second one (`"all"`) overwriting the first one (`this.partDict`).

Also regarding formatting: there should be a space between the key and the val

`partDict: "all",`

PhilR8 (changes requested):

PhilR8 (approved):

PhilR8:
LGTM!

PR #581, opened 2022-10-18
EREGCSC-1520-B Add Federal Register images to CSP policy

cgodwin1:
Follow-up to #1520

**Description-**

`https://images.federalregister.gov/*` was not in the CSP, so images loaded directly from that site were blocked.

- Add the above URL to the CSP

1. Visit `/42/433/Subpart-A/2021-03-01/#433-10-c-8` on the experimental deploy and verify the image within the reg text loads


PhilR8:
Should we use this PR to whitelist all the domains that we're finding are blocked?  I'm thinking the JSDeliver CDN that's currently breaking the swagger page, as well as the manifest files that are blocked on the prod app.  And we could poke around and see if anything else needs to be whitelisted.

thwalker6 (approved):
lgtm

PR #584, opened 2022-10-19
EREGCSC-1541 Adding spinner to export CSV

thwalker6:
Resolves #ERGECSC-1541

**Description-**

When the download csv button is clicked on the resources page, a spinner replaces the text.  When it is finished downloading it goes away.

- expected change 1

Go the resources page.

Click download CSV.  

As it is donwloading the text will appear and a spinner will take its place.

When it is done it will go back to the text.

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
+++ b/solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
@@ -6,8 +6,8 @@
             @click="createCSV"
         >
             <template v-if="!downloadingCSV">
-                <span class="desktop-btn-label">Download Spreadsheet (CSV)</span
-                ><span class="mobile-btn-label">CSV</span>
+                <span class="label desktop-btn-label">Download Spreadsheet (CSV)</span
+                ><span class="label mobile-btn-label">CSV</span>
                 <svg
                     width="18"
                     height="17"
@@ -23,8 +23,7 @@
             </template>
             <template v-else>
                 <simple-spinner
-                    spinner-type="button"
-                    size="small"
+                    size="xs"
                 ></simple-spinner>
             </template>
         </button>
@@ -133,32 +132,38 @@ export default {
     },
 };
 </script>
-<style lang="scss" scoped>
+<style lang="scss">
 .resource-btn-container {
     padding-left: 5px;
 
     #exportCSV {
         padding-top: 8px;
         padding-bottom: 8px;
+        width: 50px;
         min-height: 18px;
         align-items: center;
-        justify-content: space-around;
+        justify-content: center;
 
         @include screen-md {
-            min-width: 225px;
+            width: 220px;
         }
 
-        .desktop-btn-label {
-            @include custom-max($mobile-max / 1px) {
-                display: none;
+        .label {
+            margin-right: 7px;
+
+            &.desktop-btn-label {
+                @include custom-max($mobile-max / 1px) {
+                    display: none;
+                }
+            }
+
+            &.mobile-btn-label {
+                @include screen-md {
+                    display: none;
+                }
             }
         }
 
-        .mobile-btn-label {
-            @include screen-md {
-                display: none;
-            }
-        }
     }
 }
 </style>
--- a/solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
+++ b/solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
@@ -1,12 +1,21 @@
 <template>
-    <div class="resource_btn_container">
-        <button class="default-btn action-btn search_resource_btn desktop-btn-label`" @click="createCSV">
-            <span class="desktop-btn-label">Download Spreadsheet (CSV)</span><span class="mobile-btn-label">CSV</span>&nbsp;&nbsp;
-            <svg width="18" height="17" viewBox="0 0 18 17" fill="none" xmlns="http://www.w3.org/2000/svg">
-                <path
-                    d="M1.78361 16.5C1.36371 16.5 0.996298 16.35 0.681372 16.05C0.366447 15.75 0.208984 15.4 0.208984 15V11.425H1.78361V15H15.4304V11.425H17.005V15C17.005 15.4 16.8475 15.75 16.5326 16.05C16.2177 16.35 15.8503 16.5 15.4304 16.5H1.78361ZM8.60699 12.675L3.54194 7.85L4.67043 6.775L7.81968 9.775V0.5H9.39431V9.775L12.5436 6.775L13.672 7.85L8.60699 12.675Z"
-                    fill="white" />
-            </svg>
+    <div class="resource_btn_container" style="padding-left:5px;">
+        <button id="export-csv" class="default-btn action-btn search_resource_btn desktop-btn-label`"
+            @click="createCSV">
+            <span v-if="!downloadingCSV">
+                <span class="desktop-btn-label">
+                    Download Spreadsheet (CSV)
+                </span>
+                <span class="mobile-btn-label">CSV</span>
+                <svg width="18" height="17" viewBox="0 0 18 17" fill="none" xmlns="http://www.w3.org/2000/svg"
+                    v-if="!downloadingCSV">
+                    <path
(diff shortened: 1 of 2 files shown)


solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -1,12 +1,21 @@
 <template>
-    <div class="resource_btn_container">
-        <button class="default-btn action-btn search_resource_btn desktop-btn-label`" @click="createCSV">
-            <span class="desktop-btn-label">Download Spreadsheet (CSV)</span><span class="mobile-btn-label">CSV</span>&nbsp;&nbsp;
-            <svg width="18" height="17" viewBox="0 0 18 17" fill="none" xmlns="http://www.w3.org/2000/svg">
-                <path
-                    d="M1.78361 16.5C1.36371 16.5 0.996298 16.35 0.681372 16.05C0.366447 15.75 0.208984 15.4 0.208984 15V11.425H1.78361V15H15.4304V11.425H17.005V15C17.005 15.4 16.8475 15.75 16.5326 16.05C16.2177 16.35 15.8503 16.5 15.4304 16.5H1.78361ZM8.60699 12.675L3.54194 7.85L4.67043 6.775L7.81968 9.775V0.5H9.39431V9.775L12.5436 6.775L13.672 7.85L8.60699 12.675Z"
-                    fill="white" />
-            </svg>
+    <div class="resource_btn_container" style="padding-left:5px;">
```
PhilR8, line 2: we should avoid inline styles whenever possible

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -1,12 +1,21 @@
 <template>
-    <div class="resource_btn_container">
```
PhilR8, line 2: I figured I would leave some comments on your commit to give you an idea of what I changed and why.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
PhilR8: We should avoid inline styles if possible.  Styles should be in a class or id.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -1,12 +1,21 @@
...
-        <button class="default-btn action-btn search_resource_btn desktop-btn-label`" @click="createCSV">
-            <span class="desktop-btn-label">Download Spreadsheet (CSV)</span><span class="mobile-btn-label">CSV</span>&nbsp;&nbsp;
-            <svg width="18" height="17" viewBox="0 0 18 17" fill="none" xmlns="http://www.w3.org/2000/svg">
-                <path
-                    d="M1.78361 16.5C1.36371 16.5 0.996298 16.35 0.681372 16.05C0.366447 15.75 0.208984 15.4 0.208984 15V11.425H1.78361V15H15.4304V11.425H17.005V15C17.005 15.4 16.8475 15.75 16.5326 16.05C16.2177 16.35 15.8503 16.5 15.4304 16.5H1.78361ZM8.60699 12.675L3.54194 7.85L4.67043 6.775L7.81968 9.775V0.5H9.39431V9.775L12.5436 6.775L13.672 7.85L8.60699 12.675Z"
-                    fill="white" />
-            </svg>
+    <div class="resource_btn_container" style="padding-left:5px;">
+        <button id="export-csv" class="default-btn action-btn search_resource_btn desktop-btn-label`"
```
PhilR8: classes should be kebab-cased: `search-resource-btn` instead of `seach_resource_btn`

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -1,12 +1,21 @@
...
+            @click="createCSV">
+            <span v-if="!downloadingCSV">
```
PhilR8: you can use nested `templates` for `v-if` and `v-else` directives.  They're kind of like React fragments -- they don't actually add an element to the page.  Using a `span` adds a span and can add complexity.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -1,12 +1,21 @@
...
+                </span>
+                <span class="mobile-btn-label">CSV</span>
+                <svg width="18" height="17" viewBox="0 0 18 17" fill="none" xmlns="http://www.w3.org/2000/svg"
+                    v-if="!downloadingCSV">
+                    <path
+                        d="M1.78361 16.5C1.36371 16.5 0.996298 16.35 0.681372 16.05C0.366447 15.75 0.208984 15.4 0.208984 15V11.425H1.78361V15H15.4304V11.425H17.005V15C17.005 15.4 16.8475 15.75 16.5326 16.05C16.2177 16.35 15.8503 16.5 15.4304 16.5H1.78361ZM8.60699 12.675L3.54194 7.85L4.67043 6.775L7.81968 9.775V0.5H9.39431V9.775L12.5436 6.775L13.672 7.85L8.60699 12.675Z"
+                        fill="white" />
+                </svg></span>
+            <span class="desktop-btn-label" style="width:100%" v-else>
+                <simple-spinner spinnerType="button" :size="'small'"></simple-spinner>
```
PhilR8: the colon (`:`) in front of an attribute is shorthand for `v-bind`, which binds a prop from the component context to the attribute.  In our case we don't need to bind anything to `size` because we're explicitly setting it as a string.  So this component should look like this:

`<simple-spinner spinnerType="button" size="small"></simple-spinner>`

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -15,12 +24,17 @@
 <script>
 
 import exportFromJSON from 'export-from-json'
+import SimpleSpinner from "../../../../../regulations/js/src/components/SimpleSpinner.vue";
```
PhilR8: We don't have to dig up and out to another directory like this -- the vite config file has aliases to make this easier.  See here:

https://github.com/CMSgov/cmcs-eregulations/blob/main/solution/ui/regulations/eregs-vite/vite.config.js

So to import SimpleSpinner, all you would need to do is:

`import SimpleSpinner from "legacy/js/src/components/SimpleSpinner.vue";`

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
```diff
@@ -15,12 +24,17 @@
...
 import {
     getSupplementalContentV3,
 
 } from "../../utilities/api";
```
PhilR8: Similarly you can import api as

`import { getSupplementalContentV3 } from "@/utilities/api";`

solution/ui/regulations/js/src/components/SimpleSpinner.vue
```diff
@@ -1,5 +1,6 @@
 <template>
     <div
+        :id="boxProperties"
```
PhilR8: I removed most of the changes in here because I didn't want to make big changes to this component.  I just added a new size (`xs`) that we can use for this story and set it to have `0px` of `margin`.  That way no other implementations of this component are affected.

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
thwalker6: Sounds good

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
thwalker6: I went pythony there but noted

solution/ui/regulations/eregs-vite/src/components/resources/ResourceExportBtn.vue
thwalker6: For some reason I thought v-if's were getting tagged by the linter with v-ifs but maybe I'm thinking of someone else.  react you can now just do <> </> for empty components

peggles2 (approved):
LGTM 👍

PR #592, opened 2022-10-27
EREGCSC-1589 correction withdrawal labels

thwalker6:
Resolves # EGEGCSC-1589

**Description-**

Adds two labels to the FR-Doc, withdrawal and correction.  Manually done, not automatically by parser.  This is just the backend portion.

- expected change 1

Two new properties are added to FR documents.  1 for correction, 1 for withdrawal.  Both visible in an api call for the object, and on the admin panel.

Go into the django admin for an FR Doc.  You should see two check boxes, 1 for correction, 1 for withdrawal.

Make an api call for that Fr-doc. 

Both should be false by default. 

Make one or both true.

Make another api call, see that it updates.


peggles2 (approved):
LGTM 👍

PR #595, opened 2022-11-03
EREGCSC-1578 Re-running FR parser: Endpoints

cgodwin1:
Resolves #1578

**Description-**

FR doc post endpoint needs to be updated to not overwrite any data exception locations, and apply the changelog after locations are updated (ideally done in one step, because this is an atomic endpoint).

- FR doc post endpoint parses and applies the changelog if a doc is updated
- Everything else about an FR doc is maintained if it already exists (`create` variable passed through to serializer context)
- `update` method of FR doc create serializer is simplified (sections of code moved to class methods)
- Semi-related: improved location key retrieval method on admin panel changelog representation

_Testing is best done locally so that you can re-run the FR parser immediately, or mock the request using e.g. Postman._

1. Run the FR parser or mock the request to create a fake FR doc.
2. Visit the FR doc on the admin panel and add/remove several locations through multiple save operations. Rerun the parser and verify locations are as expected.
3. Try adding a location that has been removed in a previous save. Rerun the parser and verify locations are expected.
4. Bulk add one or more locations and rerun the parser.
5. If using Postman, add a section manually to the data and verify that it shows as expected.

### DIFF
--- a/solution/backend/resources/v3serializers/resources.py
+++ b/solution/backend/resources/v3serializers/resources.py
@@ -196,8 +196,8 @@ class FederalRegisterDocumentCreateSerializer(serializers.Serializer):
                 all_removals.remove(i) if i in all_removals else all_additions.append(i)
             for i in removals:
                 all_additions.remove(i) if i in all_additions else all_removals.append(i)
-        [instance.locations.remove(i) for i in self.get_location_objects(all_removals)]
-        [instance.locations.add(i) for i in self.get_location_objects(all_additions)]
+        instance.locations.remove(*self.get_location_objects(all_removals))
+        instance.locations.add(*self.get_location_objects(all_additions))
 
     def set_locations(self, instance, raw_sections, raw_ranges):
         locations = []
--- a/solution/backend/resources/v3serializers/resources.py
+++ b/solution/backend/resources/v3serializers/resources.py
@@ -144,8 +144,10 @@ class FederalRegisterDocumentCreateSerializer(serializers.Serializer):
 
     def update(self, instance, validated_data):
         created = self.context.get("created", False)
+        sections = validated_data.get("sections", [])
+        section_ranges = validated_data.get("section_ranges", [])
 
-        # set basic fields if instance is new
+        # set basic fields and group if instance is new
         if created:
             instance.url = validated_data.get('url', instance.url)
             instance.description = validated_data.get('description', instance.description)
@@ -156,17 +158,14 @@ class FederalRegisterDocumentCreateSerializer(serializers.Serializer):
             instance.approved = validated_data.get('approved', instance.approved)
             instance.doc_type = validated_data.get('doc_type', instance.doc_type)
             instance.category = self.get_category()
+            self.set_group(instance)
 
         # set the locations on the instance
-        self.set_locations(instance, validated_data["sections"], validated_data["section_ranges"])
+        self.set_locations(instance, sections, section_ranges)
 
         # reapply changelog if instance is not new
         if not created:
             self.apply_changelog(instance)
-
-        # set the group on the instance if instance is new
-        if created:
-            self.set_group(instance)
 
         # save and return
         instance.save()
@@ -201,7 +200,7 @@ class FederalRegisterDocumentCreateSerializer(serializers.Serializer):
 
     def set_locations(self, instance, raw_sections, raw_ranges):
         locations = []
-        for loc in (raw_sections or []):
+        for loc in raw_sections:
             title = loc["title"]
             part = loc["part"]
             section_id = loc["section_id"]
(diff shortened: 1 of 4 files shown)


solution/backend/resources/v3serializers/resources.py
```diff
@@ -141,28 +143,73 @@ def validate_doc_type(self, value):
...
+            additions = [i.copy() for i in change["additions"] + change["bulk_adds"]]
+            removals = [i.copy() for i in change["removals"]]
+            for i in additions + removals:
+                del i["id"]
+            for i in additions:
+                all_removals.remove(i) if i in all_removals else all_additions.append(i)
+            for i in removals:
+                all_additions.remove(i) if i in all_additions else all_removals.append(i)
+        [instance.locations.remove(i) for i in self.get_location_objects(all_removals)]
+        [instance.locations.add(i) for i in self.get_location_objects(all_additions)]
```
cgodwin1, line 200: I realize this code looks like it could be simplified by just applying the changelog entries one at a time using `instance.locations.remove(...)` and `add(...)`. However this method is guaranteed to only need 2 queries per FR doc to run, instead of one DB query _per location object added or removed_.

solution/backend/resources/v3serializers/resources.py
```diff
@@ -141,28 +143,73 @@ def validate_doc_type(self, value):
...
+                all_removals.remove(i) if i in all_removals else all_additions.append(i)
+            for i in removals:
+                all_additions.remove(i) if i in all_additions else all_removals.append(i)
+        instance.locations.remove(*self.get_location_objects(all_removals))
+        instance.locations.add(*self.get_location_objects(all_additions))
+
+    def set_locations(self, instance, raw_sections, raw_ranges):
         locations = []
-        for loc in (validated_data["sections"] or []):
+        for loc in (raw_sections or []):
```
thwalker6, line 204: [] should be sufficient in the new method.  before you were looking at a property that might not exist, making the "or []" necessary, but if an empty raw_sections is passed in nit will just skip the for loop.

solution/backend/resources/v3serializers/resources.py
```diff
@@ -141,28 +143,73 @@ def validate_doc_type(self, value):
...
             title = loc["title"]
             part = loc["part"]
             section_id = loc["section_id"]
             location, _ = Section.objects.get_or_create(title=title, part=part, section_id=section_id)
             location.save()
             locations.append(location)
 
-        for loc_range in (validated_data["section_ranges"] or []):
+        for loc_range in (raw_ranges or []):
```
thwalker6, line 212: Same as above

solution/backend/resources/v3serializers/resources.py
```diff
@@ -141,28 +143,73 @@ def validate_doc_type(self, value):
...
+            instance.approved = validated_data.get('approved', instance.approved)
+            instance.doc_type = validated_data.get('doc_type', instance.doc_type)
+            instance.category = self.get_category()
 
         # set the locations on the instance
+        self.set_locations(instance, validated_data["sections"], validated_data["section_ranges"])
+
+        # reapply changelog if instance is not new
+        if not created:
+            self.apply_changelog(instance)
```
thwalker6, line 165: I guess its explicit but couldnt if else do the same thing as the two ifs?

solution/backend/resources/v3serializers/resources.py
```diff
@@ -141,28 +143,73 @@ def validate_doc_type(self, value):
...
+        for change in instance.location_history:
+            additions = [i.copy() for i in change["additions"] + change["bulk_adds"]]
+            removals = [i.copy() for i in change["removals"]]
+            for i in additions + removals:
+                del i["id"]
+            for i in additions:
+                all_removals.remove(i) if i in all_removals else all_additions.append(i)
+            for i in removals:
+                all_additions.remove(i) if i in all_additions else all_removals.append(i)
+        instance.locations.remove(*self.get_location_objects(all_removals))
```
thwalker6, line 199: I would maybe change it to if all_removals: and then an if  all_adds:  that way you only call the functions if needed.  If you did that then in the function you wouldnt have to return the empty list if there is no queries or locations.

thwalker6 (changes requested):
maybe discuss the notes before merging

solution/backend/resources/v3serializers/resources.py
cgodwin1: Good point. Also needed to add a check to ensure `validated_data['sections']` existed to start with. Thanks for the catch.

solution/backend/resources/v3serializers/resources.py
cgodwin1: Yes. In this case, as you said, it's just to be explicit and show the step-by-step nature of creating/updating an FR doc. But technically, setting the group could be moved to be within the first `if created` check.

PR #596, opened 2022-11-04
Eregcsc 1592 enable flagging fr doc frontend

peggles2:
Resolves #EREGCSC 1592

**Description-**
 Enable flagging a FR doc as a correction or withdrawal

Adds the label CORR and WD

- expected change 1

Visit the home page 
check and see if the right nav has WD for withdrawal, CORR if correction. 
If multiple this is the precedence list:

**Precedence list:**
WD
CORR
NPRM or FINAL


solution/ui/regulations/css/scss/_application_settings.scss
```diff
@@ -263,6 +265,12 @@ $width-3xl: 1920px; // The largest the site container will get
     color: $secondary_indicator_text;
 }
 
+.tertiary-indicator {
+    background: $tertiary-indicator_background;
+    color: $mid-gray;
+    border: 1px solid $mid-gray;
```
PhilR8, line 271: I'd add a couple more properties to `tertiary-indicator` here so that the indicator is the same size as the primary and secondary indicators.  Right now the tertiary indicator is a few pixels taller and wider than the primary and secondary indicators, mainly because the tertiary indicator has a border and the others do not.

```
box-sizing: border-box;
padding: 1px 0;
```

`border-box` takes the border into account when determining height and width.  We're only explicitly setting width, so this will made the width 42px to match the other indicators (which don't have borders).  Without this property, the 1px border adds a px of width to each side making the indicator 44px wide.

And we're just updating the padding to counteract the 1px of top and bottom border, so the tertiary indicator becomes 20.5px tall, just like the other indicators.

solution/ui/regulations/css/scss/_application_settings.scss
```diff
@@ -102,6 +102,8 @@ $primary_indicator_text: $white;
 $secondary_indicator_background: $light_gray;
 $secondary_indicator_text: $title_color;
 
+$tertiary-indicator_background: $white;
```
PhilR8, line 105: this variable should be fully `snake_cased` to match the other variables

solution/ui/regulations/css/scss/_application_settings.scss
```diff
@@ -263,6 +265,12 @@ $width-3xl: 1920px; // The largest the site container will get
     color: $secondary_indicator_text;
 }
 
+.tertiary-indicator {
+    background: $tertiary-indicator_background;
+    color: $mid-gray;
```
PhilR8, line 270: it's interesting that you've used `$mid-gray` with a dash as the variable name, because the variable is defined on line 21 with an underscore: `$mid_gray`.  

And yet, even though you've used a dash, it is working and compiling to the correct color.

I was confused as to how this could happen, but after a little research I've learned something new: in Sass, it doesn’t matter whether your variable has an underscore or a dash - [they work both ways](https://pyronaur.com/dashes-and-underscores-in-sass-scss/).  Today I learned!

Still, it might be clearer if these are changed to `$mid_gray` with an underscore to match the actual variable declaration.

PhilR8 (changes requested):
A few small comments.

solution/ui/e2e/cypress/integration/homepage.spec.js
```diff
@@ -156,14 +156,102 @@ describe("Homepage", { scrollBehavior: "center" }, () => {
...
+                    expect($flag).to.have.text("Final");
+                    expect($flag)
+                        .to.have.css("background-color")
+                        .and.eq("rgb(2, 102, 102)");
+                });
+        });
+    });
+
+    it("Sets the label as WD when Correction is false and Withdrawal is true", () => {
+       cy.viewport("macbook-15");
```
PhilR8, line 184: indent

solution/ui/regulations/js/src/components/RelatedRuleList.vue
```diff
@@ -169,6 +169,12 @@ export default {
             return rule.title || rule.description;
         },
         type(rule) {
+            if (rule.withdrawal) {
+              return "WD"
```
PhilR8, line 173: Our existing JS indent style is four spaces.  You have two-space indents here and below

solution/ui/regulations/js/src/components/RelatedRule.vue
```diff
@@ -88,6 +88,11 @@ export default {
             };
         },
         indicatorClasses() {
+            if( this.type === "WD") {
+              return {
+                "tertiary-indicator": this.type === "WD",
+              };
```
PhilR8, line 94: needs four-space indents

PhilR8 (changes requested):

PhilR8 (approved):
LGTM!

PR #599, opened 2022-11-10
EREGCSC-1610 Disable django toolbar

thwalker6:
Resolves #Eregcsc-1610

**Description-**

When running cypress locally, will disable the django toolbar.  The toolbar would cause some cypress test to fail.

- expected change 1

Run make test on your local machine.  All cypress test should pass(at least pass if they are good :D)


PhilR8 (approved):
Nice work!  LGTM!

PR #600, opened 2022-11-10
Tech Debt: Update print API test to be less flaky

PhilR8:
**Description**
An intercept and assertion was added to the print test suite to test the format of the date in the footer. 

The problem: it is very flaky and unreliable.

- Moves intercept into `beforeEach` block as suggested in one of the many Github Issue comments about this issue

1. Run print test suite a bunch of times and see if it is more reliable than before


thwalker6 (approved):
lgtm

PR #610, opened 2022-11-15
EREGCSC-1632 Guarantee uniqueness of resource sections and subparts

cgodwin1:
Resolves #1632

**Description-**

Due to the inherited nature of the Section and Subpart models, it's not possible to require uniqueness on a combination of title, part, and section/subpart identifiers in the normal Django/Postgres way. However, this is causing issues where duplicate sections are being created somewhere. (Automatically or manually, we're not sure yet.) If this happens, the FR parser will fail to run for some FR docs.

- `validate_unique` method is overridden for Section and Subpart models, and checks if we're trying to create a duplicate.
- If a duplicate is detected, ValidationError is raised.
- `save` method is overridden for AbstractLocation to call `full_clean()`, which means other code doesn't need to change.

1. In the admin panel, try to create a new section that already exists. You should see an error on the page.
2. Optionally, locally in `docker-compose exec regulations python manage.py shell`, run the following:
    - `from resources.models import Section`
    - `Section.objects.create(title=1, part=1, section_id=1)` (this should not fail unless you already created this section)
    - Run the above again and observe the exception.


peggles2 (approved):
LGTM 👍

PR #624, opened 2022-12-05
EREGCSC-1652 Going to appendix of 441 subpart F returns error

cgodwin1:
Resolves #1652

**Description-**

Issue is that location hash for appendices is, e.g. `#Appendix-to-Subpart-F-of-Part-441`. When parsed into a location string, it is `42.Appendix.to` which is not valid. Since we are not currently linking resources to appendices, this PR just disables the right sidebar from requesting resources from an appendix.

- If location hash contains the word "appendix", the right sidebar defaults to resources for the entire part or subpart.

1. Visit `/42/441/Subpart-F/2021-11-05/#Appendix-to-Subpart-F-of-Part-441` on the experimental deploy and verify the right sidebar loads and displays "Subpart F Resources" instead of "Appendix.to resources".

### DIFF
--- a/solution/ui/regulations/js/src/components/SupplementalContent.vue
+++ b/solution/ui/regulations/js/src/components/SupplementalContent.vue
@@ -215,7 +215,11 @@ export default {
             this.fetch_content(this.title, this.part, location)
         },
         parseHash(locationHash) {
-            if (window.location.hash === "#main-content" || locationHash.toLowerCase().includes("appendix")) return "";
+            if (window.location.hash === "#main-content") return "";
+            if (locationHash.toLowerCase().includes("appendix")) {
+                this.selectedPart = undefined;
+                return "";
+            }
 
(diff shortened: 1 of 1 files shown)


PhilR8 (changes requested):
I’m just toying around with your bugfix PR.  I think there’s still one thing left to be done: updating the Subpart Resources Heading on the right sidebar when clicking away from the appendix on the left sidebar and then back to the appendix on the left sidebar.

I think we need to break out the conditional logic so that the `includes("appendix")` bit has its own condition, and then set `this.selectedPart` to `undefined` before we do our early return.

Something like this:
```
 if (window.location.hash === "#main-content") return "";
 if (locationHash.toLowerCase().includes("appendix")) {
     this.selectedPart = undefined;
     return "";
}
```

PhilR8 (approved):
LGTM!

PR #627, opened 2022-12-08
EREGCSC-1642  Error on search

thwalker6:
Resolves #616 

**Description-**

So when you do a search it errors out not because of search but because of synonyms.  We pass in synonyms as a parameter and if it has a character like "/" it thinks that there is another paramater in django.  The solution is just to add <path: synonym> on the url.

- expected change 1

Searching wont error out anymore with a "/" in the search.

Search for ICF/IID.  There should be no error on the Synonym endpoint.


PhilR8 (approved):
LGTM

PR #644, opened 2022-12-19
block public access to static assets

peggles2:
Resolves #
block public access to static assets bucket

**Description-**

Address ATO finding to set s3 buckets to block public access

- expected change 1

1. Log into AWS and navigate to the s3 buckets
2. Check that the static-assets bucket is set to blocked public access
3. Visit the website deployed on experimental yaml and check that everything is working ok.


peggles2:
bucket cleanup issues reopening in  EREGCSC-1665-s3-bucket-settings-static-assets2

PR #647, opened 2022-12-22
Regulations rss

peggles2:
Resolves #

**Description-**

- expected change 1


solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+from regcore.models import Part
+from resources.models import AbstractResource
+
+
+class PartFeed(Feed):
+    def get_feed(self, obj, request):
+        feedgen = super().get_feed(obj, request)
+        feedgen.content_type = 'application/xml'  # New standard
+        return feedgen
+    title = 'Federal Register documents RSS Feed'
```
thwalker6, line 13: I woud make put the get_feed function after description and a blank line before that.  Its not failling the linter but it should probably be done that way

solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+
+class SupplementalContentFeed(Feed):
+    title = 'supplemental content feed'
+    link = '/supplemental_content/'
+    description = 'Updates on changes and additions to supplemental content.'
+
+    def items(self):
+        return AbstractResource.objects.filter(approved=True)
+
+    # def item_title(self, item):
```
thwalker6, line 78: can this be removed?

solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+        # return item.title
+
+    # def item_description(self, item):
```
thwalker6, line 81: can this also just be removed

solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+                    'part': part,
+                    'subpart': child['identifier'][0],
+                    'last_updated': last_updated,
+                })
+            # This just ends up forwarding to the subchapter page so removing for now, but we can add back if needed
+            # if child.get('type', '') == 'section':
+            #    results.append({'title': title, 'part': part, 'section': child['identifier'][1]})
+            if child['children']:
+                childResults = self.processChildren(child['children'], title, part, last_updated)
+                if childResults:
```
thwalker6, line 32: you can get rid of the if there and just do results + childresults.  It would return an empty list anyways so adding an empty to list to a list would just return the list.  maybe just do results = results + self.processchildren(parameters)

thwalker6: Actually I would also just say pass results in as a parameter to processChildren so that it is returning the result list to the end instead of it being results = results + function it would be results = this.processChildren(results, children, title, part, lastupdated)

solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+    def get_feed(self, obj, request):
+        feedgen = super().get_feed(obj, request)
+        feedgen.content_type = 'application/xml'  # New standard
+        return feedgen
+    title = 'Federal Register documents RSS Feed'
+    link = '/latest/feed'
+    description = 'Displays the latest federal register documents'
+
+    def processChildren(self, children, title, part, last_updated):
+        results = []
```
thwalker6, line 18: If you modify the function you wont need to create an empty list, you would just be working with the results list

solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+            #    results.append({'title': title, 'part': part, 'section': child['identifier'][1]})
+            if child['children']:
+                childResults = self.processChildren(child['children'], title, part, last_updated)
+                if childResults:
+                    results = results + childResults
+        return results
+
+    def items(self):
+        date = datetime.now()
+        title = 42
```
thwalker6, line 38: Is this always jsut going to be title 42 or are we going to be including future titles.

solution/backend/regulations/rss_feed.py
```diff
@@ -0,0 +1,85 @@
...
+            children = p.structure['children']
+            results = results + self.processChildren(children, p.title, p.name, p.last_updated)
+
+        return results
+
+    def item_title(self, item):
+        if 'document_label' in item:
+            return item['document_label']
+        else:
+            return '{} {} Subpart {}'.format(item['title'], item['part'], item['subpart'])
```
thwalker6, line 58: both item_title and descriptin return the exact same value on the else statement.  Will this be consistent?  Would it make sense to just have a function called that takes those parameters for the return value so that way if it changes in one place it changes in both?

thwalker6 (changes requested):
a few notes

peggles2:
> a few notes

Thanks Thomas! This is not ready for code review yet. It was in draft mode but I moved it to a pull request to check something.

solution/backend/regulations/rss_feeds.py
```diff
@@ -0,0 +1,49 @@
+from resources.models import AbstractResource
+from django.contrib.syndication.views import Feed
+from dateutil import parser
+
+
+class ResourceFeed(Feed):
+    title = 'Federal Register documents RSS Feed'
```
cgodwin1, line 7: Should be "Resources RSS Feed" as this encompasses both Federal Register docs and Supplemental Content items.

solution/backend/regulations/rss_feeds.py
```diff
@@ -0,0 +1,49 @@
...
+    link = '/latest/feed/'
+    description = 'Displays the latest federal register documents'
```
cgodwin1, line 9: Same here

solution/backend/regulations/rss_feeds.py
```diff
@@ -0,0 +1,48 @@
...
+                'description': resource.description.replace('\x02', ''),
+                'url': resource.url,
+                'name': resource.name,
+            })
+        return results
+
+    def item_title(self, item):
+        date = item['date'].strftime("%b %d, %Y") if item['date'] else ''
+        if date and item['name']:
+            return f"{date} | {item['name']}"
```
peggles2, line 38: Every other programing language you could just do `return f"{date} | {item['name']}" if (date and item['name'])`

thwalker6 (approved):
lgtm

PR #672, opened 2023-01-20
update db instance to correct version r6

peggles2:
Resolves # wrong db instance in yml file

**Description-**
yml file had the wrong db instance. We are on r6 now. 

1. Go to yml file check that the db instance is r6 not r5
2. Go to dev and val and see that the db instance is now r6 not r5


PhilR8 (approved):
LGTM!

thwalker6 (changes requested):
Needs to be db.r6g.large

thwalker6 (approved):
lgtm

PR #677, opened 2023-01-26
EREGCSC-1656 -- Add links to GovInfo historical CFR pages

PhilR8:
Resolves [EREGCSC-1656](https://jiraent.cms.gov/browse/EREGCSC-1656)

**Description:**

Front end work for GovInfo Historical Pages feature.

- Refactors existing Copy Button Tooltip to be more generic Tooltip component
- Creates Regulation History tooltip content, populated via API call to history endpoint
- Styles everything to match [Figma Comp](https://www.figma.com/file/d4iODwrdAEMS3e45LiWi2K/Timeline-with-GovInfo-Links?node-id=1869%3A429&t=sAzCcXWoIpWrSfRr-0)

1. Visit site and go to any section in the reader view (example: [440.10](https://a01kqidbx9.execute-api.us-east-1.amazonaws.com/dev677/42/440/Subpart-A/2020-12-16/#440-10))
2. click "View § 440.10 Regulation History"
3. Make sure works as expected and matches [Figma comp](https://www.figma.com/file/d4iODwrdAEMS3e45LiWi2K/Timeline-with-GovInfo-Links?node-id=1869%3A429&t=sAzCcXWoIpWrSfRr-0)


solution/ui/regulations/js/api.js
```diff
@@ -191,17 +191,17 @@ function httpApiGetLegacy(urlPath, { params } = {}, apiPath) {
 }
 
 async function httpApiGetWithPagination(urlPath, { params } = {}, apiPath) {
-    let results = []
-    let url = urlPath
+    let results = [];
+    let url = urlPath;
     while (url) {
```
PhilR8, line 196: quite a bit of auto-formatting via Prettier in this file.

solution/ui/regulations/js/api.js
```diff
@@ -243,6 +243,29 @@ const getLastParserSuccessDate = async (apiURL, { title = "42" }) => {
...
+ * @param {Object} params - parameters needed for API call
+ * @param {string} params.title - CFR title number.
+ * @param {string} params.part - CFR part numer within title.
+ * @param {string} params.[("section"|"appendix"|"subpart")] - CFR idenfifier for node type.  Ex. for "section": "10"
+ *
+ * @returns {Array<{year: string, link: string}>}
+ */
+const getGovInfoLinks = async (apiURL, params) => {
+    // manually adjust to v3 if needed
+    const url = apiURL.replace("/v2/", "/v3/");
```
thwalker6, line 259: Just a note not a you need to do this.  This whole changing the api url to v2 to v3 should not have to be done anymore.  I understand we are converting but most of our apis are v3 not v2.  If anyhing it should be the other way around

solution/ui/regulations/js/api.js
```diff
@@ -352,27 +375,33 @@ const getSupplementalContentLegacy = async (
     return result;
 };
 
-
-const getSupplementalContentByCategory = async (api_url, categories=[1,2]) =>{
+const getSupplementalContentByCategory = async (
+    api_url,
+    categories = [1, 2]
```
thwalker6, line 380: Not your code but do we actually use the default for categories anywhere.  Having a hard coded one that is categories "1 and 2" can cause issues.  I dont think we ever do the call with no categories provided but I'm just curious if we ever do and why are we doing 1 and 2.

solution/ui/regulations/js/src/components/tooltips/CopyCitation.vue
```diff
@@ -0,0 +1,80 @@
...
+            selectedAction: null,
+            copyStatus: "idle",
+        };
+    },
+
+    watch: {
+        async copyStatus (newStatus, oldStatus) {
+            if (
+                newStatus === "pending" &&
+                (oldStatus === "idle" || oldStatus === "success")
```
thwalker6, line 50: Just trying to understand the logic. There are three statuses from what I can gather.  Is it possible for an old and new status to both be "pending" at the same time or is there another stautus from these three.  Just because if not then wouldnt if newsttus ==="pending" have the same affect.

solution/ui/regulations/js/api.js
PhilR8: not sure, but it's out of scope to debug a change like this.  This could be noted and collected in an omnibus tech debt story, like the formatting story you mentioned in slack.

solution/backend/regulations/templates/regulations/partials/node_types/SECTION.html
```diff
@@ -7,12 +7,50 @@
...
+            position="over"
+            hover
+            click
+        >
+            <copy-citation
+                slot="tooltip-content"
+                hash="{{node.label | join:'-'}}"
+                formatted-citation="{{formatted_citation}}"
+            ></copy-citation>
+        </tooltip-container>
```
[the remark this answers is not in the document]
PhilR8: This actually used to be a template fragment!  In fact you can see it in this PR:

https://github.com/CMSgov/cmcs-eregulations/pull/196/files#diff-d1d5b7e2e336cfff5825c95796febb2b1d9c65685eea884f44ae55156c180fe6

However, in the past we bumped up against template rendering performance issues in Django with heavily fragmented templates.  In fact I'm just looking back through some conversations we had on slack around the time the copy button/tooltip story (October 2021) and here are some links that were discussed:

- the Django Docs specifically mention that using `include` to add fragment templates (fragments) is not as performant as other methods like `block`: https://docs.djangoproject.com/en/4.1/topics/performance/#template-performance
- This StackOverflow discussion about template performance recommends against using ANY of the fancy Django template methods, like `includes`: https://stackoverflow.com/questions/8569387/django-guidelines-for-speeding-up-template-rendering-performance

thwalker6 (approved):

PR #680, opened 2023-01-27
Fix destruction of db issues

thwalker6:
Resolves #EREGSCSC-1713

**Description-**

Remove experimental was failing when invoked.  Django was trying to establish a connection to a db of "Main" on the instance.  Since Main no longer existed on our new DB it couldnt connect.  To fix this db_name was just changed to eregs.
Another change is to ensure on failure it can be reran to remove the database.  If the parser.yml file gets removed successfully but the db is not, you cannot simply just rerun the github action(though you can just try executing the lambda directly).  This will say "Ok this failed to remove the create parser stack, lets just remove the database directly.

- expected change 1

Create a branch off of this one and create a pr with it.  
Close the PR.  
It will successfully remove the environment/db.

Reopen the PR.
go into aws console and delete the parser lamda for the PR. 
Close the pr
Database remove step will still run.

### DIFF
--- a/.github/workflows/remove-experimental.yml
+++ b/.github/workflows/remove-experimental.yml
@@ -52,6 +52,11 @@ jobs:
           popd
       # remove the regulations site
       - name: remove experimental regulations site server
+      # If previous step fails the database will not be removed and you cannot rerun it
+      # In this instance we want to just remove the database so that we dont have 
+      # to go into the database and do it manually.
+      # If process is manually stopped it will not run this step.
+        if: success() || failure()
         env:
           PR: ${{ github.event.number }}
         run: |
--- a/solution/backend/serverless-experimental.yml
+++ b/solution/backend/serverless-experimental.yml
@@ -91,7 +91,7 @@ functions:
   drop_database:
     environment:
       STAGE: ${self:custom.stage}
-      DB_NAME: eregs
+      DB_NAME: ${self:custom.stage}
     handler: dropdb.handler
     timeout: 300
     layers:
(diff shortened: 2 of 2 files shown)


.github/workflows/remove-experimental.yml
```diff
@@ -52,6 +52,11 @@ jobs:
           popd
       # remove the regulations site
       - name: remove experimental regulations site server
+      # If previous step fails the database will not be removed and you cannot rerun it
+      # In this instance we want to just remove the database so that we dont have 
+      # to go into the database and do it manually.
+      # If process is manually stopped it will not run this step.
+        if: success() || failure()
```
peggles2, line 59: Doesn't this negate each-other? I mean the process is success or a failure. What is the purpose of the if condition?

solution/backend/serverless-experimental.yml
```diff
@@ -91,7 +91,7 @@ functions:
   drop_database:
     environment:
       STAGE: ${self:custom.stage}
-      DB_NAME: main
+      DB_NAME: eregs
```
peggles2, line 94: is this the same DB_NAME as the one defined on line 12?

peggles2 (changes requested):
Nice work, just a few comments.

.github/workflows/remove-experimental.yml
thwalker6: Manual cancellation.  So if for whatever reason I close the pr but I'm like "dont do it no", and cancel it, it wont run the step.

solution/backend/serverless-experimental.yml
thwalker6: no.  And we dont want it to be.  If you connect to the db instance using stage, so the pr name, and you drop it, it causes an error for terminating connection. So you need to connect to a database that is not the same as the one you are dropping.

peggles2 (approved):
LGTM 👍

PR #681, opened 2023-01-27
Db script format for deletion

thwalker6:
This is the shell for what to do for deleting things.  

First command will get the list of databases, then we will just run the script one at a time until all the ones we know are not being used are gone.  We should not run a loop or a lambda for deleting them. Just do it once at a time, save some potential headaches.  You must already be connected to the main database to run these commands and you should not delete a database you are currently connected to.  Connect to the main one.


peggles2 (approved):
LGTM 👍 
Just one comment, add a description about what this script is for and what it's purpose is.

PR #690, opened 2023-02-03
EREGSCSC-1702

thwalker6:
Resolves #EREGSCSC-1702

**Description-**
Step 1 for 1702 which is setting deletion policy for the rds instance and cluster to retain.  This will prevent them from being deleted when removed from the serverless.yml file.  
The database port and endpoint are also now pulled from aws app config at building of the application file to allow them to be removed in the next pr without them being deleted.

Not in PR but required change:
Two values have been added to val and prod for endpoint and port number under /eregulations/db/host and /eregulations/port.  They should match the endpoint for the database and port number respectively.

- expected change 1
Nothing really visible.    Changes will be deployed using the cloudformation process for this part but the way values are pulled in will be different.  

Deploy to DEV.    Make sure cloud formation deploys correctly for dev.  Go to DEV and make sure the site still loads correctly.  Even though the endpoints are not changing it is important to make sure that the site is still working since the values for port and endpoint are being populated in a different way.

Deploy to VAL  Make sure cloud formation properly deploys to VAL.  Then verify that the site is up just like dev.

Deploy to PROD: Make sure cloud formation properly deploys to prod.  Verify endpoint is working in prod like other environments.


peggles2 (approved):
Nice job. Looks good to me.

PR #693, opened 2023-02-06
Nodeupgrade github actions

thwalker6:
Resolves #

**Description-**

- expected change 1

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v1
         with:
-          node-version: '14.x'
+          node-version: '16.x'
       # setup python
       - uses: actions/setup-python@v2
         if: success() && steps.findPr.outputs.number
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-experimental.yml
```diff
@@ -32,9 +32,9 @@ jobs:
           state: open
       # should build first and save the artifact
       - name: Setup Node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
-          node-version: '14.x'
+          node-version: '15.x'
```
PhilR8, line 37: we should only use even-numbered releases of node -- odd-numbered versions are not promoted to LTS.  See here: https://github.com/nodejs/release#release-schedule

So we should set it to 16, not 15.

PhilR8 (changes requested):

.github/workflows/deploy-experimental.yml
PhilR8: We should probably update that in our README to be 16+ as well.

PhilR8 (approved):
LGTM!

PR #696, opened 2023-02-07
EREGCSC-1648 -- Cypress 12 Deployment Issues

PhilR8:
Resolves [EREGCSC-1648](https://jiraent.cms.gov/browse/EREGCSC-1648)

**Description:**

Cypress was recently updated to v12.  The end to end tests worked well, were much faster, and much less flaky on our Experimental Deployments, which use `deploy-experimental.yml` as a config file.

However, the tests would not run when deploying to `dev`/`val`/`prod` (known as the "production deployment").  This deployment uses a different config file -- `deploy.yml`.  

It was noted that our experimental deployments use Node v16, while our production deployments were using Node v14.  A likely explanation is that the Cypress v12 upgrade installed or upgraded dependencies that need Node 16+ to run successfully.

As such, this pull request updates Node to v18 for both the Experimental Deployments and the Production Deployments.  Node v18 was chosen because it is the [current LTS release](https://github.com/nodejs/release#release-schedule).

- Reverts exploratory changes to `deploy.yml` that added a separate job for cypress testing -- that separate job has been removed.
- Updates all steps and jobs using Node to use Node 18.14 (the current LTS version)
- Removes unnecessary Vite plugin that was causing [peer dependency errors with Node 16+ and NPM 7+](https://stackoverflow.com/questions/66239691/what-does-npm-install-legacy-peer-deps-do-exactly-when-is-it-recommended-wh).

1. Make sure tests pass and site works on Experimental Deployment
2. After merging PR into main, make sure site builds and deploys to `dev` environment correctly
3. Make sure tests pass in `dev` environment

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -275,6 +275,11 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+      # Setup node environment
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.x'
       # Run the cypress tests
       - name: end-to-end tests
         uses: cypress-io/github-action@v5
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: lts/hydrogen # 18 LTS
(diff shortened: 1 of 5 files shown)


peggles2 (changes requested):
This looks great! Thanks so much for fixing it. 
Just one update if possible update the README prerequisites to node v18 and higher.
https://github.com/CMSgov/cmcs-eregulations#prerequisites

peggles2 (approved):
LGTM 👍

PR #716, opened 2023-02-23
EREGCSC-1689: Cloudfront associated with WAF ACL

thwalker6:
Resolves # 1689 & 1690

Description- Enable WAF/ACL on Cloudfront.   Ticket was originally just for 1689 but when adding the WAF/ACL resource to serverless, it requires you to say if you want logging or not. To make it simple, logging was added.

This pull request changes...

- All new experimental deploys have waf/acl enabled as well as dev, prod and val.  
- 
 
Go to the resource for this PR in Cloudfront,
See that aws/waf has a value there.
When deployed to dev, verify it gets added.  ONce its deployed to dev it takes about 15 minutes to create the resource.
Same as above for val and prod.

### DIFF
--- a/solution/backend/resources/v3views/resources.py
+++ b/solution/backend/resources/v3views/resources.py
@@ -27,7 +27,7 @@ from regcore.views import SettingsAuthentication
 
 
 @extend_schema(
-    description="Retrieve a list ofs all resources. "
+    description="Retrieve a list of all resources. "
                 "Includes all types e.g. supplemental content, Federal Register Documents, etc. "
                 "Searching is supported as well as inclusive filtering by title, part, subpart, and section. "
                 "Results are paginated by default.",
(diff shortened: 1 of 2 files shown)


solution/backend/resources/v3views/resources.py
```diff
@@ -27,7 +27,7 @@
 
 
 @extend_schema(
-    description="Retrieve a list of all resources. "
+    description="Retrieve a list ofs all resources. "
                 "Includes all types e.g. supplemental content, Federal Register Documents, etc. "
                 "Searching is supported as well as inclusive filtering by title, part, subpart, and section. "
                 "Results are paginated by default.",
```
peggles2, line 33: Typo here.

peggles2 (changes requested):

peggles2 (approved):

peggles2 (approved):
LGTM 👍

PR #723, opened 2023-03-01
EREGCSC-1689: adding rule for firewall

thwalker6:
Resolves #EREGCSC1689

**Description-**

Adding a rule to cloudwatch firewall.  US terrirories only

- EREGS will only be accessible in the US

Will be able to access it within the US.


solution/static-assets/serverless.yml
```diff
@@ -115,6 +115,25 @@ resources:
           SampledRequestsEnabled: true
           CloudWatchMetricsEnabled: true
           MetricName: eregs-${self:custom.stage}-cloudfront-metric
+        Rules:
```
peggles2, line 118: should this also be added to the experimental-serverless.yml file?

PR #725, opened 2023-03-02
EREGCSC-1564 Revised Jump-to

thwalker6:
Resolves #EREGCSC-1564

**Description-**
Added new functionality for Jump to.  Changed it to be using a vue component as opposed to the django template for functionality.  It still is called in the django template jump-to for sidebar but that gets messy with the current css.  
This is merely for a functionality perspective.

Also fixes swagger.  Swagger was being blocked by django.

- Checks to see how many titles exist  If there is only 1 title it defaults to the one title and there is no drop down.  Should be 42 but it actually checks to the see the title available.   If it exist it will show a title dropdown.  IF there are multiple titles it will display a input with "Title" and drop downs will be "titlename CFR"
- Part down will filter by the selected title.  If defaulted to 1 title it will only have that one title(42) show up for parts related to that title.  If the title is changed it will default part back to nothing.
- After part and title are selected you can then hit go.  Not doing so ahead of time will cause it to not go anywhere.
- Section is optional.  If you do not select a section it will just take you to the TOC for the part, if title is in there it will take you to the section.
- When a part or subpart is loaded the left-sidebar will default to wherever you are at.  Since it should only be title 42 right now it will just show the part.   The css was not configured to take into account 45 right now so it will look janky if you have 45.  But you would be able to go from a title 42 to a title 45.
- Changed the form type to a class for continuity of CSS.  This will change soon anyways.

Go to the homepage on the environment.  It should look mostly the same as the prod site.  There is a slight difference in spacing aroudn the dot and some minor css was added to try and make it look as similar as possible(its pretty close).
Hit go.  it will not go anywhere.
Select a part, hit go, will take you to the part toc. Left sidebar should autodefault to having part 400 selected for part.
Go back to the homepage, put in a part and a section.  It will take you to the section.  
Left sidebar should auto fill with the part(section is empty).

Run it on your local and make sure title 45 exist.
There should be a title field for the jump-to.  Select a title.  hit go
Should not navigate
Select a part, click go, should go to the toc.
Sidebar should auto fill in the part and the title.
Go back to home, select title, and part fill in a valid section.  Should take you to the section.  Sidebar should populate title and part.


solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
```diff
@@ -0,0 +1,158 @@
...
+    </div>
+</template>
+<script>
+import { getTitles, getParts } from "../api";
+
+export default {
+    name: "JumpTo",
+
+    props: {
+        apiurl: {
```
PhilR8, line 70: this should be camelCased

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+    <div>
+        <div class="jump-to-label">Jump to Regulation Section</div>
+        <div class="form">
```
PhilR8, line 4: We probably should try to keep this as a semantic `form` element if possible, since it is helpful for a11y and assistive tools like screen readers.

We're using forms in at least one other place in the app.  You can look at the example here:

https://github.com/CMSgov/cmcs-eregulations/blob/main/solution/ui/regulations/eregs-vite/src/components/SearchInput.vue

The main thrust is that instead of using `@click` to invoke the `getLink` method, you'd use `@submit.prevent` on the form itself.

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
```diff
@@ -0,0 +1,158 @@
...
+
+    methods: {
+        async getParts(title) {
+            const partsList = await getParts(this.apiurl, title);
+            this.filteredParts = partsList.map((part) => part.name);
+        },
+        getLink() {
+            let link = `${this.base}/goto/?title=${this.selectedTitle}&part=${this.selectedPart}`;
+            if (this.selectedSection !== "") {
+                link += `&section=${this.selectedSection}&${this.selectedpart}-version='latest'`;
```
PhilR8, line 151: `selectedpart` needs to be camelCased here; it's showing up as `undefined` in the `goto` URL.  Which is actually still taking you to the right place lol

PhilR8 (changes requested):
I've requested some changes, mostly minor. There's also a small bug that I'll detail below.  In general, though, this is really nice work. 

The bug: if you don't select a part but add a section, the user is still able to click GO.  

The `goto` URL is requested without a part:

https://97ihhcvo4d.execute-api.us-east-1.amazonaws.com/dev725/goto/?title=42&part=&section=10&undefined-version=%27latest%27

which returns a 500 error.

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
PhilR8: I actually didn't realize this `goto` url path existed -- I love that you can specify `latest` and it will go to the latest version.  I feel like I should have used this elsewhere in the Search/Resources pages.

thwalker6: Im thinking it works because even though we have it say part-version="latest" we dont need it to.

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
```diff
@@ -0,0 +1,158 @@
...
+                    id="jumpto_title"
+                    v-model="selectedTitle"
+                    name="title"
+                    aria-label="Regulation title number"
+                    class="ds-c-field"
+                    required
+                >
+                    <option value="" disabled selected>Title</option>
+                    <option v-for="t in titles" :key="t" :value="t">
+                        {{ t }} CFR
```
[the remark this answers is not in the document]
thwalker6: This was to satisy eslint.  i would have done "tit" but didnt seem professional and thats honestly how I would have short handed it.

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
thwalker6: Actually this worked because it was undefined!  just want -version=latest, no part number

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
thwalker6: done

solution/ui/regulations/eregs-component-lib/src/components/JumpTo.vue
PhilR8: Gotcha, that makes sense.

PhilR8 (approved):
LGTM!  Nice job.

PR #726, opened 2023-03-02
EREGCSC-1755 -- Highlighting Bug

PhilR8:
Resolves [EREGCSC-1755](https://jiraent.cms.gov/browse/EREGCSC-1755)

**Description**
Parentheses and other special characters cause problems when using them to highlight text in the reg text reader view.

- Updates regex logic to handle parens and other special characters
- limits highlightable terms to strings with a length greater than one (1)
- Adds actual user-defined query as a highlightable string in the URL
- Ignores citation node at bottom of section when highlighting text

1. Search for "1903(r)", click a result to visit the reader view, and make sure "1903(r)" is highlighted in the reg text
2. Make sure all previous searches work


thwalker6 (approved):
lgm

PR #733, opened 2023-03-13
EREGCSC-1744: Minimum viable homepage updates for 45

thwalker6:
Resolves # EREGCSC-1744

**Description-**

Minimum updates to make the homepage ready for title 45

-Adjust the spacing between the end of Title 42 and the start of Title 45 (it's a bit squished)

-Remove all of the hardcoded "preceding" and "subsequent" notes from the homepage – they're outdated anyway, and probably not super effective; we should deal with that problem in a different way

1. Go to the homepage
2. Scroll down to title 45.  Spacing should be more than it was before.  Will look less squished.  Margin for top of h1 was changed from 0px to 16px.  This was taken from margin-top of the toc so that the first title doesn't move from where it was previously while adding padding for subsequent titles.

4. There should be no preceding or subsequent notes on the homepage now.

### DIFF
--- a/solution/backend/regulations/templates/regulations/homepage.html
+++ b/solution/backend/regulations/templates/regulations/homepage.html
@@ -38,13 +38,13 @@
                     <div class="toc-container">
 
                         {% for title in structure %}
-
+                            <div class="toc-title-container">
                             <h1> {{title.label}} </h1>
 
                             {% for child in title.children %}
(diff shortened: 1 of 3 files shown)


peggles2 (approved):
LGTM

thwalker6:
Waiting on the validation portion before merging.

PR #734, opened 2023-03-13
EREGCSC-1547 -- New Header

PhilR8:
Resolves [EREGCSC-1547](https://jiraent.cms.gov/browse/EREGCSC-1547)

**Description**

New header component created using Vue, not a Django template.

* Header now a Vue component, used directly in Django templates as well as within the Resources and Search SPAs
* New components use the Vue Composition API, which is available as of Vue 2.7 and in Vue 3. See more [here](https://vuejs.org/guide/extras/composition-api-faq.html#what-is-composition-api)
* Header and Jump To styles have been updated to match [Figma comp](https://www.figma.com/proto/Gx6pz0IxWqGsweFVdezKNe/eRegs-Homepage---Multi-View-Update?page-id=2585%3A11042&node-id=2585-11833&viewport=-991%2C3093%2C0.08&scaling=scale-down&starting-point-node-id=2585%3A11833&show-proto-sidebar=1)
* Some reorganization of Vue components -- for example, JumpTo.vue has been moved into `eregs-vite`.
   - Justification for this move can be found here at [EREGCSC-1782](https://jiraent.cms.gov/browse/EREGCSC-1782) -- the eventual goal is to move all components and api/utility methods into a single project (`eregs-vite`) and export them to be used in that project and any other projects (such as `eregs-component-lib`).  This will make things much easier and reduce quite a bit of duplication.
* Created new `directives` folder to contain custom Vue directives that can be globally imported into both Vite projects.  
   - The `clickaway` directive is now its own file and can be used on any Vue component in either project.
 
1. Visit experimental deployment
2. Navigate around site using the header and make sure everything works


solution/ui/regulations/eregs-component-lib/src/components/index.js
```diff
@@ -33,4 +32,8 @@ export { default as CopyCitation } from "./tooltips/CopyCitation.vue";
 export { default as GovInfoLinks } from "./tooltips/GovInfoLinks.vue";
 export { default as TooltipContainer } from "./tooltips/TooltipContainer.vue";
 
-// Other components you plan to publish go here
+// from Single Page App
+export { default as HeaderComponent } from "../../../eregs-vite/src/components/header/HeaderComponent.vue";
+export { default as HeaderLinks } from "../../../eregs-vite/src/components/header/HeaderLinks.vue";
+export { default as HeaderSearch } from "../../../eregs-vite/src/components/header/HeaderSearch.vue";
+export { default as JumpTo } from "../../../eregs-vite/src/components/JumpTo.vue";
```
PhilR8, line 39: aliases in both `vite.config` files still need some ironing out to cut down on the need for relative paths like this

solution/ui/regulations/eregs-component-lib/src/components/tooltips/TooltipContainer.vue
```diff
@@ -82,41 +82,6 @@ const leftWarning = (el) => el.getBoundingClientRect().left < 130;
 export default {
     name: "TooltipContainer",
 
-    components: {},
-
-    // https://www.vuesnippets.com/posts/click-away/
-    // https://dev.to/jamus/clicking-outside-the-box-making-your-vue-app-aware-of-events-outside-its-world-53nh
-    directives: {
-        clickaway: {
```
PhilR8, line 90: this custom directive is being moved to its own file so it can be injected at the App level for use in both `eregs-vite` components and `eregs-component-lib` components

solution/ui/regulations/eregs-vite/src/components/header/HeaderComponent.vue
```diff
@@ -0,0 +1,33 @@
+<script setup>
+import { ref, onBeforeMount, onMounted, onUnmounted, onUpdated } from "vue";
+
```
PhilR8, line 3: need to clean up unnecessary imports and lifecycle methods in this file

solution/ui/regulations/eregs-vite/src/components/header/HeaderLinks.vue
```diff
@@ -0,0 +1,84 @@
+<script setup>
+import { ref } from "vue";
+
+import HeaderChevronUp from "../svgs/header-chevron-up.vue";
+import HeaderChevronDown from "../svgs/header-chevron-down.vue";
```
PhilR8, line 5: as mentioned above, relative pathing is needed for any components that are used in `eregs-component-lib` -- ironing out the aliases in the config for both Vite pipelines is TODO

solution/ui/regulations/css/scss/partials/_header.scss
```diff
@@ -22,302 +35,337 @@ header {
...
+                @include custom-min($eds-width-sm / 1px) {
                     top: $header_height;
                 }
             }
         }
     }
 
-    display: flex;
-    flex-direction: row;
+    .header--content-container {
```
PhilR8, line 45: very first attempt at using BEM: https://getbem.com/naming/

solution/ui/regulations/eregs-vite/src/components/JumpTo.vue
```diff
@@ -1,15 +1,13 @@
...
         <form @submit.prevent="getLink">
             <div class="jump-to-input">
                 <select
                     v-if="defaultTitle === ''"
-                    id="jumpToTitle"
                     v-model="selectedTitle"
                     name="title"
                     aria-label="Regulation title number"
-                    class="ds-c-field"
+                    class="jump-to-title ds-c-field"
```
thwalker6, line 10: Is it better to have this as a class than an ID?  On the cypress test we end up doing a .first() which to me in the sense of testing implies that the same elemenet can be used multiple times on the same page, which is not the case.  jump to should only be on the header.  If there is a .second(I'm either making this up or this is a thing) that would be bad.  The Id's allow us to with consistency apply values to those boxes.  I'm sure either way is fine and is just a philosopical difference but I could be wrong

thwalker6:
header update

solution/ui/regulations/eregs-vite/src/components/JumpTo.vue
PhilR8: having it as an ID when there were two instances of the Jump To on the homepage failed the accessibility tests because, as you say, having an ID implies something can only be used once on a page.  But since there will be a brief moment in time where there are two Jump To components on the homepage, these were changed to classes.

These can be changed back to IDs as part of the Acceptance Criteria for the future story where we remove the Jump To component from the homepage hero banner.

PhilR8:
> header update
> 
> After the jump-to is selected, part needs to go back to "Part" and the go button should go back to white.

That's not how the Jump To component currently works on prod.  I didn't touch the Jump To logic because that was already done in a [EREGCSC-1564](https://jiraent.cms.gov/browse/EREGCSC-1564).  Any changes to how the Jump To component works is out of scope for this story.

solution/ui/regulations/eregs-vite/src/components/JumpTo.vue
PhilR8: Although I'm talking through the homepage updates with Jonathan now and it might just make sense to remove the Jump To from the Hero banner in this story, which would allow me to revert these classes back to IDs and address your concerns here.

PhilR8: I've changed these back to IDs and updated the tests.  Resolved!

thwalker6 (approved):

PR #735, opened 2023-03-14
EREGCSC 1548- Reg page left nav shows part info 

thwalker6:
Resolves # EREGCSC-1548

**Description-**

Removing jump -to from right sidebar replace with text for part and subchapter

- Right sidebar jump to is gone and replaced with text


PhilR8 (approved):
LGTM

PR #745, opened 2023-03-24
EREGCSC-1708 -- fix prod cypress failures

PhilR8:
Resolves [EREGCSC-1708](https://jiraent.cms.gov/browse/EREGCSC-1708)

**Description**
Our unique production environment has two URLs that can cause headaches for tests and api calls.  This PR will attempt to mitigate these issues to cause fewer headaches.

- Removed complicated and brittle logic that determines the proper `base` URL on the front end
- Pass in `homepage` URL from Django instead, which should always just know if the URL for the site has a subdirectory or not

1. Build locally and make sure the Jump To in the header works from every route (`/`, `/resources`, `/search`, `/about`, ToC for a part, reader view for a subpart)
2. Build locally and make sure highlighting in a Subpart section works when visiting from the search page
3. Make sure all tests pass on experimental deployment
4. make sure all tests pass in `dev` and `val` when merged into `main`
5. make sure tests pass on `prod` when approved to deploy to `prod`
6. Visit https://regulations-pilot.cms.gov/ and make sure the Jump To in the header works from every route (`/`, `/resources`, `/search`, `/about`, ToC for a part, reader view for a subpart)
7. Visit https://regulations-pilot.cms.gov/ and make sure highlighting in a Subpart section works when visiting from the search page

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/views/Search.vue
+++ b/solution/ui/regulations/eregs-vite/src/views/Search.vue
@@ -10,7 +10,7 @@
         <header id="header" class="sticky">
             <HeaderComponent :home-url="homeUrl">
                 <template #jump-to>
-                    <JumpTo />
+                    <JumpTo :home-url="homeUrl"/>
                 </template>
                 <template #links>
                     <HeaderLinks
--- a/solution/ui/regulations/eregs-vite/src/views/Search.vue
+++ b/solution/ui/regulations/eregs-vite/src/views/Search.vue
@@ -201,6 +201,10 @@ export default {
             type: String,
             default: "/about/",
         },
+        apiUrl: {
+            type: String,
+            default: "/v3/",
+        },
         homeUrl: {
             type: String,
             default: "/",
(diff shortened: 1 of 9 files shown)


solution/ui/regulations/eregs-vite/src/views/Search.vue
```diff
@@ -240,10 +244,6 @@ export default {
     },
     data() {
         return {
-            base:
-                import.meta.env.VITE_ENV && import.meta.env.VITE_ENV !== "prod"
-                    ? `/${import.meta.env.VITE_ENV}`
-                    : "",
```
PhilR8, line 246: I initially wrote this logic, and now I am crusading to expel it from the app.  

I've written another story to handle the API methods file, since it is beyond the scope of this bugfix story, and also will require refactoring the `fetchJson` api method that is used as the basis for all API calls from the Single Page Apps.  So, a bigger job that needs a dedicated tech debt story.

thwalker6 (approved):

PR #746, opened 2023-03-24
EREGCSC-1779 Searching for certain characters causes error

cgodwin1:
Resolves #1779

**Description-**

Searching for some characters caused search to fail with an error popup. This was due to the synonyms endpoint using a path parameter (e.g. `/v3/synonym/X`) which prevented special characters from being encoded properly. (E.g. `%` should be encoded to `%25` but wasn't). This PR changes it to use a query parameter (`/v3/synonyms?q=X`) which is encoded properly.

- Synonyms endpoint uses query parameter
- Cypress test updated
- api.js file updated

1. Visit the experimental deploy and search for the following: `%`, `138% FPL`, `138 % FPL`, `"`, `""`, `#`, `.`, `..` , `?` 
2. Verify for each of the above, there are no errors

### DIFF
--- a/solution/ui/e2e/cypress/e2e/api.spec.cy.js
+++ b/solution/ui/e2e/cypress/e2e/api.spec.cy.js
@@ -18,7 +18,7 @@ const API_ENDPOINTS_V3 = [
     `/v3/resources/locations/sections`,
     `/v3/resources/locations/subparts`,
     `/v3/resources/supplemental_content`,
-    `/v3/synonym/${SYNONYM}`,
+    `/v3/synonyms?q=${SYNONYM}`,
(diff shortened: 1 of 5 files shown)


peggles2 (dismissed):
Looks good. Can we also add a test for this as well?

solution/ui/e2e/cypress/e2e/api.spec.cy.js
```diff
@@ -47,3 +50,16 @@ describe("API testing", () => {
         });
     });
 });
+
+describe("Synonyms endpoint special character testing", () => {
+    SPECIAL_CHARACTERS.forEach(character => {
```
PhilR8, line 55: very nice

thwalker6 (approved):

PR #747, opened 2023-03-24
Ereggcsc 1760 - metadata full text resources results - (original had problems) 

peggles2:
Resolves #[EREGCSC-1760](https://jiraent.cms.gov/browse/EREGCSC-1760)

Description-
Setup resources to query search.gov and from the list of results get the rest of the data from our database.
Add snippet to the result set. But for now the ui will just show description. Our next step we will display both snippet and description.

This pull request changes...

update resources.py
add a route in urls.py
Create serializer to return search.gov's snippets that we apply to the response
Update api.js so that if it finds searchgov in the url to call the route path that looks it up with searchgov, otherwise to call the original route path.

steps to view and verify change
Visit https://dv9wcjg6r3.execute-api.us-east-1.amazonaws.com/dev740/v3/resources/search?q=medical
Notice the data contains "Snippet" and "Description"
visit https://dv9wcjg6r3.execute-api.us-east-1.amazonaws.com/dev740/search/?q=medical&searchgov
notice the url has searchgov so the page will make a request to search.gov. Notice description is there but snippet is not there yet. (You can saw that in the api call in step 1)
click next page and check see if pagination working correctly
visit https://dv9wcjg6r3.execute-api.us-east-1.amazonaws.com/dev740/search/?q=ex%20parte
notice only 1 search result which is coming from our db
Visit https://dv9wcjg6r3.execute-api.us-east-1.amazonaws.com/dev740/search/?q=ex%20parte&searchgov
notice lots more results as search.gov can retrieve results from pdf files as well.


solution/backend/resources/views/resources.py
```diff
@@ -34,6 +42,97 @@
...
+
+    def list(self, request, *args, **kwargs):
+        q = self.request.query_params.get("q")
+        page = int(self.request.query_params.get("page", 1))
+        url = request.get_full_path()
+        self.get_gov_results(q, page)
+        urls = dict([(i['url'], i["snippet"]) for i in self.gov_results["results"]])
+        queryset = self.get_queryset(urls.keys())
+        resources = list(queryset)
+        urls = dict([(i['url'], i["snippet"]) for i in self.gov_results["results"]])
```
thwalker6, line 118: this line got duplicated for assigning the values for URLs at some point.  Just remove it since the proper value for urls is already assigned on 115.

PR #749, opened 2023-03-27
Syncing up github actions

thwalker6:
Resolves #

**Description-**

Some of the steps were using different versions of github actions.  All processes should work fine.  We updated node for other steps but missed a couple and some are outdated using deprecated github commands and newer versions removed that.

- Set output is no longer used.  Some of the older libraries used it still

1. Close the PR.  Removal step should most likely work properly.  There is a weird bug there that might stop it but it should work if it since it was fully built.
2. Reopen the PR.  The build steps should all execute(make sure the removal step is finished running first)


peggles2 (approved):
LGTM 👍

PR #752, opened 2023-03-28
Add Cypress test suite for Error Page

PhilR8:
Resolves error pages not having any end to end test coverage

**Description**

The header for every page on the app was recently redesigned and refactored (see [EREGCSC-1547](https://jiraent.cms.gov/browse/EREGCSC-1547) ).  Unfortunately, the Error pages did not get the updated header and for the past several days these error pages did not have a header at all.  

The header was added back to the error pages via a hotfix pull request (see [here](https://github.com/CMSgov/cmcs-eregulations/pull/751) ).

This pull request will add tests so that if header functionality or body text should change (like if the header disappears again), then the tests will fail.

**This pull request changes**

- adds new test suite for 404 error page

1. Make sure tests pass on a hosted environment

### DIFF
--- a/solution/backend/regulations/templates/error_base.html
+++ b/solution/backend/regulations/templates/error_base.html
@@ -33,7 +33,7 @@
             <h1 class="error-header">{% block error_header %}{% endblock %}</h1>
             {% block error_body %}{% endblock %}
         </div>
-        <img class="error-image" src="{% static 'images/browser.svg' %}" />
+        <img class="error-image" src="{% static 'images/browser.svg' %}" alt=""/>
     </main>
 {% endblock %}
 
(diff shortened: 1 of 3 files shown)


solution/backend/regulations/templates/error_base.html
```diff
@@ -33,7 +33,7 @@ <h5 class="error-code">Error {% block error_code %}{% endblock %}</h5>
             <h1 class="error-header">{% block error_header %}{% endblock %}</h1>
             {% block error_body %}{% endblock %}
         </div>
-        <img class="error-image" src="{% static 'images/browser.svg' %}" />
+        <img class="error-image" src="{% static 'images/browser.svg' %}" alt=""/>
```
PhilR8, line 36: https://www.w3.org/WAI/tutorials/images/decorative/

thwalker6 (approved):

PR #753, opened 2023-03-28
EREGCSC-1823 Missing results in regs search

thwalker6:
Resolves # EREGCSC-1823

**Description-**

Certain part versions are missing in regulations search due to the index not containing all of them particularly the most recent version.  This is a data migration to just save each part as it currently is.  Whenever a part is saved it updates in the index

- More results will appear in serach

1. Pull the branch onto your local.  Run the data migrations.  Search for FMAP.  You should get more than 3 results.  
2. Cant really properly test until its in dev.  In dev search for FMAP.  Should get more results
3. In val as well should get more than 3 results
4. In prod should get part 431 appearing in search.

### DIFF
--- a/solution/backend/regcore/search/migrations/0007_populate_index.py
+++ b/solution/backend/regcore/search/migrations/0007_populate_index.py
@@ -1,0 +1,19 @@
+from django.db import migrations
+from regcore.models import Part
+
+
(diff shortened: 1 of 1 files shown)


peggles2 (approved):
LGTM

PR #755, opened 2023-03-29
EREGGCSC-1819-searchgov-results-sorted

peggles2:
Resolves # EREGCSC1819

**Description-**
Match search.gov results sorting with the queryset returned.

- Ordering of results on the resources page will match search.gov result order

1. Make a search on the resources page with &gov to use search.gov results.
2. Make the api call for relevance.  
3. Ordering should be the same.
4. Mae sure seed data matches prod or you will have missing results, the order would still be correct.


solution/backend/resources/views/resources.py
```diff
@@ -54,59 +60,84 @@ def get_annotated_url(self):
...
             "total": gov_results["web"]["total"],
             "results": results,
         }
 
+    def get_gov_results(self, query, page):
+        key = 'M1igE4Qcfo8LLQr7o_I9KLA6qkybmlC9IRhVCCbFbl4='
+        offset = (page - 1) * self.limit
+
+        rstring = f'https://search.usa.gov/api/v2/search/?affiliate=reg-pilot-cms-test&access_key={key}' \
+                  f'&query={urlparse.quote_plus(query)}&limit={self.limit}&offset={offset}'
```
thwalker6, line 83: we need to encode it again on the backend. Django decoded the query from the front end which would work but since we have to make an external search it gets chopped off on the way to search.gov

thwalker6 (approved):

PR #756, opened 2023-03-30
Change how vector column is called in

thwalker6:
Resolves #

**Description-**

- expected change 1

### DIFF
--- a/solution/backend/regcore/search/models.py
+++ b/solution/backend/regcore/search/models.py
@@ -22,12 +22,11 @@ class SearchIndexQuerySet(models.QuerySet):
             cover_density = True
         search_query = SearchQuery(query, search_type=search_type, config='english')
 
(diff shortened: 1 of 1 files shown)


solution/backend/regcore/search/models.py
```diff
@@ -22,12 +22,11 @@ def search(self, query):
...
 
-        return self.annotate(vector_column=RawSQL("vector_column", [], output_field=SearchVectorField()))\
-            .annotate(rank=SearchRank(
-                "vector_column",
+        return self.annotate(rank=SearchRank(
+                RawSQL("vector_column", [], output_field=SearchVectorField()),
                 search_query, cover_density=cover_density)
             )\
-            .filter(rank__gte=0.04)\
+            .filter(rank__gte=0.2)\
```
thwalker6, line 29: This adjustment got ranking to be higher, to around what it was recently so increasing the threshold of rank makes sense.  Before it got really low for some reason.(like scientific notation low in some cases)

PR #764, opened 2023-04-06
EREGCSC-1477 -- Title select on Resources page

PhilR8:
Resolves [EREGCSC-1477](https://jiraent.cms.gov/browse/EREGCSC-1477)

**Description**

If multiple titles are available (currently fetched from the `/v3/titles` endpoint), then there will be a Title filter on the Resources page that allows users to switch between titles.

UPDATED BEHAVIOR:
- Behavior is unchanged if only one title is available
- If there is more than one title available to the site (e.g. 42 and 45), then a new TITLE filter will appear on the Resources page
- If there is more than one Title available, no title will be selected by default
   - User can do keyword/text search and Category filtering without selecting a Title
   - User cannot filter by part/subpart/section before selecting a title
- Title filter is unique and not part of the "other" filters (Part, Subpart, etc)
   - Title does not have a Chip
   - When clicking the "Clear All" Chip, Title is not removed or reset
   - Once you select a Title, you cannot "unselect" it without leaving and returning to the page
- Download CSV results will pull from all resources if no title is selected, but if a title is selected, the results will be limited to that title.

1. Compare to Figma [here](https://www.figma.com/file/Gx6pz0IxWqGsweFVdezKNe/eRegs-Homepage---Multi-View-Update?node-id=3926-13274&t=vXFRokdXIlh5iLV6-0)
2. Poke around and make sure things work as expected
3. All end to end tests pass


solution/ui/regulations/eregs-vite/src/utilities/api.js
```diff
@@ -281,9 +281,43 @@ const setCacheItem = async (key, data) => {
 };
 
 // ---------- api calls ---------------
-const getLastUpdatedDates = async (apiUrl, title = "42") => {
-    const result = await httpApiGet(`title/${title}/parts`);
-    return Object.fromEntries(new Map(result.map((obj) => [obj.name, obj.date])));
+const getPartTOC = async (title, part) =>
+    httpApiGet(`title/${title}/part/${part}/version/latest/toc`);
+
```
PhilR8, line 286: these smaller methods were moved higher in the file to avoid linting error where method was used before it was declared.  It works either way due to Javascript's [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting), but it's nice to please the linter and declare methods before using them.

solution/ui/regulations/eregs-vite/src/utilities/api.js
```diff
@@ -359,66 +406,71 @@ const getRegSearchResults = async ({
...
-    }
-) => {
+const getSupplementalContent = async ({
+    partDict,
+    title,
+    categories,
+    q = "",
+    start,
+    max_results = 1000,
+    paginate = true,
```
thwalker6, line 426: should we switch these to camelcasing here.  ESLINT does not like it, i know its outside the scope but we are touching this so it would make sense to clean it up.  if you feel it is outside of scope though I'm fine with that too.

PhilR8: I think this is a good place where the snake_cased params could be switch to camelCase.  I think it's better as a tech debt story/PR though.  This block of params only changed because Prettier formatted it.

thwalker6 (approved):
lgtm

PR #767, opened 2023-04-07
Removing failed unittest

thwalker6:
Resolves # none

**Description-**

Removes a failing python unittest due to retired models

 All current python unittest should pass

- Run make python.unittest.  All test should pass


peggles2 (approved):
Looks good! Just add a description on why we are removing and its good to merge! Thanks!

PR #785, opened 2023-04-13
EREGCSC-1800 Medium - dev,prod,val RDS.11 RDS instances should have automatic backups enabled

cgodwin1:
Resolves #1800

**Description-**

To satisfy ATO requirements, we must have a minimum database backup period of 7 days.

- Sets `BackupRetentionPeriod` and `BackupRetentionMinimum` to 7

1. Run `serverless package --stage dev` and verify no errors
2. Run `serverless package --stage val` and verify no errors
3. Run `serverless package --stage prod` and verify no errors
4. Deploy this PR and check AWS console to make sure backups are being retained for 7 days

### DIFF
--- a/solution/backend/serverless.yml
+++ b/solution/backend/serverless.yml
@@ -233,7 +233,8 @@ resources:
         Engine: aurora-postgresql
         EngineVersion: "15.2"
         DatabaseName: 'eregs'
-        BackupRetentionPeriod: 3
+        BackupRetentionPeriod: 7
+        BackupRetentionMinimum: 7
         DBClusterParameterGroupName:
           Ref: AuroraRDSClusterParameter15
         VpcSecurityGroupIds:


solution/backend/serverless.yml
```diff
@@ -233,7 +233,8 @@ resources:
         Engine: aurora-postgresql
         EngineVersion: "15.2"
         DatabaseName: 'eregs'
-        BackupRetentionPeriod: 3
+        BackupRetentionPeriod: 7
+        BackupRetentionMinimum: 7
```
thwalker6, line 237: I dont see this paramater in the documenation anywhere BackupRetentionMinimum.

thwalker6 (changes requested):

solution/backend/serverless.yml
cgodwin1: Go to https://docs.aws.amazon.com/securityhub/latest/userguide/rds-controls.html#rds-11 it lists the parameter here. Then click on "AWS Config rule: [db-instance-backup-enabled](https://docs.aws.amazon.com/config/latest/developerguide/db-instance-backup-enabled.html)" and it'll show the list of possible params.

thwalker6: Ok, this seems with an out of sync thing with AWS documentation so it makes me a little bit on the "make sure this works correctly".  When pushed be on the look out if it recognizes the paramater.

thwalker6 (approved):
LGTM

PR #789, opened 2023-04-14
EREGCSC-1837 Make FR-type an editable field

thwalker6:
Resolves # EREGCSC-1837

**Description-**

Allows doc_type to be editable for FR-documents in django admin.  Admin is only editable as a drop down value.  If there is an already existing value, that value will still appear in the drop down until it is changed.
Options are:

RFI
NPRM
Final Rule
None

- Doc-type is editable in fr-documents

Log into django admin
Select an FR-document
Doc Type will be editable through a drop down.
Select a different doc type
reopen FR-document to verify it changed.

### DIFF
--- a/solution/backend/resources/admin.py
+++ b/solution/backend/resources/admin.py
@@ -320,7 +320,7 @@ class SupContentForm(ResourceForm):
 
 
 class FederalResourceForm(ResourceForm):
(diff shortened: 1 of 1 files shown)


peggles2 (approved):
LGTM 👍

PR #802, opened 2023-04-21
instead of end needs to be fi

peggles2:
Resolves #

**Description-**

There was a bug in the script instead of fi it had 'end' which is wrong syntax for a shell script

- expected change 1

1. run the script manually on your local and notice it does not give an error anymore.


cgodwin1 (approved):
LGTM

PR #806, opened 2023-04-27
EREGCSC-1814 Sync up prod data

thwalker6:
Resolves # EREGCSC-1814

**Description-**

- Adds a command for getting wanted seed data from prod(resource data)
- Adds a make command for dumping seed data
- Changes local.seed to loop through a list instead of listing out all objects successively
- Separates puts abstract object types into its own json file instead of sharing with child.

 Below instructions are also in the readme file that was added.
Make sure you have pg_dump, a way to run sql, and your vpn ready.

To get up to date seed data you need to do three things and have a couple requirements. 
    1.  The ability to connect to our vpn
    2.  The database connection address and port number
    3.  A running docker container of our database
    4.  pg_dump, you can set this up in linux, windows, or mac.

The pg_dump command will pull in all of resources app in djangos tables as well as the synonym tables in the search app.  It pulls it in as a sql command inesert into column.  It will pull in some pg_audit stuff at the end that will cause an error but the rest of it will run fine and updates the tables as expected.  The next part is to clear out the resources tables and the synonym table which is done by an additional script.  The final step is to run the sql file created by pg_dump.

First run the pgdump command below.  Replace the db_name with the prod db address, and the port number for that port number.  Make sure you are also connected to the vpn in order to access AWS

pg_dump -h <db_address> -p <port_number> -U eregsuser -f backup.sql -t 'search_sy*' -t 'resources_*' --data-only --column-inserts eregs

The output for the backup.sql file will be wherever you ran the command from.  You can rename the backup.sql file if you wish but that is what its called in the command above.

Next run the clear-tables.sql file in this folder.  It will remove any currrent supplemental content, fr-docs, locations, synonyms files.  Does not touch regcore stuff or any admin tables.

Run the backup.sql file on your local database.  You can do this through command line or through a database tool.  Just make sure it has postgres capabilities.

To update our fixture data run the make command "make local.dump".  This will use django to create the json files for the seed data.

To test that the seed data is working properly run "make local.seed".  Everything should be working correctly once uploaded.


solution/_templates/scripts/get-seed-data/clear-tables.sql
```diff
@@ -0,0 +1,15 @@
...
+delete from resources_category; 
+delete from resources_federalregisterdocument; 
+delete from resources_federalregisterdocumentgroup; 
+delete from resources_resourcesconfiguration; 
+delete from resources_section; 
+delete from resources_subcategory; 
+delete from resources_subpart; 
+delete from resources_supplementalcontent; 
+delete from search_synonym; 
+delete from search_synonym_synonyms;
```
peggles2, line 15: I am not comfortable with just having a script that deletes the data to be updated on its own. I think this needs to be tightened up a little.
1. put something in place that says "are you sure and this is what this script will do" 
2. Add dump command before the delete right in here so its always backing up if you run this.
3. Put a check in here that checks the environment so that this can only be run on dev and not on prod or val. 
4. Write tests

solution/Makefile
```diff
@@ -161,20 +161,21 @@ python.unittest:
 python.shell:
 	docker-compose exec regulations python manage.py shell
 
+python.deletetables:
+	docker-compose exec regulations python manage.py cleantables
```
cgodwin1, line 165: Rename `python.deletetables` to cleantables for consistency. We're not deleting the table

peggles2 (approved):
LGTM 👍

PR #808, opened 2023-04-28
EREGCSC-1868: Outbound link don' use redirects

thwalker6:
Resolves # EREGCSC 1868

Description-
Outbound links on show more are using the redirects instead of the url for the abstract resources. Needs to be fixed.

This pull request changes...

Supplemental content on showmore will just go to the outbound site instead of a redirect

steps to view and verify change

Go to a subpart for regulations.
Go to the sidebar and pick a resource category  that has a show more.
Url should be a normal url not one for a generic piece of supplemental content.

### DIFF
--- a/solution/backend/resources/serializers/resources.py
+++ b/solution/backend/resources/serializers/resources.py
@@ -1,30 +1,39 @@
 import re
 
-from django.urls import reverse
-from drf_spectacular.utils import extend_schema_field, OpenApiTypes
 from django.db.models import Q
+from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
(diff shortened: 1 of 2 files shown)


solution/backend/resources/serializers/resources.py
```diff
@@ -1,31 +1,40 @@
 import re
 
-from django.urls import reverse
-from drf_spectacular.utils import extend_schema_field, OpenApiTypes
 from django.db.models import Q
+from drf_spectacular.utils import extend_schema_field
```
thwalker6, line 4: Up here I removed OpenApiTypes , and reverse since neither are used anymore.  The rest I simply reordered.

peggles2 (approved):
LGTM

PR #809, opened 2023-05-02
EREGCSC-1837 Admin dropdown for new content

thwalker6:
Resolves # EREGCSC-1837

**Description-**

Restricts input on admin panel for document type to be a drop down.  Will not override current values.  Normally you would put this in the model but if any current values do not match up to it, if done in the model, you will be forced to change it upon editing it.

- When clicking "Add Federal Register Document" in the admin panel, you will be given a drop down for document type.

Log into the admin panel.
Go to federal register document.
Click "Add Federal Register Document".
On Doc type it should be a drop down field.


peggles2 (approved):
LGTM 👍

PR #813, opened 2023-05-04
EREGCSC-1843 Automatically link to SSA in the reg text

cgodwin1:
Resolves #1843

**Description-**

This PR implements linking directly to sections on the uscode.house.gov website. It implements a Django template filter that uses regex to scan for all currently known formats of statute references, such as: "Section 1902(a) of the Act", "Sections 1902(a) and 1903(b)", "Section 1902(a)(b)(c) and (d) of the Affordable Care Act", and so on. The regex then substitutes the section refs with links.

- If Statute Link Converters exist, you should see references to sections that exist in the DB being automatically linked
- References to sections that do not exist in our DB are not linked.

Unfortunately the only way to fully validate this is to visit eRegs and search through the site to find statute references. See if they're linked correctly. The experimental deploy is good for this as it already has most of the statute link converters loaded. Note that paragraph-level linking is not yet supported.

Additionally, review and run the unit test: `docker-compose exec regulations python manage.py test regulations`. This will confirm that the regex is working correctly and replacing the links appropriately.


solution/backend/regulations/admin.py
```diff
@@ -50,10 +60,9 @@
 
     def import_conversions(self, text, url, act):
         conversions = []
-        text = re.sub(r"</?[^>]+>", "", text)  # Strips HTML/XML tags from the response text
-        text = re.sub(r"[—–]", "-", text)  # Replace em dash and en dash with regular dash
-        matches = re.findall(r"[Ss][Ee][Cc].?\s*(\d+[A-Za-z0-9]*(?:-+[A-Za-z0-9]+)?).?\s*\[\s*(\d+)\s*[Uu].?[Ss].?[Cc].?\s*"
-                             r"(\d+[A-Za-z0-9]*(?:-+[A-Za-z0-9]+)?)\s*\]", text)
+        text = markup_regex.sub("", text)  # Strips HTML/XML tags from the response text
```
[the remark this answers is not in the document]
cgodwin1: This is fine as it's needed to extract data from the XML. This is user provided, but only by admins and SMEs and therefore has no effect on the end-user experience. If someone tries to enter a URL that deliberately slows things down, it would only time out for that person.

solution/backend/regulations/views/reader.py
```diff
@@ -56,6 +57,15 @@ def get_context_data(self, **kwargs):
         for location in locations:
             resource_count[location.display_name] = location.num_locations
 
+        conversions = {}
+        for section, usc, act, title in StatuteLinkConverter.objects.values_list("section", "usc", "act", "title"):
+            if act not in conversions:
+                conversions[act] = {}
+            conversions[act][section] = {
```
thwalker6, line 64: Section is not a required field of the model so what will happen if there is no section?

solution/backend/regulations/templatetags/link_statutes.py
```diff
@@ -0,0 +1,64 @@
+from functools import partial
+import re
```
thwalker6, line 2: switch the ordering of from functools and import re.  Imports should be ordered lexicographically within these groups, ignoring case.

thwalker6 (changes requested):

solution/backend/regulations/views/reader.py
cgodwin1: Resolved in pairing discussion

solution/backend/regulations/views/reader.py
```diff
@@ -9,6 +9,7 @@
 from django.http import HttpResponseRedirect
 
 from regcore.models import Part
+from regulations.models import StatuteLinkConverter
 from resources.models import Category, SubCategory, AbstractLocation
```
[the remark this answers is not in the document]
cgodwin1: Resolved in [3ef9a03](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/813/commits/3ef9a037b0ce9c64339bb119918cc6f45e2273de)

thwalker6 (approved):

PR #814, opened 2023-05-04
update code to point to search.gov

peggles2:
Resolves # [EREGCSC-1878](https://jiraent.cms.gov/browse/EREGCSC-1878)

**Description-**
Based on the results of https://jiraent.cms.gov/browse/EREGCSC-1874, search.gov results are good enough for prod.

My understanding is that we implemented https://jiraent.cms.gov/browse/EREGCSC-1783 to make it easy to switch to search.gov, so this is probably more of a task than a story.

It'd be nice if we still had an option to view the old results (&metadata or whatever), but that's not necessary.

- expected change 1

steps to view and verify change. 

1. Visit https://s3pud6e89l.execute-api.us-east-1.amazonaws.com/dev814/search/?q=ex%20parte
2. Notice 50 results on Production this returns 2 because database search can not search on pdf files.


thwalker6 (approved):
LGTM

PR #816, opened 2023-05-08
update code to use site_name as well

peggles2:
Resolves #site_name needs to be passed for prod thats different from dev

**Note**
This ticket is a blocker for [EREGCSC-1878](https://jiraent.cms.gov/browse/EREGCSC-1878) . 

**Description-**
We have created a production search.gov site and plan to use the other as a sandbox. The production site has its own key and sitename. The previous pull request did not take into account that on prod, and val the sitename is different

- expected change 4
-  yml files that set the environment variables in serverless. 
- code to reference the environment variable instead of hard coded site name
- updated Dockerfile template that requires to add the site_name as a environment variable locally

visit [experimental page](https://q3yulgc3qi.execute-api.us-east-1.amazonaws.com/dev816/search/?q=covid&search.gov) notice that the search returns results.
On val do the same test to ensure it still works before going to prod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ git clone https://github.com/Enterprise-CMCS/cmcs-eregulations
 
 ```
 cd solution
-cp DockerFile.template DockerFile
+cp Dockerfile.template Dockerfile
 ```
 - Update the dockerfile with correct environment variables values
 ## Running eRegs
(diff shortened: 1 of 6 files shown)


thwalker6 (approved):

README.md
```diff
@@ -27,7 +27,7 @@ git clone https://github.com/Enterprise-CMCS/cmcs-eregulations
 
 ```
 cd solution
-cp DockerFile.template DockerFile
+cp Dockerfile.template Dockerfile
```
peggles2, line 30: There was a spelling error with Dockerfile.

PR #818, opened 2023-05-09
EREGCSC-1885 Return metadata in search.gov issues

thwalker6:
Resolves # EREGCSC-1885

**Note**
This ticket is a blocker for [EREGCSC-1878](https://jiraent.cms.gov/browse/EREGCSC-1878) . 

**Description-**

In the event that search.gov gives us no results, bad results, or times out, will return our database results

- Returns our results in the event search.gov fails

1. Probably easier to do on your local.  In your docker file put an incorrect search.gov key and make a search for covid.  You should get 51 results.  Put the correct key in.  In the event you are hitting search.gov you should be getting more.


solution/backend/resources/views/resources.py
```diff
@@ -61,29 +62,39 @@ def get_annotated_url(self):
...
-            })
-        self.gov_results = {
-            "total": gov_results["web"]["total"],
-            "results": results,
-        }
+        if "errors" in gov_response or (gov_response["web"]["total"] == 0 and not gov_response["web"]["results"]):
+            self.gov_results = {
+                'total': 0,
+                'results': results
+            }
```
peggles2, line 71: Can we put something in the logs that there was an error? That way should we notice that we are switching to our metadata we can look at the logs and see exactly what the issue was.

solution/backend/resources/views/resources.py
```diff
@@ -61,29 +62,39 @@ def get_annotated_url(self):
...
         rstring = f'https://search.usa.gov/api/v2/search/?affiliate=reg-pilot-cms-test&access_key={key}' \
                   f'&query={urlparse.quote_plus(query)}&limit={self.limit}&offset={offset}'
-        gov_results = json.loads(requests.get(rstring).text)
-        self.format_gov_results(gov_results)
+        try:
+            gov_response = json.loads(requests.get(rstring, timeout=3).text)
+        except ValueError:
+            gov_response = "errors"
+        except ConnectTimeout:
+            gov_response = "errors"
```
peggles2, line 95: can we make the gov_response for ValueError different from ConnectionTimeout. This will allow us to troubleshoot in the logs in the future.

solution/ui/regulations/eregs-vite/src/views/Search.vue
```diff
@@ -355,24 +355,22 @@ export default {
...
-                    page_size: pageSize,
-                    fr_grouping: false,
-                };
-
-                const response = isSearchGov
-                    ? await getSearchGovResources(commonParams)
-                    : await getSupplementalContent(djangoParams);
+                let response = await getSearchGovResources(commonParams)
 
+                if(response.count === 0){
```
PhilR8, line 365: These new lines need to be linted.  (Needs trailing semicolon, space between `if` and the condition, etc)

solution/ui/regulations/eregs-vite/src/views/Search.vue
```diff
@@ -355,24 +355,22 @@ export default {
...
+                    const djangoParams = {
+                        ...commonParams,
+                        partDict: "all",
+                        page_size: pageSize,
+                        fr_grouping: false,
+                    };
+                    response = await getSupplementalContent(djangoParams);
```
PhilR8, line 372: If we are changing this behavior, it would be nice to add some test coverage to make sure it works as expected.  We could specifically intercept these two requests and return some new fixtures that have 0 results for the `getSearchGovResources` interception and has > 0 results for the `getSupplementalContent` interception.

solution/backend/resources/views/resources.py
thwalker6: They are different errors.  The timeout error happens when its over 3 seconds in the request.get.  The value error is if the page returns something funky that isnt json formattable.
json.loads(requests.get(rstring, timeout=3).text)

thwalker6: This response is for the next part where we transform the data.  We are still getting an error regardless and it will be throwing the exception.  We are just handling it in order for the app to not stop.

solution/backend/resources/views/resources.py
thwalker6: We do not currently have logging enabled on our django application and in order to do that properly is out of the scope of this ticket.  if we want to do logging, we should create another ticket.

solution/ui/regulations/eregs-vite/src/views/Search.vue
thwalker6: linted

solution/backend/resources/tests/test_viewset_functions.py
```diff
@@ -57,12 +57,6 @@ def test_format_gov_results(self):
         self.assertEqual(results['results'][0]['url'], 'www.site1url.com')
         self.assertEqual(results['results'][0]['snippet'], '...site1 snippet')
 
-    def test_format_gov_results_failure(self):
-        with open("resources/tests/fixtures/gov_info_failure.json") as f:
```
thwalker6, line 61: since we are handling the failures in a different way this test wont work.

solution/backend/resources/views/resources.py
```diff
@@ -79,10 +82,14 @@ def format_gov_results(self, gov_results):
...
         offset = (page - 1) * self.limit
-
         rstring = f'https://search.usa.gov/api/v2/search/?affiliate=reg-pilot-cms-test&access_key={key}' \
                   f'&query={urlparse.quote_plus(query)}&limit={self.limit}&offset={offset}'
-        gov_results = json.loads(requests.get(rstring).text)
+        try:
+            gov_results = json.loads(requests.get(rstring, timeout=3).text)
+        except ConnectTimeout:
+            raise NotFound("Errors.  Search.gov took to long to return a response.")
+        except ValueError:
```
thwalker6, line 91: When search.gov is having too many request it doesnt return a json response but a web page that was causing a parser error.

solution/ui/regulations/eregs-vite/src/utilities/utils.js
```diff
@@ -48,9 +49,11 @@ function parseError(err) {
         error.requestId = requestId;
         error.root = err;
         error.status = status;
-
         return error;
     } catch {
+        if(err.detail){
```
thwalker6, line 54: we need this to use the error message we provide from django as a value for logic

solution/ui/e2e/cypress/e2e/search.spec.cy.js
```diff
@@ -82,6 +82,83 @@ describe("Search flow", () => {
...
+        ).should("be.visible");
+        cy.get(
+            ".reg-results-content .reg-results-container .result:nth-child(1) .result__link"
+        ).should("be.visible");
+        cy.get(
+            ".reg-results-content .reg-results-container .result:nth-child(1) .result__link a"
+        ).should("have.attr", "href");
+    });
+
+    it("should return no resoures when going on page 3 without enough resources returned", () => {
```
PhilR8, line 123: These tests look good.  There are a couple teeny tiny linting and spelling things, like "resoures" being misspelled on line 123.

solution/ui/e2e/cypress/e2e/search.spec.cy.js
```diff
@@ -82,6 +82,83 @@ describe("Search flow", () => {
...
+            fixture: "search-gov-not-enough.json",
+            statusCode: 400,
+        }).as("resources");
+        // this call is never executed since the conditional in js will not call it.
+        cy.intercept(`**/v3/resources/?&**${SEARCH_TERM2}**page=3**`, {
+            fixture: "meta-data.json"
+        }).as("metadata");
+        cy.viewport("macbook-15");
+        cy.visit(`/search/?q=${SEARCH_TERM2}&page=3`, { timeout: 60000 });
+        cy.wait('@resources');
```
PhilR8, line 134: single quotes should be double quotes

solution/ui/regulations/eregs-vite/src/utilities/utils.js
```diff
@@ -48,9 +49,11 @@ function parseError(err) {
...
+            return err
```
PhilR8, line 55: linting -- needs trailing semicolon

peggles2 (approved):
LGTM

PR #828, opened 2023-05-22
EREGCSC-1620 Vitest implementation

thwalker6:
Resolves # EREGCSC-1620

**Description-**

Adding vitest unit test to our application.  This will allow us to test our individual vue components and functions outside of cypress.

- We will be able to run unit test for components in javascript
- We will be able to run unit test on functions in our application.

Navigate to the solution directory in the project.  
Run  "make test.vitest".
You should see a result of all test passing.
Do not exist the test, the test keep running until you hit control c in the terminal, instead go to the "cmcs-ergulations/solution/ui/regulations/eregs-vite/src/components/jumpto.vue" file.
Add an html element into it. 
The jumpto test will fail for snapshot.
Undo your changes.
In the solution directory run "make test.vitest.coverage"
IN the terminal you should see a report of all the test showing code coverage results.


cgodwin1 (approved):
LGTM

PR #831, opened 2023-05-25
EREGCSC-1917 Section history not loading correctly causing errors

cgodwin1:
Resolves #1917

**Description-**

Section history endpoint sometimes fails if an individual connection takes too long and times out. Because this typically means that the year is invalid (the response would have been 400 if it succeeded) we can simply catch the error.

- Error is caught and the year is ignored instead of the endpoint failing completely
- Unit test is updated to handle this new case
- Cypress test is removed for this endpoint since it relies on an external service and is therefore not reliable

1. Run unit tests for regcore `docker-compose exec regulations python manage.py test regcore`
2. Try the endpoint on the live site
3. Verify cypress tests pass


peggles2 (approved):
LGTM 👍

PR #845, opened 2023-06-08
EREGCSC-1922 Create necessary API endpoint(s) for statute page data

cgodwin1:
Resolves #1922

**Description-**

This PR provides a new API endpoint, `/v3/statutes`, that allows a user to a list of all statutes and relevant data contained in our database. This endpoint is not paginated at this time.

- New endpoint exists.
- New endpoint takes `?act=` parameter to narrow down search, e.g. `/v3/statutes?act=Social Security Act`. By default it will display all.

1. Visit the new endpoint and try no act specified and one or more acts.
2. Run unit tests for regulations
3. Make sure cypress test passes


solution/backend/regulations/tests/test_statute_links.py
```diff
@@ -134,3 +137,55 @@ def test_link_statutes(self):
...
+        },
+        {
+            "section": "7890",
+            "title": 42,
+            "usc": "77777",
+            "act": "Social Security Act",
+            "name": "Something else.",
+            "statute_title": "Z",
+            "source_url": "https://test.gov/test3.xml",
+        },
```
thwalker6, line 170: not a flag but maybe we should as a team discuss using fixture files for larger json objects.  If its like one object within the list its typically fine but when you have multiple it might make sense to just create a fixture file to be called upon by other functions if needed or even our other types of test like cypress or jest.

solution/backend/regulations/urls.py
```diff
@@ -42,5 +43,8 @@
     path('supplemental_content/<id>/', SupplementalContentView.as_view(), name='supplemental_content'),
     path('cache/', CacheView.as_view(), name='cache'),
     path('resources/', ResourcesView.as_view(), name='resources'),
-    path('statutes/', StatuteView.as_view(), name='statues')
+    path('v3/statutes', StatuteLinkConverterViewSet.as_view({
+        "get": "list",
```
thwalker6, line 47: move this to the end of the list for the endpoints and do a simpilar structure to how we do it for regcore. Even though we only have 1 api for regulations currently I think it might make sense to have it separated out a little bit from views.

`path("v3/", include([
        path("statutes/", StatuteLinkConverterViewSet.as_view({
        "get": "list",),
        path("search/", include('regcore.search.urls')),`

thwalker6 (changes requested):
small change, some suggestions/things for team discussion

solution/backend/regulations/tests/test_statute_links.py
cgodwin1: Addressed

thwalker6 (approved):
lgtm

PR #852, opened 2023-06-13
HOTFIX: Disable assertion that relies on changing data

PhilR8:
Resolves failing Cypress assertion in `part.spec`

**Description**
We have an assertion that relies on data returned from an actual API call.  That data has changed -- it has been updated and includes a new value in the 0 index position of an array.  We should not assert on data that may change.

- Comments out assertion that relies on data that may change.

1. Make sure Cypress tests pass.


thwalker6 (approved):
lgtm

PR #855, opened 2023-06-16
EREGCSC-1937 Ruff Implementation

maira-samtek:
Resolves #

**Description-**

- added ruff implementation and for ruff added pre-commit and github action

1. Make sure to setup pre-commit
2. Verify github action runs
3. Try to commit any rule violation on branch


ruff.toml
```diff
@@ -0,0 +1,42 @@
+# Enable pycodestyle (`E`) and Pyflakes (`F`) codes by default.
+select = ["E", "F"]
+ignore = []
+
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+fixable = ["A", "B", "C", "D", "E", "F", "G", "I", "N", "Q", "S", "T", "W", "ANN", "ARG", "BLE", "COM", "DJ", "DTZ", "EM", "ERA", "EXE", "FBT", "ICN", "INP", "ISC", "NPY", "PD", "PGH", "PIE", "PL", "PT", "PTH", "PYI", "RET", "RSE", "RUF", "SIM", "SLF", "TCH", "TID", "TRY", "UP", "YTT"]
```
thwalker6, line 6: remove all fixable rules minus e and f.  Some of this stuff will need to be correccted slowly and sticking with those two for now is best.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
+    rev: v0.0.272
+    hooks:
+      - id: ruff
+        args: [ --fix, --exit-non-zero-on-fix ]
```
thwalker6, line 7: remove args for now while we figure out what we want to do in the pipeline.  having both --fix and --exit-non-zero-on fix can give some backwards results.  --fix will fix it and then --exi-non-zero-on-fix will give an error even if auto correction fixes it.

thwalker6 (changes requested):
Can you add the following to our make file as well.

`ruff:
	docker-compose exec regulations ruff check .;
ruff-fix:
	docker-compose exec regulations ruff --fix .

This will allow us to run the make commands locally.  One for just the check and the other for auto fix.`

thwalker6 (approved):
looks good.  lgtm

PR #857, opened 2023-06-20
EREGCSC-1940 Link statute refs to full paragraph depth

cgodwin1:
Resolves #1940

**Description-**

Now that the house.gov team has fixed the majority of their deeply-nested links, we can more reliably link to paragraphs like 1902(a)(1)(C) and so on.

- Textual links to deeply-nested substructures will link directly there instead of the top-level paragraph.
- Tests are updated.

1. Import statute link converters and verify they are linking to the correct depth
2. Run unit tests for regulations


peggles2 (approved):
LGTM 👍

PR #863, opened 2023-06-23
EREGCSC-1973 Create endpoints to support title/act switcher

cgodwin1:
Resolves #1973

**Description-**

To support the statute table page, we need to be able to filter statutes by not just act, but act + title. (Like Social Security Act Title 2 etc.) We also need to be able to retrieve a list of acts and titles.

- Add additional parameter to `/v3/statutes`, `title`. If `title` is specified, `act` must be or an error will be raised.
- Added endpoint `/v3/acts` to show a list of acts and titles in alphabetical and numerical order.
- Converted all titles from roman numerals to integers. Migration must be run but from now on, any link converter import will do this automatically.

1. As always, run unit tests.
2. If running locally, attempt to run latest migration to verify it works and correctly converts roman numerals to integers.
3. Go to `/v3/statutes?act=Social Security Act` and see that this works as expected. All SSA sections are visible.
4. Go to `/v3/statutes?act=Social Security Act&title=16` and see that this works too. Only SSA sections from title 16 are visible.
5. Go to `/v3/statutes?title=16` and see that this does NOT work. You will see an error saying that an act must be specified.
6. Go to `/v3/acts` and you will see an alphabetically- and numerically-sorted list of acts and titles.


solution/backend/regulations/admin.py
```diff
@@ -98,11 +120,17 @@ def import_conversions(self, text, url, act):
             }
             if section in toc:
                 data["name"] = toc[section]["name"]
-                data["statute_title"] = toc[section]["statute_title"]
+                title = toc[section]["statute_title"]
+                try:
+                    title = int(title)
+                except ValueError:
+                    title = roman_to_int(title)
+                if title:
```
thwalker6, line 128: is if title necessary.  No matter what title should be assigned due to exception catching .

solution/backend/regulations/models.py
```diff
@@ -19,7 +19,7 @@ class StatuteLinkConverter(models.Model):
     usc = models.CharField(max_length=128, verbose_name="USC Section")
     act = models.CharField(max_length=128, verbose_name="Act Name")
```
thwalker6, line 20: should act be required.  We require it in the api call.

thwalker6 (approved):
lgtm

PhilR8 (approved):
`statute_title` and `statute_title_roman` look great!  Thanks for taking the time to make this change.  LGTM!

PR #867, opened 2023-06-27
EREGCSC-1866 Link U.S.C. Citations

cgodwin1:
Resolves #1866

**Description-**

We want to link to citations of the format "42 U.S.C. 1234" and similar. The included section ID regex matches section numbers reliably but needs updated to apply to USC citations as well. Therefore we need a second set of regex to match the new cases, as well as additional code to reliably discriminate between dashes in section ID's (e.g. section 1234-1A) versus "through" dashes (e.g. sections 1000-1005).

- Additional functions are run via regex substitution to replace U.S.C. citations with house.gov links.
- Uses numerical comparison to determine if dashes are part of the ID or a continuation.
- Many edge cases are fixed (different types of dashes, HTML encoded unicode, unusual formatting like 1234aa-1234bb-5 is correctly linked, for example.)
- Tests are updated.

1. Run unit tests.
2. Browse the site to look for examples of U.S.C. linking.


thwalker6 (approved):
looks good.  Lgtm

PR #870, opened 2023-07-07
EREGCSC-2002 Fix statute endpoint ordering

cgodwin1:
Resolves #2002

**Description-**

I forgot to actually do ordering on the statute endpoint. This PR implements it. The tests that I wrote should have caught it, but didn't, so I modified it to insert the test objects in the reverse order so that it is guaranteed to fail if ordering isn't working.

- Add ordering to statute endpoint
- Update test to check for this

1. Check `/v3/statutes` and make sure objects are ordered by act name, then statute title respectively.
2. Run unit tests. If you want proof, go to `backend/regulations/views/statutes.py` and remove `.order_by("act", "statute_title")` on line 41. Then run the test again and it will fail.

### DIFF
--- a/solution/backend/regulations/tests/fixtures/statute_link_api_test.json
+++ b/solution/backend/regulations/tests/fixtures/statute_link_api_test.json
@@ -1,15 +1,5 @@
 [
     {
-        "section": "11111",
-        "title": 45,
-        "usc": "12345",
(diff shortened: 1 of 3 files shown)


thwalker6 (approved):
lgtm

PR #871, opened 2023-07-07
EREGCSC-1988 Link to citations when section symbol is used instead of the word "section"

cgodwin1:
Resolves #1988

**Description-**

Some statute refs use "§" (or worse, unicode HTML `&#xA7;`) instead of the word "section" and its variations. This PR handles that as well as adds more edge cases.

- `§ 1903` will now match.
- `Sec. 1903` will also match.
- Additional edge cases should also be covered, with variations in spacing, plurality, and punctuation. 

1. Check `/42/431/Subpart-A/2023-01-01/#431-15`. "Sec." ref should be linked.
2. Check `/45/75/Subpart-B/2021-11-15/#75-101-e-1-v`. "§" ref should be linked.
3. Run unit tests.


peggles2 (approved):
LGTM 👍

PR #876, opened 2023-07-11
EREGCSC-2009 Settings files broken out by environment

thwalker6:
Resolves # EREGCSC-2009

**Description-**

This is to set up different settings for different environments.  Dev, val and Prod will use deploy.yml, local environment will use local.yml and test will use test_settings.yml.  

The docker-compose file now has a parameter that will pass in an argument for which settings to use.  This is only used on the local to differentiate between a test environment and a local one.  This allows pytest to make sure it uses the correct settings as well as disables django admin toolbar for when cypress runs locally.  Currently a test would fail since the toolbar would block an icon on one of the test.

- Not much on the surface.  Local testing environments will only create one database and prod will have two databases(which is needed for the remove_db lamba).  Debug is also set to off when test are ran.   
- Fixes local running on cypress test from docker.  The django admin toolbar would block an icon when it was ran in docker and this makes it so toolbar doesn't exist for the cypress
- Make test and make test.local are changed to make test.cypress and make test.cypress.local.    We have 3 different test suites that can be run so this just differentiates them.

1. Application deploys successful to dev.
2. The remove of the build will successfully remove the database.  Needs to be verified after merging.
3. Django toolbar appears on the local environment.
4. make test.cypress will successfully run all cypress test in the docker container.
5. make local will set django_settings_module to cmcs_regulations.settings.local.  You can see it in docker-desktop on the environment variables
6.  Running test.cypress or test.pytest will change that to cmcs_regulations.settings.test_settings
 - after the test are done running it switches it back to cmcs_regulations.settings.local


solution/Makefile
thwalker6, line 1: this is just a typo correction

solution/backend/cmcs_regulations/settings/local.py
thwalker6, line 1: local is not being used yet but will be when the paramters all get updated into aws.

local is to be used in docker compose
test_settings will be used for pytest in github.  Will more or less match local since we dont do the dual database thing on our local.
deploy will be used for our environments Dev, Val, Prod, but we can differentiate them later if we want.

solution/backend/createdb.py
```diff
@@ -3,7 +3,7 @@
 
 
 def handler(event, context):
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cmcs_regulations.settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cmcs_regulations.settings.deploy")
```
thwalker6, line 6: default is set to deploy for now.  When app config is set up it will be local

solution/backend/cmcs_regulations/settings/local.py
```diff
@@ -0,0 +1,20 @@
+from .base import * # noqa
+import os
+import socket
+
+
+DEBUG = os.environ.get("DEBUG", False)
+if DEBUG:
```
thwalker6, line 7: this turns on django toolbar

solution/Makefile
```diff
@@ -152,9 +152,9 @@ local.collectstatic: ## Create static assets.
 local.opensearch: ## Run a local opensearch environment
 	docker-compose -f docker-compose.opensearch.yml up -d
 
-test: ## run the cypress e2e suite
-	docker-compose -f docker-compose.yml -f docker-compose.e2e.yml up e2e
-test.local: ## run cypress tests locally without docker
+test.cypress: ## run the cypress e2e suite
```
thwalker6, line 155: we have a lot more test types so test.cypress makes more sense

solution/backend/cmcs_regulations/settings/base.py
```diff
@@ -286,10 +269,3 @@
     "http://*.analytics.google.com",
 ]
 CSP_INCLUDE_NONCE_IN = ["script-src"]
-
-
-if DEBUG:
```
thwalker6, line 291: this is for django toolbar to appear.  Moved to the local.py  file since we dont want this in prod nor do we want it on our test

solution/backend/cmcs_regulations/settings/deploy.py
```diff
@@ -0,0 +1,21 @@
+from .base import * # noqa
```
thwalker6, line 1: noqa tells ruff to ignore this line.  it does not like * imports

PhilR8 (approved):
let's get this merged

PR #881, opened 2023-07-12
EREGCSC-2024 Removal of ruff.yml files

thwalker6:
Resolves # EREGCSC-2024

**Description-**

Currently we have two ruff.toml files and a pyproject file.  Pyproject can hold the settings for multiple python tools(we currently use it for pytest) and it can do ruff as well.  We have two ruff.toml files since the github action is currently using a different one than our local. This change allows us to just juse pyproject.

- No more ruff.toml files.  Only our pyproject will remain.

1. Run "make ruff" on your local, ruff should run succesffully.
2. Our build pipeline ruff step will run  since it will find pyproject.toml


thwalker6:
i closed the PR to test the remove step was working correctly.  No other reason

PhilR8 (approved):

PR #897, opened 2023-07-20
EREGCSC-2056 -- Remove browser alert on error

PhilR8:
Resolves an issue that has been a headache for a while

**Description**

When an error is encountered on the front end of the app, a browser `alert` is fired with the error information displayed in the `alert` as text.

This `alert` has become the topic of discussion recently; it often displays nonsensical markup that means nothing to anyone, and recently it has been interfering with our end to end tests, causing them to hang instead of fail (we think).

As a result, the `alert` dialogs are being completely removed.  Error information will be displayed in the Javascript console in the browser dev tools. 

- Removes single line in the `parseError` utility function that triggers the browser `alert` on error

1. Force an error and make sure the browser alert is not shown


thwalker6 (approved):
lgtm

PR #898, opened 2023-07-21
Eregcsc 2010 permission groups

thwalker6:
Resolves # EREGCSC-2010

**Description-**

New user groups had to be created for different types of users.  Currently we only will have three.  We need the internal resource type to get the groups finished out.  This will allow us to use seed data from prod to populate the tables to our lower environments   Seed data for site configurations statute links have also been updated

- User groups are added to the see data
- User groups have been created in production
- statute links configs have been added to seed data

 On your local environment make sure all migrations are up to date and run make local.seed.
Log into admin.
There should be three user groups.
Go to production and look at the user groups for admin permissions.
There should be the three groups there.


peggles2 (changes requested):
Looks good! Just wondering if we can add a test to ensure this is never ran on prod.

peggles2 (approved):
Agreed to add tests to EREGCSC_2062

PR #907, opened 2023-07-25
[HOTFIX] Update Statutes end to end test so that v3/statutes intercept works

PhilR8:
Resolves deployment issue where intercept for Statutes e2e test was not written correctly and was not working.

**Description**

Currently our `dev` and `val` environments do not have statute data -- the `v3/statutes` endpoint returns an empty array.  This is being worked on in another story.

However, our Cypress end to end tests generally intercept API calls and use fixtures/mock data instead of actual data.  So the empty array response from the `v3/statutes` endpoint should not have been an issue.

- adds a double asterisk glob matcher in the `v3/statutes` intercept URL so that the fixture of mock data is returned instead of real data.

1. Run e2e tests locally.  Watch the Statutes test and step through it when it is done, making sure the rendered table contains fixture data and not actual API response data.

### DIFF
--- a/solution/ui/e2e/cypress/e2e/statutes.spec.cy.js
+++ b/solution/ui/e2e/cypress/e2e/statutes.spec.cy.js
@@ -4,7 +4,7 @@ describe("Statute Table", () => {
             req.headers["x-automated-test"] = Cypress.env("DEPLOYING");
         });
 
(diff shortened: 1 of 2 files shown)


peggles2 (approved):
LGTM

PR #913, opened 2023-07-27
EREGCSC-1964-B Modify statute override to work for "and"-type references

cgodwin1:
Resolves #1964

**Important note:**

Due to the fact that the data structure is entirely different in this PR, existing link overrides will NOT be converted when this PR is deployed to prod. They will be deleted and have to be re-entered. There's really no way around this because existing overrides don't differentiate between types of references like they do now.

**Description-**

The previous method of link override worked well for single citations (Section 1902 of the Act) but wasn't sufficient if you had two sections in a ref and only wanted to prevent one of them from linking (e.g. "Sections 1902 and 1903 of the Act", only stop linking 1902.)

- `do_not_link` is replaced with 2 fields, `statute_ref_exceptions` and `usc_ref_exceptions` (to prevent possible collisions).
    - Note the use of JSONField with a custom schema to store this data. Due to the limited number of link overrides foreseen, this data structure will be faster than a many-to-one with another model, as it prevents the need for additional database queries. This _may not_ be appropriate if overrides are predicted to grow into the hundreds or thousands, but that is very unlikely at this time.
- For each entry, you must specify either an act (defaults to SSA) or a title, then the individual section + paragraph citation. This is the major change as you no longer must specify the entire reference string, thus allowing individual citations to be blocked.
- Tests are updated to make sure this new way is working.

1. As always, run unit tests. 
2. In the admin panel, go to `Statute Link Configuration`, and observe the two new fields.
3. Go to a reader view page in eRegs that contains statute refs, and add a statute ref exception. Refresh the page and see that it isn't linked.
4. Do the same for a U.S.C. ref.
5. Find refs of both types that are "and"-type (e.g. "Sections 1902 and 1903" or "42 U.S.C. 1 and 2"). Do the same as before, but only block _one_ of the citations. (I.e. if the ref is "42 U.S.C. 1 and 2", enter a title of "42" and a section of "1" only.)
6. Refresh the page and see that the other is still linked.


peggles2 (approved):

peggles2:
Approved pending passing cypress test.

PR #914, opened 2023-07-28
EREGCSC-2041 -- Set up url and configs for OIDC

thwalker6:
Resolves # EREGCSC-2041

**Description-**

In order for us to set up oAuth we need a call back url.  This is getting us set up.

- A call back url is set up for our app at {{host}}//oidc/callback/

1. Go to https://gqhz3je798.execute-api.us-east-1.amazonaws.com/dev914/oidc/callback  .  You will see a forbidden message.  We do not have a client secret yet so it wont work.
2. Go to the admin panel and log in.  You should be able to log in still with the old method of django.  You can do this either through your local or the dev site.  We expanded the django auth backends to include a new method as well as the old method.  Without explicitly adding the old method in settings it will not work.

### DIFF
--- a/.github/workflows/git-secrets.yaml
+++ b/.github/workflows/git-secrets.yaml
@@ -4,7 +4,7 @@ jobs:
   gitleaks-scan:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Run gitlakes docker	    
       uses: docker://zricethezav/gitleaks
       with:
--- a/solution/backend/cmcs_regulations/settings/base.py
+++ b/solution/backend/cmcs_regulations/settings/base.py
@@ -72,7 +72,6 @@ OIDC_RP_IDP_SIGN_KEY = os.environ.get("OIDC_RP_IDP_SIGN_KEY", None)
 AUTHENTICATION_BACKENDS = (
     'django.contrib.auth.backends.ModelBackend',
     'mozilla_django_oidc.auth.OIDCAuthenticationBackend',
-    # ...
 )
 OIDC_RP_CLIENT_ID = os.environ.get("OIDC_RP_CLIENT_ID", None)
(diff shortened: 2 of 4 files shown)


.github/workflows/git-secrets.yaml
```diff
@@ -4,7 +4,7 @@ jobs:
   gitleaks-scan:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
thwalker6, line 7: Gitleaks had a warning for using v2 being old so I just changed it to v3.  Documentation for gitleaks github action also has it using v3 currently.

solution/backend/cmcs_regulations/settings/base.py
```diff
@@ -66,6 +67,17 @@
...
+# Add 'mozilla_django_oidc' authentication backend
+AUTHENTICATION_BACKENDS = (
+    'django.contrib.auth.backends.ModelBackend',
+    'mozilla_django_oidc.auth.OIDCAuthenticationBackend',
+)
+OIDC_RP_CLIENT_ID = os.environ.get("OIDC_RP_CLIENT_ID", None)
+OIDC_RP_CLIENT_SECRET = os.environ.get("OIDC_RP_CLIENT_SECRET", None)
+OIDC_OP_AUTHORIZATION_ENDPOINT = os.environ.get("OIDC_OP_AUTHORIZATION_ENDPOINT", None)
+OIDC_OP_TOKEN_ENDPOINT = os.environ.get("OIDC_OP_TOKEN_ENDPOINT", None)
+OIDC_OP_USER_ENDPOINT = os.environ.get("OIDC_OP_USER_ENDPOINT", None)
```
thwalker6, line 80: putting all of these values into environment variables until we figure out what is a secret and what isnt.  Since we dont actually have the token stuff yet we have to set them to none.  no impact on the application

PR #916, opened 2023-07-31
Eregcsc 1966 box api proof of concept

peggles2:
Resolves #

**Description-**

- expected change 1


solution/Dockerfile.template
```diff
@@ -12,6 +12,9 @@ RUN pip install --no-cache-dir --upgrade pip setuptools \
 
 ENV SEARCHGOV_KEY=""
 ENV SEARCHGOV_SITE_NAME=""
+ENV BOX_CLIENT_ID=''
+ENV BOX_CLIENT_SECRET=''
+ENV BOX_REDIRECT_URL=''
 
 CMD ["python", "manage.py", "runserver", "0.0.0.0:8000"]
```
peggles2, line 20: If we did use box, we will need to update our Dockerfile to have these values. They are located in https://app.box.com/developers/console/app/2059452/configuration . 
If we use JWT token we would also need to add the location of the RSA file, JWT key and passphrase discussed here[ ](https://github.com/box/box-python-sdk#server-to-server-auth-with-jwt)

solution/backend/cmcs_regulations/settings/base.py
```diff
@@ -79,6 +79,11 @@
 OIDC_OP_TOKEN_ENDPOINT = os.environ.get("OIDC_OP_TOKEN_ENDPOINT", None)
 OIDC_OP_USER_ENDPOINT = os.environ.get("OIDC_OP_USER_ENDPOINT", None)
 
+# BOX API SETTINGS
+BOX_CLIENT_ID = os.environ.get("BOX_CLIENT_ID")
+BOX_CLIENT_SECRET = os.environ.get("BOX_CLIENT_SECRET")
+BOX_REDIRECT_URL = os.environ.get("BOX_REDIRECT_URL")
+
```
peggles2, line 86: Note that this will look for the keys in the environment so we would need to update AWS with Paramstore value for BOX_CLIENT_ID, BOX_CLIENT_SECRET and BOX_REDIRECT_URL in addition if we choose JWT, we would also need to add the location of the RSA file, JWT key and passphrase

PR #919, opened 2023-08-02
EREGCSC-2062 -- Added populate content to deploy for val and dev

thwalker6:
Resolves # EREGCSC-2062

Populate dev and val with populate content on build

**Description-**

Currently dev and val seed data are only populated when you log into aws and run the lambda.  This will execute the lambda on those environments on deploy.  The populate content lamda only exist in dev and val.  it does not exist in prod anymore so it has no chance of populting.  On deploy it will check the envrionment and if it is not prod it will run the lambda.

- Populatecontent lambda executes in dev and val.

1. When the deploy happens in dev the step for running in dev will run.
2. When the deploy happens in val the lamda will run.
3. The step will be skipped in prod(the lambda does not exist in prod which has already been verified).


cgodwin1 (approved):
LGTM

PR #923, opened 2023-08-04
EREGCSC-2029 - Implement I sort ruff

thwalker6:
Resolves # EREGCSC-2029

**Description-**

This is to implement ISORT in ruff.  It forces correct ordering of of our imports for python.  Isort allows auto reordering for it to make it easy for us to correct them.

- The python imports are set to order in a particular way.
- It will automatically do them if we ask it to "make ruff-fix"

1. Application runs without breaking.  All test pass and deployments.


peggles2 (approved):
LGTM 👍

PR #931, opened 2023-08-15
EREGCSC-2111 -- Determine ShowIfEmpty status for sidebar items on the front end

PhilR8:
Resolves[ EREGCSC-2111 ](https://jiraent.cms.gov/browse/EREGCSC-2111)

**Description**

The right sidebar of the reader view has resources categories that can have or not have children.  We should have the ability to show or hide these categories even if they are empty.  Unfortunately, this was being incorrectly filtered in the Django view and wasn't working properly.  

This PR will move the logic to determine showing or hiding an empty category to the front end.

- No longer passes `sub_categories` into the template from the Django view.  This was refactored long ago to just pass in an empty object; we don't need to do that, so it has been removed.
- Removes code and logic everywhere (Vue component, utility function, unit tests) that assumed an empty `sub_categories` object was passed in to the Django template.
- Passes in all categories from the Django view into the Django template and the Vue component -- no longer filters on `show_if_empty=True` in the Django view.
- Determines on the front end if an empty category should be hidden or shown in the Vue component.
- Updates tests to make sure `showIfEmpty` is properly working in the Vue component.

1. Visit [experimental deployment](https://cx4wcijizl.execute-api.us-east-1.amazonaws.com/dev931/) and go to section 440.130
2. Visit admin page for experimental deployment and go to Resources and Regulations > Categories
3. Toggle Show if Empty for a bunch of categories and make sure everything behaves as expected
4. Specifically, 440.130 should still show "Reports and Evaluations" even when "Show if Empty" is set to `false`
   - compare against the [`dev` site](https://cqeki073yh.execute-api.us-east-1.amazonaws.com/dev/42/440/Subpart-A/2020-12-16/#440-130), which hides "Reports and Evaluations" when "Show if Empty" is `false`

### DIFF
--- a/solution/backend/regulations/views/reader.py
+++ b/solution/backend/regulations/views/reader.py
@@ -22,7 +22,6 @@ from regulations.views.utils import find_subpart
 from resources.models import (
     AbstractLocation,
     Category,
-    SubCategory,
 )
 
 
@@ -54,7 +53,6 @@ class ReaderView(CitationContextMixin, TemplateView):
         tree = self.get_content(context, document, toc)
         node_list = self.get_supp_content_params(context, [tree])
         categories = list(Category.objects.filter(show_if_empty=True).contains_fr_docs().order_by('order').values())
-        sub_categories = list(SubCategory.objects.filter(show_if_empty=True).contains_fr_docs().order_by('order').values())
 
         locations = AbstractLocation.objects.filter(part=reg_part).select_subclasses().annotate(
             num_locations=Count(
@@ -93,7 +91,6 @@ class ReaderView(CitationContextMixin, TemplateView):
             'node_list':    node_list,
             'view_type':    self.get_view_type(),
             'categories':   categories,
-            'sub_categories': sub_categories,
             'resource_count': resource_count,
             'link_conversions': conversions,
             'link_config': statute_link_config,
--- a/solution/backend/regulations/views/reader.py
+++ b/solution/backend/regulations/views/reader.py
@@ -52,7 +52,7 @@ class ReaderView(CitationContextMixin, TemplateView):
         part_label = toc['label_description']
         tree = self.get_content(context, document, toc)
         node_list = self.get_supp_content_params(context, [tree])
-        categories = list(Category.objects.filter(show_if_empty=True).contains_fr_docs().order_by('order').values())
+        categories = list(Category.objects.contains_fr_docs().order_by('order').values())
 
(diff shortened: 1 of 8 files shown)


solution/ui/regulations/test/unit-test/SupplementalContent.test.js
```diff
@@ -34,13 +32,35 @@ describe("Supplemental Content", () => {
...
+                subparts: ["A"],
+            },
+        });
+
+        await flushPromises();
+
+        const subEmptyButVisible = screen.queryByText("Empty but Visible");
+        expect(subEmptyButVisible).toBeTruthy();
+
+        const subEmptyAndHidden = screen.queryByText("Empty and Hidden");
```
PhilR8, line 60: `queryBy...` returns `null` instead of an error, which allows us to assert against a query that doesn't find anything.  See https://testing-library.com/docs/queries/about/

solution/ui/regulations/test/unit-test/SupplementalContent.test.js
```diff
@@ -50,25 +70,29 @@ describe("Supplemental Content", () => {
...
         const subG = await screen.getByLabelText(
             "expand Subregulatory Guidance"
         );
         expect(subG).toBeTruthy();
-        await fireEvent.click(subG);
 
+        await fireEvent.click(subG);
         expect(subG.textContent).toStrictEqual("Subregulatory Guidance ");
+
         const stateMedBtn = await screen.getByText(
```
PhilR8, line 83: the rest of the changes in this file are formatting changes

solution/backend/regulations/views/reader.py
```diff
@@ -53,8 +52,7 @@ def get_context_data(self, **kwargs):
         part_label = toc['label_description']
         tree = self.get_content(context, document, toc)
         node_list = self.get_supp_content_params(context, [tree])
-        categories = list(Category.objects.filter(show_if_empty=True).contains_fr_docs().order_by('order').values())
-        sub_categories = list(SubCategory.objects.filter(show_if_empty=True).contains_fr_docs().order_by('order').values())
+        categories = list(Category.objects.contains_fr_docs().order_by('order').values())
```
PhilR8, line 55: we're just sending all categories into the template for use on the front end.  That's now where we will decide if empty categories will be shown or hidden

peggles2 (approved):
LGTM 👍

PR #933, opened 2023-08-16
EREGCSC-2093 Link CFR-Type References

cgodwin1:
Resolves #2093

**Description-**

This PR implements linking within the reg text for references of the type "42 CFR part 123" and similar. See the Google Doc for info. This also builds the framework for which all other reg ref linking will be built.

- References of the form `A CFR (part(s)) B(, C, and/or D...)` are linked to an intermediate redirect page
- The intermediate redirect page (`/reg_redirect/`):
    1. Queries eRegs to determine if the title/part combo exists in our database.
    2. If it does, redirects there.
        a. If the part exists but the section doesn't, redirect to the top of the part view.
        b. If the section exists but the paragraph doesn't, link to the top of the section.
    3. If it does not exist, redirect to eCFR in hopes that it will be available there.
- A configuration page on the admin panel exists and includes link exceptions like with statute refs.

1. Run unit tests
2. Make sure linking is enabled on the admin panel under "Regulation Link Configuration"
3. Go to any page containing CFR-type refs and verify they are working correctly
4. Add an exception on the admin panel and verify that the exception is not linked
5. Disable linking and make sure it stops working for all CFR-type refs (but *not* for other types of refs!)


PhilR8 (approved):
Everything looks good, all tests pass.  LGTM!

PR #938, opened 2023-08-21
EREGCSC-1848 -- better JS/CSS cache busting, take two

PhilR8:
Resolves [EREGCSC-1848](https://jiraent.cms.gov/browse/EREGCSC-1848)

**Description:**

We want to be able to push CSS and JS updates to the site at any time of day and have users transparently download the newest, latest CSS/JS files even if they've already visited the site that day. This should not require any action from the user. They should just be served the newest, freshest JS/CSS files by the browser as soon as they have been deployed.

This means the browser must be told that something has changed and that it should fetch a new copy of the CSS/JS files from the server instead of using a cached local copy.

We are going to grab an [artifact from the Github Action workflows called `run_id`](https://docs.github.com/en/actions/learn-github-actions/contexts#about-contexts), which is unique to each workflow:
   - `deploy-experimental.yml` for Pull Request deployments
   - `deploy.yml` for `dev`/`val`/`prod` deployments

As a result, every time the `deploy.yml` workflow is run, a new unique `run_id` will be created and used to force the browser to reload JS and CSS files.

- Gets `run_id`and passes it into Django deployment when Github Action is run
- uses `run_id` as [parameter to append to end of JS and CSS link/script tags in Django templates to "bust" those cached files](https://www.wkoorts.com/2009/07/12/css-auto-reload-with-django-templates/) when this parameter changes and force browser to reload the files from the server

1. Go to [Experimental Deployment](https://3rs9ts08we.execute-api.us-east-1.amazonaws.com/dev938), open the Network Tab in your browser's DevTools, and look at the name of the JS/CSS files being requested by the browser.  The JS/CSS files should end with a parameter that is a string of 10+ numbers. The filenames should look something like this:
   - `main.css?5939141493`
   - `eregs-main.iife.js?5939141493` (Django template used for most content) or `index.js?5939141493` (Vue Single Page App used for most content) depending on which page you're on 
2. Visit this PR's [Experimental Deployment Action's `deploy-django` step summary page on Github](https://github.com/Enterprise-CMCS/cmcs-eregulations/actions/runs/5939141493/job/16105052416)
   - The URL should have the same `runs` number that you see appended to the JS/CSS files
3. Expand the `deploy regulations site server` step and then expand the `Run pushd solution/backend` command
4. Scroll down to `env` (around line 13) and make sure the `RUN_ID` env variable matches all the numbers mentioned in the steps above
5. Visit a few other deployment action summary pages, experimental or otherwise, and notice that the `runs` URL parameter for each PR is unique

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -119,7 +119,7 @@ jobs:
           pushd solution/backend
           npm install serverless -g
           npm install
-          RUN_ATTEMPT=${{ github.run_attempt }} serverless deploy --config ./serverless-experimental.yml --stage dev${PR} | tee output.log
+          serverless deploy --config ./serverless-experimental.yml --stage dev${PR} | tee output.log
           serverless invoke --config ./serverless-experimental.yml --function create_database --stage dev${PR}
           serverless invoke --config ./serverless-experimental.yml --function reg_core_migrate --stage dev${PR}
           serverless invoke --config ./serverless-experimental.yml --function create_su --stage dev${PR}
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -119,7 +119,7 @@ jobs:
           pushd solution/backend
           npm install serverless -g
           npm install
-          serverless deploy --config ./serverless-experimental.yml --stage dev${PR} --param deploy_num=${{ gitub.run_id }} | tee output.log
+          serverless deploy --config ./serverless-experimental.yml --stage dev${PR} | tee output.log
           serverless invoke --config ./serverless-experimental.yml --function create_database --stage dev${PR}
           serverless invoke --config ./serverless-experimental.yml --function reg_core_migrate --stage dev${PR}
           serverless invoke --config ./serverless-experimental.yml --function create_su --stage dev${PR}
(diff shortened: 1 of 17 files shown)


.github/workflows/deploy-experimental.yml
```diff
@@ -115,6 +115,7 @@ jobs:
         if: success() && steps.findPr.outputs.number
         env:
           PR: ${{ steps.findPr.outputs.pr }}
+          RUN_ID: ${{ github.run_id }}
```
PhilR8, line 118: see documentation on `run_id` here: https://docs.github.com/en/actions/learn-github-actions/contexts

solution/backend/regulations/templates/regulations/spa_base.html
```diff
@@ -0,0 +1,7 @@
+{% extends "regulations/base.html" %}
+{% load static %}
+
+{% block post_footer %}
+    <script src="https://cdn.jsdelivr.net/npm/vue@2.7.14"></script>
+    <script src="{% static '/vite/index.js' %}?{{ DEPLOY_NUMBER }}"></script>
+{% endblock %}
```
PhilR8, line 7: spa-specific assets in one place

cgodwin1 (approved):
LGTM!

PR #941, opened 2023-08-23
EREGCSC-2148 - Internal document set up

thwalker6:
Resolves # ERECSC-2148

**Description-**

Setting up API calls and s3 buckets.

- Api calls are created for getting a list of subjects, categories and files

1. Only works correctly on local.  We know it works correctly through the demo.


solution/ui/regulations/utilities/api.js
```diff
@@ -55,7 +55,12 @@ function configure(obj) {
     config = { ...config, ...obj };
 }
 
-function fetchJson(url, options = {}, retryCount = 0) {
+function fetchJson({
+    url,
+    options = {},
+    retryCount = 0,
+    cacheResponse = true,
```
PhilR8, line 62: adding `cacheResponse` flag to base API fetch method.  If passed, API response is not set in localStorage and it is also will not get from localStorage.

solution/backend/file_manager/admin.py
```diff
@@ -0,0 +1,71 @@
...
+                             Bucket=settings.AWS_STORAGE_BUCKET_NAME,
+                             Key="uploaded_files/" + file._name,
+                             ContentType=file.content_type)
+
+    def del_file(self, obj):
+        s3_client = establish_client()
+        try:
+            s3_client.delete_object(Bucket=settings.AWS_STORAGE_BUCKET_NAME, Key=f"{obj.file.name}")
+        except Exception:
+            print("Unable to delete")
```
cgodwin1, line 55: A todo for future, print a useful message for the user

solution/ui/regulations/utilities/api.js
```diff
@@ -100,7 +105,11 @@ function fetchJson(url, options = {}, retryCount = 0) {
...
 
     return Promise.resolve()
-        .then(() => localforage.getItem(url.replace(apiPath, merged.method)))
+        .then(
+            () =>
+                cacheResponse &&
+                localforage.getItem(url.replace(apiPath, merged.method))
+        )
         .then((value) => {
             if (value && Date.now() < value.expiration_date) {
```
PhilR8, line 114: if `cacheResponse` is `falsy`, `value` will be `undefined` and it will short circuit this `if` condition immediately and drop us into the `else` block, which will fetch from the server.

solution/backend/file_manager/admin.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def delete_model(self, request, obj):
+        try:
+            self.del_file(obj)
+            obj.delete()
+            print('File Deleted')
+        except Exception:
+            print('Could not delete from server.')
```
cgodwin1, line 63: Same as above, for the future

solution/ui/regulations/utilities/api.js
PhilR8: Also of note: we are now passing in a single object as the only function parameter.  No more worrying about parameter order when invoking this function.

solution/ui/regulations/utilities/api.js
```diff
@@ -123,7 +132,9 @@ function fetchJson(url, options = {}, retryCount = 0) {
                         )
                     )
                     .then(() => delay(backoff))
-                    .then(() => fetchJson(url, options, retryCount + 1));
+                    .then(() =>
+                        fetchJson({ url, options, retryCount: retryCount + 1 })
```
PhilR8, line 136: Note that we are passing in a single object as an argument now

cgodwin1 (approved):
LGTM. Future work includes displaying error messages to the user instead of debug print statements. Not important for demo purposes.

PR #953, opened 2023-09-05
EREGCSC-2201 - Generate presigned url for uploads

thwalker6:
Resolves # EREGCSC-2201

**Description-**

In order to work around  file size restrictions and to get around the lambda, presigned URLs will be used for uploading to the s3 bucket.

- Files will no longer be uploaded through django but instead the lambda will return a presigned url to upload the file directly from the user.
- The model will no longer use a models.filefield for file name but instead use a character field.

1. Upload a file through the django admin as currently done. and click save
2. Download the file after uploading it.

3.  Go to swagger and find the endpoint for v3/file_manager/file_name/{file_name}
4. For the file name put in "file.doc"
5. It will return  a restponse with a url for aws, and fields needed to build the key.


peggles2 (approved):
Lgtm

PR #954, opened 2023-09-05
EREGCSC-2011 - Setup EUA mapping

peggles2:
Resolves #[EREGCSC-2011](https://jiraent.cms.gov/browse/EREGCSC-2011)

**Description-**
This pull request is the initial settup of the eua mapping

- Dockerfile.template - includes the additional eua values we should add to our Dockerfile to run this locally. All values can be found on AWS dev parameter store. 
- base.py - cleaned it up. Removed the commented out values and refactored some code. 
- deploy.py cleaned up redundant code.
- added the extra needed keys, updated the REDIRECT_LOGIN AND REDIRECT_LOGOUT so that the redirect url reflects when its on experimental and the subdomain is devxxxx where xxxx is some pull request number. 
- Featureflag condition wrapped around login to EUA, so that it does not show up on prod till we are ready. 
- admin.py updated with a OidcAdminAuthenticationBackend which handles the callback when the user is authenticated.
- context_processor.py updated to included EUA_Featureflag so that the template can use this value to determine if it should display the eua login or not.
- serverless-expermiental.yml updated with BASE_URL that it was missing and a new one called STAGE_ENV which tells us the subdomain name. In addition to that, added all the OIDC okta keys necessary for eua
- serverless.yml same values that were configured for serverless-experimental.yml.
- docker.compose.yml updated to deploy with euasettings.yml when run locally. 
- requirements.txt - added an older version of cryptology due to an issue with the new one working on AWS Lambda. See this ticket [EREGCSC-2218](https://jiraent.cms.gov/browse/EREGCSC-2218)

1. visit [in action](https://c3suvrrjg6.execute-api.us-east-1.amazonaws.com/dev954)/admin
2. sign in
3. expect to be signed in and redirected to the admin login page with a message like this:
You are authenticated as , but are not authorized to access this page. Would you like to login to a different account?
4. You can login to the admin page without eua and give the account that has signed in access.  (note the username and automatic access based on access thats returned from eua, of that will be updated in the next ticket) 
5. sign out
6. sign back in with eua
7. Notice you have been signed into admin panel with eua 

**Note**  
currently the DJANGO_SETTINGS_MODULE has to be set to  cmcs_regulations.settings.euasettings manually in the lambda until its ready to be deployed. To do this visit the lambda 
https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/cmcs-eregs-site-dev954-reg_site?tab=configure
click on the Configuration tab.
click edit
Change DJANGO_SETTINGS_MODULE TO cmcs_regulations.settings.euasettings


.github/workflows/deploy-experimental.yml
```diff
@@ -44,6 +44,12 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r ./solution/static-assets/requirements.txt
+          pip install
+          --platform **manylinux2014_aarch64 **
+          --target=./python/lib/python3.9/site-packages
+          --implementation cp
+          --python 3.9
+          --only-binary=:all: --upgrade
```
peggles2, line 52: https://repost.aws/knowledge-center/lambda-python-package-compatible

solution/backend/cmcs_regulations/settings/euasettings.py
```diff
@@ -0,0 +1,33 @@
...
+OIDC_RP_CLIENT_SECRET = os.environ.get("OIDC_RP_CLIENT_SECRET", None)
+OIDC_OP_AUTHORIZATION_ENDPOINT = os.environ.get("OIDC_OP_AUTHORIZATION_ENDPOINT", None)
+OIDC_OP_TOKEN_ENDPOINT = os.environ.get("OIDC_OP_TOKEN_ENDPOINT", None)
+OIDC_OP_USER_ENDPOINT = os.environ.get("OIDC_OP_USER_ENDPOINT", None)
+OIDC_OP_JWKS_ENDPOINT = os.environ.get("OIDC_OP_JWKS_ENDPOINT", None)
+OIDC_RP_SIGN_ALGO = 'RS256'
+LOGIN_REDIRECT_URL = BASE_URL + '/admin/'
+LOGOUT_REDIRECT_URL = BASE_URL + '/'
+
+EUA_FEATUREFLAG = os.environ.get('EUA_FEATUREFLAG', 'false')
```
thwalker6, line 29: Use a boolean instead of a string representation.

solution/backend/cmcs_regulations/templates/admin/login.html
```diff
@@ -19,9 +19,11 @@
         </style>
     {% endblock %}
    {{ block.super }}
-   <div class="okta_login_container">
-       Or Login With&nbsp; <a href="https://www.google.com/">EUA</a>
-   </div>
+   {% if EUA_FEATUREFLAG == 'true' %}
```
thwalker6, line 22: when its changed to a boolean we can just do { % if EUA_FEATUREFLAG %}

thwalker6 (changes requested):
Make statments truthy since this is how we currently do it, also the build failed on go

solution/backend/cmcs_regulations/templates/admin/login.html
peggles2: AWS parameter store only returns values as string. To simplify its just best to keep it consistent.

thwalker6 (approved):
lgtm

PR #963, opened 2023-09-14
EREGCSC-2190 Update File Upload Endpoint

thwalker6:
Resolves # EREGCSC-2190

**Description-**

Updating our endpoints for file manager to be more "rest compliant".  The original endpoints didnt follow naming conventions that were liked as a team so we are just updating them to be better.

- Just the naming of the endpoints all functionality remains the same.

Log into the admin panel.  Upload a file, download a file.  

go to https://ny2apwavx6.execute-api.us-east-1.amazonaws.com/dev963/policy-repository/.
There should be a file there already and if you upload more they should be listed as well.

Due to a known issue do not use a file that has a comma in the title for an uploaded document.  Is fixed in a separate PR.


cgodwin1 (approved):
LGTM!

PR #967, opened 2023-09-19
EREGCSC-2217 - Buckets use SSL

thwalker6:
Resolves # EREGCSC-2217

**Description-**

Security scan issue on our S3 bucket due to it not requiring SSL.   This merely adds it.

- SSL is required for accessing the bucket

1. Bucket upload works successfully as well as retrieval.  Upload and download a doucment.  Full verification wont be done until it hits prod.
2.  Bucket was first created with the ticket without SSL.  Item was uploaded.  Then SSL was added.  Item was still accessible.  This test was done to make sure it wouldnt rewrite the existing bucket.

### DIFF
--- a/solution/backend/file_manager/admin.py
+++ b/solution/backend/file_manager/admin.py
@@ -84,7 +84,7 @@ class UploadedFileAdmin(BaseAdmin):
     def download_file(self, obj):
         if obj.id:
             link = reverse("file-download", kwargs={"file_id": obj.uid})
-            html = '<input type="button" onclick="location.href=\'{}\'" value="Doswnload File" />'.format(link)
+            html = '<input type="button" onclick="location.href=\'{}\'" value="Download File" />'.format(link)
         else:
             return "N/A"
         return format_html(html)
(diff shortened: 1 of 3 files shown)


peggles2 (approved):

peggles2 (changes requested):
This is good to have but it does not fix the issue

SecurityHub Finding - S3.8 S3 Block Public Access setting should be enabled at the bucket-level

Adding this block to the bucket should solve it for you (but this is what could accidentally block CloudFront if it was connecting publicly)
PublicAccessBlockConfiguration:
  BlockPublicAcls: true
  BlockPublicPolicy: true
  IgnorePublicAcls: true
  RestrictPublicBuckets: true
  docs here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-publicaccessblockconfiguration.html

thwalker6:
> This is good to have but it does not fix the issue
> 
> SecurityHub Finding - S3.8 S3 Block Public Access setting should be enabled at the bucket-level
> 
> Adding this block to the bucket should solve it for you (but this is what could accidentally block CloudFront if it was connecting publicly) PublicAccessBlockConfiguration: BlockPublicAcls: true BlockPublicPolicy: true IgnorePublicAcls: true RestrictPublicBuckets: true docs here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-publicaccessblockconfiguration.html

This is not that ticket.  This has  nothing to do with it

peggles2 (approved):

PR #970, opened 2023-09-22
EREGCSC-2214 -- Subject Filter for Policy Repository

PhilR8:
Resolves [EREGCSC-2214](https://jiraent.cms.gov/browse/EREGCSC-2214)

**Description:**

Add ability to filter Policy Documents by subject

- Adds Subject Filter to the Policy Respository page
- Adds Vue Router so URL query params control page state
- Adds basic styles from [this Figma comp](https://www.figma.com/file/KiWw729Nct6tY9gzmvEkVO/Subject-filter-only?node-id=108%3A1909&mode=dev)

1. Visit [policy repository page at experimental deployment](https://y65lx1o46e.execute-api.us-east-1.amazonaws.com/dev970/policy-repository) or build locally
2.  Use the subject selector.  Try using the cursor and the keyboard
3. Add and remove chips
4. Use the Back and Forward buttons
5. Select a bunch of chips and the copy and paste the URL into a new browser tab.  The same subjects should be selected on page load


solution/backend/cmcs_regulations/settings/base.py
```diff
@@ -212,7 +213,7 @@
     "https://docs.google.com/forms/d/e/1FAIpQLSdcG9mfTz6Kebdni8YSacl27rIwpGy2a7GsMGO0kb_T7FSNxg/viewform?embedded=true"
 )
 
-DEPLOY_NUMBER = os.environ.get("DEPLOY_NUMBER", "0000")
+DEPLOY_NUMBER = os.environ.get("DEPLOY_NUMBER", datetime.now())
```
PhilR8, line 216: changing the default to `datetime.now` so when working locally, the assets are fetched every refresh

solution/ui/regulations/.eslintrc.json
```diff
@@ -15,6 +15,7 @@
     },
     "plugins": ["vue"],
     "rules": {
+        "eqeqeq": "off",
```
PhilR8, line 18: allows `==` operator

solution/ui/regulations/css/scss/_application_settings.scss
```diff
@@ -192,6 +197,15 @@ $font_size_xxl: 30px; // Aligns with the CMS Design System, but their variable i
 $font_size_hero_h1: 26px;
 $font_size_hero_h1_tablet: 20px;
 
+@mixin font-bold-keep-width {
+    text-shadow: -0.06ex 0 0 currentColor, 0.06ex 0 0 currentColor;
+
+    @supports (-webkit-text-stroke-width: 0.04ex) {
+        text-shadow: -0.03ex 0 0 currentColor, 0.03ex 0 0 currentColor;
+        -webkit-text-stroke-width: 0.1ex;
+    }
```
PhilR8, line 206: grabbed [this trick from StackOverflow](https://stackoverflow.com/questions/5687035/css-bolding-some-text-without-changing-its-containers-size/46452396#46452396) to keep character width the same while still adding font weight.

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -0,0 +1,54 @@
...
+
+const $router = useRouter();
+const $route = useRoute();
+
+const subjectClick = (event) => {
+    const subjects = $route.query.subjects
+        ? $route.query.subjects.split(",")
+        : [];
+
+    if (subjects.includes(event.target.dataset.id)) return;
```
PhilR8, line 21: don't do anything if an already selected subject is clicked again

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
```diff
@@ -68,16 +84,20 @@ const getPartsLastUpdated = async () => {
...
+const getDocList = async (requestParams = "") => {
+    policyDocList.value.loading = true;
+
     try {
         policyDocList.value.results = await getPolicyDocList({
             apiUrl: props.apiUrl,
             cacheResponse: !props.isAuthenticated,
+            requestParams,
         });
     } catch (error) {
```
thwalker6, line 102: would it make sense to have policyDocList.value.loading = false on the exception.  Its no longer loading but an error

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
```diff
@@ -86,8 +106,176 @@ const getDocList = async () => {
...
+    // create new selectedParams key for array of objects for selections
+    selectedParams.paramsArray.push({ id, name, type });
+};
+
+// method to remove selected params
+const removeSelectedParams = (paramArgs) => {
+    const { id, type } = paramArgs;
+
+    // update paramString that is used as reactive prop for watch
+    const paramStringArray = selectedParams.paramString.split("&");
```
thwalker6, line 139: what if I put &subjects=4&subjects=5.  Right now it breaks.  We use the URL as a source of truth and it is unlikely someone would manually type that but if it came to more parameter's could this be an issue? Not something that needs to be resolved but I'm in a big QA mode this second

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
```diff
@@ -86,8 +106,176 @@ const getDocList = async () => {
...
+});
+
+// called on load
+const getDocSubjects = async () => {
+    try {
+        const subjectsResponse = await getPolicyDocSubjects({
+            apiUrl: props.apiUrl,
+        });
+
+        policyDocSubjects.value.results = subjectsResponse.sort(sortSubjects);
```
thwalker6, line 187: This is setting this in alpha order?  Why not just have this done on the backend and remove that extra processing?

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
PhilR8: the `finally` block on line 104 runs no matter what, even if an exception is thrown.  So it would flip `policyDocList.value.loading` to false even with an error.

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
thwalker6: Actually might be better to do this on the backend anyways.  We have a "true sort" method on the backend we can use that would allow us to take things like numbers into consideration when it comes to sorting things.  We can easily break this ticket out but JS might be looking at this as a straight string which can cause weird ordering if there are numbers in subject names(if this is a thing)

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
PhilR8: fixed in [9417c5b](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/970/commits/9417c5b933be8035e332ad6e384acec449e67dab).  If there are duplicate query params, it just uses the first one.

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
PhilR8: Sorting in the browser isn't inefficient (see [perf stats here](https://www.measurethat.net/Benchmarks/ShowResult/465724)) but we could move sorting to the back end if we wanted to. 

Right now, though, there's still a question of which name will be used (`full_name`, `short_name`, `abbreviation`).  Keeping it on the front end lets us easily change that hierarchy; once its solidified we could move it to the back end if needed.

solution/ui/regulations/eregs-vite/src/views/PolicyRepository.vue
thwalker6: Didnt see the finally.  All good then

thwalker6 (approved):
lgtm

PR #988, opened 2023-10-10
Eregcsc 2263 fts on uploaded content

thwalker6:
Resolves # EREGCSC-2263

**Description-**

In order to have full text search available for uploaded content a new model had to be created for the index.  This will allow us to vectorize the models in an efficient manner.  This model also will take in supplemental content and federal register documents.  This new index will allow us to preform a FTS on all three resources in a single endpoint.

The parameter of resource-type is either internal(internal douments) or external( fr documents and supplemental content).
Internal documents will only appear in the search if you are logged in, otherwise it will only show external.
You can filter, by subjects, document type, categories, locations, resource-type, and with a query.

The model of uploaded files had a name change from "document_id" to "document_name"
On that model subject was changed to subjects to reflect the many to many relationship.

- Adds a new endpoint for "content-search" which allows you to search on all resources(internal and external)
- Endpoint allows filtering by subject and resource type
- Proper naming of properties of subject to subjects
- Uploaded file property of docment_id changed to document_name
- Increased code coverage
- Indexing of seed data for search
- Add command for pytest-cov in our make command(allows you to see code coverage results)

1. Unit test pass
2. Access the endpoint of v3/content-search(swagger is broken on dev environments so either do your local or just put the endpoint in the url.  If using local make sure you run make local.seed and import migrations
3.  Search for 'covid'.  You should get a list of supplemental content and fr documents in search
4.  Make an internal document have it have covid in the name or description
4.  make a search of covid not logged. /v3/content-search/?q=covid&resource-type=internal
 It should only return supplemental content fr docs but not the one you created
5.  Login and make a search for covid with a parameter of 'internal'  you should retrieve that document in search it will return only your internal document you created.


solution/backend/cmcs_regulations/settings/base.py
```diff
@@ -50,7 +50,8 @@
     'django.contrib.sitemaps',
     'django.contrib.syndication.views',
     'django_jsonform',
-    'file_manager'
+    'file_manager',
+    'content_search',
```
thwalker6, line 54: search is its own little app for resources.  This is still independent of regcore search

solution/backend/content_search/models.py
```diff
@@ -0,0 +1,138 @@
...
+            ci.supplemental_content = updated_doc
+        else:
+            ci.fr_doc = updated_doc
+        return ci
+    return None
+
+
+def update_search(sender, instance, created, **kwargs):
+    try:
+        if isinstance(instance, UploadedFile):
```
thwalker6, line 128: we are trying to find the index related to the object here.  If it cant find it, it will throw an exception and then will create it

solution/backend/file_manager/models.py
```diff
@@ -30,12 +30,12 @@ def __str__(self):
...
 
 class UploadedFile(models.Model):
-    document_id = models.CharField(max_length=512, null=True, blank=True)
+    document_name = models.CharField(max_length=512, null=True, blank=True)
     file_name = models.CharField(max_length=512, null=True, blank=True)
     date = VariableDateField(null=True, blank=True)
     summary = models.TextField(null=True, blank=True)
     internal_notes = models.TextField(null=True, blank=True)
-    subject = models.ManyToManyField(Subject, blank=True, related_name="uploads")
+    subjects = models.ManyToManyField(Subject, blank=True, related_name="uploads")
```
thwalker6, line 38: subjects make more sense since there can be multiple

solution/backend/file_manager/models.py
```diff
@@ -30,12 +30,12 @@ def __str__(self):
 
 
 class UploadedFile(models.Model):
-    document_id = models.CharField(max_length=512, null=True, blank=True)
+    document_name = models.CharField(max_length=512, null=True, blank=True)
```
thwalker6, line 33: this is part of another ticket but to make it simpler and faster for others just changed here

solution/backend/resources/tests/test_admin.py
```diff
@@ -97,7 +97,8 @@ def test_add_supplemental_content(self):
             results = self.supplementalAdmin.add_content(output)
         good_section = Section.objects.get(section_id=200)
         good_category = AbstractCategory.objects.get(name="test")
-        row1 = {"name": "<a href=/admin/resources/supplementalcontent/1/change/ target='blank'>content-1</a>",
+        id = str(SupplementalContent.objects.get(name='content-1').id)
+        row1 = {"name": f"<a href=/admin/resources/supplementalcontent/{id}/change/ target='blank'>content-1</a>",
```
thwalker6, line 101: this test was a little faulty.  If there was any supplemental content ever on the test db it fails since it wont be "1".  We just get the ID now before the check.

solution/backend/resources/views/mixins.py
```diff
@@ -72,7 +72,6 @@ def get_location_filter(self, locations):
 class LocationExplorerViewSetMixin(OptionalPaginationMixin, LocationFiltererMixin):
     PARAMETERS = LocationFiltererMixin.PARAMETERS + OptionalPaginationMixin.PARAMETERS
 
-    location_filter_prefix = ""
     location_filter_max_depth = 2
```
thwalker6, line 75: wasnt letting file upload overload the location filter prefix.  Removing it fixes it.

solution/Makefile
```diff
@@ -177,7 +177,10 @@ test.pytest:
 	DJANGO_SETTINGS_MODULE="cmcs_regulations.settings.test_settings" docker-compose -f docker-compose.yml up -d
 	docker-compose exec regulations pytest -vvv
 	docker-compose -f docker-compose.yml up -d  ## resets the django_settings_module to the default
-
+test.pytest.cov:
+	DJANGO_SETTINGS_MODULE="cmcs_regulations.settings.test_settings" docker-compose -f docker-compose.yml up -d
```
thwalker6, line 181: just a handy tool if you want tosee your code coverage percentage.  Example everyting in content_search is 92-100 percent covered

solution/backend/common/functions.py
```diff
@@ -50,7 +51,7 @@ def loadSeedData():
         ("cmcs_regulations/fixtures/auth.permission.json", Permission),
         ("cmcs_regulations/fixtures/auth.group.json", Group),
     ]
-
+    ContentIndex.objects.all().delete()
```
thwalker6, line 54: deleting the index first fixes some issues with seed data

solution/backend/content_search/signals.py
```diff
@@ -0,0 +1,14 @@
...
+from resources.models import FederalRegisterDocument, SupplementalContent
+
+from .functions import add_to_index
+
+
+@receiver(post_save, sender=FederalRegisterDocument)
+@receiver(post_save, sender=SupplementalContent)
+@receiver(post_save, sender=UploadedFile)
+def add_index(sender, instance, created, **kwargs):
+    add_to_index(instance)
```
thwalker6, line 14: If one of these three model types are saved, after it is saved add it to the index

solution/backend/pyproject.toml
```diff
@@ -1,6 +1,13 @@
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "cmcs_regulations.settings.test_settings"
 
+[tool.coverage.run]
```
thwalker6, line 4: we dont want pytest coverage to look at the following directories.

solution/backend/common/functions.py
```diff
@@ -50,7 +50,15 @@ def loadSeedData():
...
     ]
+    # The vector column for the search index has issues with using seed data.  As in it takes forever.
+    # We remove the column first, then we add it back after fixtures are done.
 
+    file_index = ContentIndex.objects.all()
+    for con in file_index:
+        con.locations.clear()
+        con.subjects.clear()
+        con.save()
+    file_index._raw_delete(file_index.db)
```
thwalker6, line 61: _raw_delete is faster but its very blunt.  Ignores things the many to many relationships and just essentially deletes the table.  Not something you always want to do but its a good use here.

solution/Makefile
```diff
@@ -200,7 +207,7 @@ file_manager_objects = file_manager.subject file_manager.documenttype
 auth_objects = contenttypes.contenttype auth.permission auth.group
 
 local.dump:
-	$(foreach object, $(objects), docker-compose exec regulations python manage.py dumpdata $(object) --indent 4 > ./backend/resources/fixtures/$(object).json;) \
+	$(foreach object, $(resource_objects), docker-compose exec regulations python manage.py dumpdata $(object) --indent 4 > ./backend/resources/fixtures/$(object).json;) \
```
thwalker6, line 210: better name

solution/ui/regulations/utilities/api.js
```diff
@@ -574,7 +574,7 @@ const getSupplementalContent = async ({
  * @returns {Promise<{count: number, next: string, previous: string, results: Array<Object>}>} - Promise that contains response object when fulfilled
  */
 const getSearchGovResources = async ({ page = 1, q = "" }) =>
-    httpApiGet(`resources/search?q=${encodeURIComponent(q)}&page=${page}`);
+    httpApiGet(`resources/search?q=${encodeURIComponent(q)}&page=${page}&location_details=true&category_details=true`);
```
thwalker6, line 577: before it never cared about the detail paramaters due to a hidden bug

solution/backend/common/serializers.py
```diff
@@ -0,0 +1,19 @@
+from drf_spectacular.utils import extend_schema_field
+from rest_framework import serializers
+
+from resources.serializers.categories import AbstractCategoryPolymorphicSerializer, MetaCategorySerializer
+from resources.serializers.locations import AbstractLocationPolymorphicSerializer, MetaLocationSerializer
+
+
+class DetailsSerializer(serializers.Serializer):
```
thwalker6, line 8: if we include subjects  into subject details we can kind of use this as a central "include or exclude" details.   Locations is currently shared amongst different types but subjects could also be included here.

solution/backend/index_resources.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+import os
+
+
```
thwalker6, line 4: this is only going to be called on experimental, dev, and val similar to seed data

cgodwin1 (approved):
LGTM!

PR #989, opened 2023-10-10
hotfix - Dev seed data issue

thwalker6:
Resolves # hotfix

**Description-**

Dev seed data seems to have some conflictions with content type IDs which we use for admin  purposes.  This is preventing those possibly troublesome tables from being imported.

- Seed data for contenttypes and permissions will no longer be seeded.


PhilR8:
This is still a draft PR

PR #991, opened 2023-10-11
Eregcsc 2012 initial logins from eua

peggles2:
Resolves #EREGCSC-2012

**Description-**
This pull request updates the admin panel code to receive the okta values that are returned.
If the user has any jobcode their status is automatically set to 'active' for now. 
If the user does not have any jobcode the user is not created in the db.
When a user signs in, if they already have an account their account just gets updated. 

- Update the README file 
- bugfix on euasettings.py
- update admin.py to create/update the user and to give user 'active' access if they have any jobcodes. 
- write tests

1. visit /admin page
2. click on eua link
3. sign into eua 
4. notice you get redirected to admin panel with this message:
Screen Shot 2023-10-10 at 1 38 56 PM
5. sign into admin panel with an admin account
6. click on users and notice the user is listed there. 
7. click on the users name
Screen Shot 2023-10-11 at 10 09 51 AM

9. The user has active status but not admin. 
Screen Shot 2023-10-11 at 10 10 05 AM

11. Give the user admin access. 
12. sign out
13. sign back in with eua, notice you now are in the admin panel through eua.


cgodwin1 (approved):
LGTM!

PR #993, opened 2023-10-13
EREGCSC-update-allowed-hosts

peggles2:
Resolves #[EREGCSC-2257](https://jiraent.cms.gov/browse/EREGCSC-2257)

**Description-**
We recently introduced a new domain, 'eregulations.cms.gov,' into our system and configured our API Gateway to accommodate it. However, we encountered an error in our AWS logs that reads: "ERROR Invalid HTTP_HOST header: 'eregulations.cms.gov.' The resolution to this issue involves adding 'eregulations.cms.gov' to the list of allowed hosts in our Django settings file, which, in our case, is 'base.py.' This adjustment will enable our application to properly handle requests from the 'eregulations.cms.gov' domain.

- base.py

This may not be easy to verify if it fixes the issue till we push to prod. But we should check make sure it does not impact the current behavior. Recommend to visit experimental link, click around ensure the site is up and running. 

All tests should pass.


PhilR8 (approved):
LGTM!

PR #994, opened 2023-10-16
EREGCSC-2257-dynamic-domain

peggles2:
Resolves #EREGCSC-2257

**Description-**
We want to update our domain name to eregulations.cms.gov. In the mean time we like to keep the domain name regulations-pilot.cms.gov and then phase it out. 
The domain has been configured yet we are getting a 400 error. This is because currently our custom_url is retrieved from AWS Paramstore. Which currently is set to regulations-pilot.cms.gov. To fix the problem, we are updating the context-processors.py that sets the custom_url value. Here we check if the request is from the domain eregulations.cms.gov. If it is, then it will replace the custom_url with eregulations.cms.gov. This will be a temp solution till we phase out regulations-pilot.cms.gov. 

- expected change:

- update context_processors.py
- write tests
- fix a bug where python tests were failing locally on an aws test. (EREGCSC-2304) 
- Fix a typo in allowed hosts. 

1. visit experimental page, make sure everything is working normally
2. make sure tests are passing.


PhilR8 (approved):
LGTM!

PR #998, opened 2023-10-16
EREGCSC-2251 Lambda for text extraction

cgodwin1:
Resolves #2251

**Description-**

This PR implements the lambda function that allows for text extraction from various file types. Currently only plain text files are supported with more to come in future PRs.

See [the readme](https://github.com/Enterprise-CMCS/cmcs-eregulations/blob/2251-text-extractor/solution/text-extractor/README.md) for more information on how the lambda works. Locally, it uses a program called "lambda-proxy" which is a temporary solution to the issue of using lambda docker containers locally.

- New lambda function "text-extractor" is deployed using Lambda Docker containers.
- Lambda is accessible via an async API Gateway endpoint (send a POST request, it returns immediately and runs the lambda in the background).
- Local transparent proxy provides access to the text extractor lambda function via `localhost:8001` outside of containers or `text-extractor:8000` within containers.
- Unit testing using unittest module, with `moto` providing mocks for AWS services.
- New environment variable available in eRegs: `TEXT_EXTRACTOR_URL`, which is the URL to send POST requests to to execute the lambda function.

1. Run unit tests (`make text-extractor.test`, container must be running, start with `make text-extractor`)
2. To verify that uploading works, we need a server that accepts POST requests to mimic eRegs. Install netcat in the Docker container with `docker-compose exec text-extractor yum install nmap-ncat`
3. In another terminal, run `docker-compose exec text-extractor nc -lk -p 8002 -c 'echo -e "HTTP/1.1 200 OK\n\n Hello World"'` to start the fake server.
4. Using your favorite tool, send a POST request to `http://localhost:8001` with the following JSON body:
```
{
    "id": 1,
    "uri": "https://raw.githubusercontent.com/Enterprise-CMCS/cmcs-eregulations/main/solution/static-assets/requirements.txt",
    "backend": "web",
    "post_url": "http://localhost:8002"
}
```
5. You should see a JSON response stating "Function exited normally." This means success, the lambda function was able to retrieve the file, parse it, and upload it to the fake server.
6. Test again but change the uri to a file that isn't plain text, for example `https://avatars.githubusercontent.com/u/102048837?v=4` which is a PNG file. See that the response is "Failed to initialize text extractor: \'image/png\' is an unsupported file type."

To view the output of the POST request sent from the Lambda containing the extracted text, Ctrl+C the netcat fake server instance and run `docker-compose exec text-extractor nc -lk -p 8002` instead. Then send the POST request again. The Lambda function will eventually timeout because the fake server doesn't send a reply without the `-c` argument, but you will see on the terminal an output looking like:
```
POST / HTTP/1.1
Host: localhost:8002
User-Agent: python-requests/2.31.0
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Length: 35

{"id": 1, "text": "This is a test"}
```
This test confirms that the Lambda will send the correct data to eRegs.

To test S3 integration locally, you will need an AWS access key and secret key. Keep in mind that the unit tests already test this, but if you wish you can get your access keys and send a request like so:
```
{
    "id": 1,
    "uri": "key_name_of_object_in_s3_bucket",
    "backend": "s3",
    "post_url": "http://localhost:8002",
    "s3": {
        "aws_access_key_id": "your_access_key_here",
        "aws_secret_access_key": "your_secret_key_here",
        "aws_storage_bucket_name": "your_bucket_name_here"
    }
}
```
You will also need a valid bucket and the name of a text file inside the bucket.


solution/lambda-proxy/pyproject.toml
thwalker6, line 1: have whatever it is point to the other pyproject.toml.  You should be able to dictate the path.  I've had to set it up for other things as well.  Dont want duplicates if possible.

solution/text-extractor/backends/s3.py
```diff
@@ -0,0 +1,35 @@
+import boto3
+import botocore.exceptions
+
+from .backend import FileBackend
+from .exceptions import BackendException, BackendInitException
+
+
+class S3Backend(FileBackend):
+    backend = "s3"
```
thwalker6, line 9: I could be wrong here but I think the point of setting the backend is so that on post_params you can easily change it.  

So it would be post_params[self.backend]["aws_access_key_id"].  That way if someone changes the class they can just e like x = S2Backend()
x.backend= 'whatever'
then the post params are more dynamic

solution/Makefile
```diff
@@ -193,6 +198,21 @@ python.migrate:
...
+text-extractor.stop:
+	docker-compose stop text-extractor
+
+text-extractor.start:
+	docker-compose start text-extractor
+
+text-extractor.restart:
+	make text-extractor.stop; make text-extractor.start
+
+text-extractor.test:
```
thwalker6, line 213: add pytest to the requirements and change this to
```
text-extractor.test:
	docker-compose exec text-extractor pytest
```

thwalker6 (approved):
lgtm

PR #1003, opened 2023-10-19
Eregcsc 2281 default sort method

thwalker6:
Resolves # EREGCSC-2281

**Description-**

Sets up pagination for content_search and file_manager. 

- Pagination is added by default to both content_search and file_manager
- Update python unittest to reflect
- Updated front end to handle the modificaiton
- Updated cypress fixture to reflect pagination

1. Go to swagger(easier to see the api)
2. For file_manager just run the query.  Will show pagination.  Turn pagination to false, will return only the results with no pagination
3. For content-search run the query.  Will show paginated results.  You can do it without pagination though its going to be over 1000 results so not advised.
4. Go to /policy-repository.
5. Click on a subject.  will display any results for that subject.
6.  All cypress and pytest will pass


solution/backend/file_manager/admin.py
```diff
@@ -65,10 +65,12 @@ def save_model(self, request, obj, form, change):
         if path:
             obj.file_name = self.clean_file_name(path._name)
             self.upload_file(path, obj)
-        form.instance.save()
-        add_to_index(form.instance)
         super().save_model(request, obj, form, change)
 
+    def save_related(self, request, form, formsets, change):
```
thwalker6, line 70: save_related is needed to take care of the many to many relationships (locations, and subjects)

solution/backend/common/test_functions/common_functions.py
```diff
@@ -0,0 +1,5 @@
+import json
+
+
+def get_paginated_data(response):
```
thwalker6, line 4: responses with pagination are returned as an ordereddict which makes comparing the data annoying.  So we convert it to a json object to make it easier to look at.

solution/backend/file_manager/admin.py
thwalker6: It was previously not doing this because it was done in the normal save method.

solution/ui/regulations/eregs-vite/src/views/PolicyRepositorySearch.vue
```diff
@@ -96,11 +96,12 @@ const getDocList = async (requestParams = "") => {
     policyDocList.value.loading = true;
 
     try {
-        policyDocList.value.results = await getPolicyDocList({
+        const contentList = await getPolicyDocList({
             apiUrl: props.apiUrl,
             cacheResponse: !props.isAuthenticated,
             requestParams,
         });
+        policyDocList.value.results = contentList.results;
```
PhilR8, line 104: this looks good to me and has been validated as working on the experimental deployment.

solution/backend/content_search/tests/fixtures/sample_files.json
```diff
@@ -12,7 +12,7 @@
         "summary": "Important notes that cannot be public"
     },
     {
-        "document_name": "Policy reference file",
+        "document_name": "Policy reference file fire fire fire",
```
peggles2, line 15: why name it fire fire fire?

peggles2 (approved):
LGTM I just added a question.

solution/backend/content_search/tests/fixtures/sample_files.json
thwalker6: search relevancy.  there is a search in unit test for fire

PR #1004, opened 2023-10-19
Fix bug where file- upload not saved before index

thwalker6:
Resolves # EREGCSC-2281

**Description-**

When a new uploaded file is being added to the index through the admin panel if it wasnt saved already it was causing issues. 

Had it save before getting sent to index fixing the issue

- When an uploaded file is saved the index gets updated correctly.

1. Log in to the admin panel and add an uploaded file
2. Will be able to upload the file successfully.


PhilR8 (approved):
lgtm

PR #1006, opened 2023-10-20
EREGCSC-2253 Extract text from PDFs (scanned & text) via Textract

cgodwin1:
Resolves #2253

**Description-**

This PR implements PDF extraction to the text-extractor lambda. It uses the Textract service.

- PdfExtractor class exists in the `extractors` directory of text-extractor.
- PDF extraction now works through Textract locally (with access keys & region) and deployed (no keys needed).
- DetectDocumentText is mocked for testing purposes.
- Binary packages are now added to `packages.txt` and installed with yum on container build.
- eRegs is deployed with a `TEXT_EXTRACTOR_ARN` environment variable that can be used for boto3.

_These steps are copied and modified from my previous PR._

1. Find some AWS access keys that DO NOT require a security token. You may need to create an IAM role with the company's AWS account to do this. If you need one and don't have one, it's easy to set up, just ask. You absolutely cannot use Cloudtamer's access keys, they require a token and that is simply not supported. You will need: `aws_access_key_id` and `aws_secret_access_key` only.
2. Run unit tests (`make text-extractor.test`, container must be running, start with `make text-extractor`)
3. To verify that uploading works, we need a server that accepts POST requests to mimic eRegs. Install netcat in the Docker container with `docker-compose exec text-extractor yum install nmap-ncat`
4. In another terminal, run `docker-compose exec text-extractor nc -lk -p 8002 -c 'echo -e "HTTP/1.1 200 OK\n\n Hello World"'` to start the fake server.
5. Using your favorite tool, send a POST request to `http://localhost:8001` with the following JSON body:
```
{
    "id": 1,
    "uri": "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf",
    "backend": "web",
    "post_url": "http://localhost:8002",
    "textract": {
        "aws_access_key_id": "YOUR_KEY_ID_HERE",
        "aws_secret_access_key": "YOUR_SECRET_KEY_HERE",
        "aws_region": "us-east-1"
    }
}
```
6. After a moment you should see a JSON response stating "Function exited normally." This means success, the lambda function was able to retrieve the file, parse it, and upload it to the fake server.

To view the output of the POST request sent from the Lambda containing the extracted text, Ctrl+C the netcat fake server instance and run `docker-compose exec text-extractor nc -lk -p 8002` instead. Then send the POST request again. The Lambda function will eventually timeout because the fake server doesn't send a reply without the `-c` argument, but you will see on the terminal an output looking like:
```
POST / HTTP/1.1
Host: localhost:8002
User-Agent: python-requests/2.31.0
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Length: 36

{"id": 1, "text": "Dummy PDF file "}
```
This test confirms that the correct text is extracted and that the Lambda will send the correct data to eRegs.


thwalker6 (approved):
lgtm

PR #1007, opened 2023-10-23
EREGCSC-2295 Add sort count to subjects

thwalker6:
Resolves # EREGCSC-2295

**Description-**
To easily list how many resources are associated to a subject we want the endpoint to reveal in a count how many are related to the subject.  There are two fields added that will list the counts:
external_resources
uploads

- expected change 1

Go into swagger.
make sure you have resources of internal and external assigned to some subjects.
Call the subjects endpoint.  
Uploads field will relfect how many internal_resources are associated to the subject.
external_resources will refer to how many abstract resources are assigned to the subect.

New unit test will pass.

### DIFF
--- a/solution/backend/file_manager/views.py
+++ b/solution/backend/file_manager/views.py
@@ -6,7 +6,7 @@ from django.contrib.postgres.search import (
     SearchRank,
     SearchVector,
 )
-from django.db.models import Prefetch, Q
+from django.db.models import Count, Prefetch, Q
 from django.http import HttpResponseRedirect
 from drf_spectacular.utils import extend_schema
 from rest_framework import viewsets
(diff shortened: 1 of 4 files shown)


solution/backend/file_manager/views.py
```diff
@@ -43,13 +44,30 @@ def list(self, request):
 
 @extend_schema(
     description="Retrieve a list of subjects.",
+    parameters=[
+                OpenApiQueryParameter("q",
+                                      "Search fors text within file metadata. Searches document name, file name, "
```
PhilR8, line 49: some strange indents here

PhilR8: otherwise this is looking good to me

thwalker6: just how the linter looks for max characters per line

thwalker6: With flake8  and other linters you need the " to match with the line before it if you put each part on its own line.  Otherwisse you will get a linting error

PhilR8 (approved):

PR #1011, opened 2023-10-25
Hotfix approved only index

thwalker6:
Resolves # Issue regarding non approved content in the index

**Description-**

This will make it so only approved FRDocs and supplemental content are indexed.  If a piece is marked to false, it will remove it from the index.

- Anything not approved will not appear in the index.

1. Click this link and you should get no results
https://3mkeajv1n1.execute-api.us-east-1.amazonaws.com/dev1011/v3/content-search/?q=bibbity

2. In the admin go to the supplemental content with the name of bibbity and switch it to approved.
3. redo the search and it will return just that piece of content
4. Switch it back to not approved and redo the search
5. It will now no longer appear in the search


peggles2 (approved):
LGTM 👍

PR #1013, opened 2023-10-26
Hotfix - remove text extractor from experimental

thwalker6:
Resolves #hotfix

**Description-**

The text extractor on experimental deploys is taking too many IP addresses and causes issues if too many experimental branches are open.  This creates a new deploy file that can be used whenever we need to make modificaitons to the extractor

- Deploy experimental will not have the extractor

1. Deploy experimental will not have the extractor


.github/workflows/deploy-experimental.yml
```diff
@@ -73,49 +73,49 @@ jobs:
...
+  #         serverless deploy --stage dev${PR}
+  #         popd
   deploy-django:
     environment:
       name: "dev"
     outputs:
       url: ${{ steps.deploy-regulations-site-server.outputs.url }}
     runs-on: ubuntu-20.04
-    needs: [deploy-static, deploy-text-extractor]
+    needs: [deploy-static, ] #[deploy-static, deploy-text-extractor]
```
PhilR8, line 118: I was curious if trailing commas are allowed in YAML syntax.  It appear that they are [indeed allowed](https://dev.to/this-is-learning/yaml-collections-sequences-and-mappings-4meb#:~:text=A%20flow%20sequence%20is%20a,but%20empty%20node%20entries%20disallowed.&text=or%20any%20other%20white%20space%20variation.).  Still, I don't think we use them anywhere else.

PhilR8 (approved):
LGTM

PR #1015, opened 2023-10-26
[HOTFIX] Clicking ToC and then adding subjects

PhilR8:
Resolves issue on production where wrong data type is being pushed into router state. A string is being treated as an array, which is splitting the string into characters when updating the route's query params.

**Description:**

The flow to reproduce this bug consistently:
1. Visit `/policy-repository`
2. Click on a subject near or at the bottom of the Table of Contents (so the subject ID has a high number with more than one digit. ex: 190)
3. Add one more subject using the sidebar subject selector

Expected result:
Two subjects will be selected

Actual result:
Three or more subjects will be selected

- ToC click will push a string wrapped in an array into router state instead of just a string so all existing functionality will work as expected
- Adds Cypress test to validate the fix

1. Follow reproduction steps above and ensure only two subjects are selected, and that they're the correct two subjects
2. Make sure new e2e test passes


thwalker6 (approved):
lgtm

PR #1016, opened 2023-10-27
EREGCSC-2312 -- Add `rel="noopener noreferrer"` to all external links

PhilR8:
Resolves [EREGCSC-2312](https://jiraent.cms.gov/browse/EREGCSC-2312)

**Description:**

In order to meet MACPRO ACT (ATO) expectations, we need to ensure that all links are marked as "noopener".

This story was quite a bit more complicated than initially thought, but now every external link should (?) be properly marked with the `rel` tag and we have an end to end test for it.

- Adds `rel="noopener noreferrer"` to all anchor elements that have `target="_blank"`, which means they are external links (whether they have the `external` class or not)
- Updates pytest fixtures to reflect addition of `rel` tag
- Creates a Cypress utility method to check that these links have the appropriate `rel` tag and values

1. Click around site and inspect external links to see if they have the `rel` tag and necessary values
2. All tests pass


solution/backend/regulations/templates/browser_not_supported.html
```diff
@@ -1,6 +1,13 @@
...
 
-<div id="browser-unsupported-modal" class="ds-c-dialog-wrap" data-state-name="browser-unsupported-modal" data-state="shown" aria-labelledby="dialog-heading" role="dialog">
+<div
+    id="browser-unsupported-modal"
+    class="ds-c-dialog-wrap"
+    data-state-name="browser-unsupported-modal"
+    data-state="shown"
+    aria-labelledby="dialog-heading"
+    role="dialog"
+>
```
PhilR8, line 10: formatting change

peggles2 (approved):
LGTM 👍

PR #1021, opened 2023-10-31
EREGCSC-2325 Enable text extractor to work on EREGS

thwalker6:
Resolves # EREGCSC-2317 Hook up text extractor to eregs

**Description-**

Currently EREGS does not actually link up to the text extractor.  This allows the connection.

- Creates an internal network for our docker containers to communicate over local host.  We need to do this in order for the containers to talk to eachother over locahost.  instead of localhost though it is actually /host.docker.internal:{{port-number}}
- Creates a connection between main eregs and the text extractor.
- **IF** there are no changes that they need to make between on the text extractor it will link up to the main text exrractor on dev instead of doing a deploy.
- If the are making changes to the text extractor they will need to uncomment out the text extractor on the serverless-experimental file.  This will handle the rest.
- To use a localized text extractor you will need to make sure the dockerfile for the text extractor has the keys for textract.  Otherwise it cant use it for PDF scanning.
- if set to false it will actually connect to dev text extractor.  This wont work properly until the text extractor is deployed to the main branch.
- Adds a button in the admin panel to extract text from whatever piece of content is associated with it.  Only works for text files and PDF's currently.
-  The admin panel will tell you if the index has been populated yet.  It will also display the first 100 characters of the populated value if it has been.

1. Log into the admin panel on experimental and upload a piece of content.  https://lhrl5e2swd.execute-api.us-east-1.amazonaws.com/dev1021/admin/
2. After the file has been uploaded and saved notice that it will say "Not populated" under "Index populated".
3. Click the button for get content.
4. Should redirect you to an api page with a 202 response.  this says that it connected to the lambda essentially.
5. Go back to the admin page for the content, it should say "Populated: " + text of the document.
6. Also works for supplemental content and FR docs.  You can try those too if you like.  The only issue with them is some of them are just large PDFS(hundreds of pages) so just find one with a smaller PDF for it for testing purposes.  We need to add the ability to add only selected pages into the PDF in some cases but that is a future story.


solution/backend/cmcs_regulations/urls.py
```diff
@@ -39,6 +40,8 @@
     path('sitemap.xml', sitemap, {'sitemaps': sitemaps}, name='django.contrib.sitemaps.views.sitemap'),
     path('latest/feed/', ResourceFeed()),
     path('oidc/', include('mozilla_django_oidc.urls')),
+    path('api/token/', jwt_views.TokenObtainPairView.as_view(), name='token_obtain_pair'),
```
thwalker6, line 43: Creates a jwt token if needed for using the api on an authenticated account.

solution/backend/content_search/functions.py
```diff
@@ -130,3 +133,16 @@ def index_group(resources):
     '''
     for res in resources:
         add_to_index(res)
+
+
+def build_lambda_client():
+    if settings.USE_AWS_TOKEN:
+        # Assumes we are using credentials.  Works for local but an only post to an external URL.
+        return boto3_client('lambda',
```
thwalker6, line 141: we need less credentials when running on the lambda to lambda than when we go local to aws.

solution/backend/content_search/views.py
```diff
@@ -117,3 +123,60 @@ def post(self, request, *args, **kwargs):
...
+        index = ContentIndex.objects.get(uid=uid)
+        if not index:
+            return Response({'message': "Index does not exist"})
+        json_object = {
+            'id': uid,
+            'uri': index.url,
+            'post_url': post_url,
+            'token': token,
+        }
+        if index.file:
```
thwalker6, line 151: If it is an uploaded file it needs to get the bucket credentials

solution/backend/content_search/views.py
```diff
@@ -117,3 +123,60 @@ def post(self, request, *args, **kwargs):
...
+                json_object['backend'] = 's3'
+                json_object["s3"] = {
+                                        "aws_access_key_id": settings.AWS_ACCESS_KEY_ID,
+                                        "aws_secret_access_key": settings.AWS_SECRET_ACCESS_KEY,
+                                        "aws_storage_bucket_name": settings.AWS_STORAGE_BUCKET_NAME
+                                    }
+            except ValueError:
+                json_object['backend'] = 'web'
+
+        if settings.USE_LOCAL_TEXTRACT:
```
thwalker6, line 163: we need textract credentials if running locally

solution/docker-compose.yml
```diff
@@ -1,4 +1,8 @@
 version: '3'
+networks:
```
thwalker6, line 2: internal network for our docker containers

solution/backend/file_manager/admin.py
```diff
@@ -42,13 +42,13 @@ class Meta:
 @admin.register(UploadedFile)
 class UploadedFileAdmin(BaseAdmin):
     form = UploadAdminForm
-    list_display = ("document_name",)
+    list_display = ("document_name", 'index_populated')
```
peggles2, line 45: can we keep the quotes consistent?

solution/backend/resources/admin.py
```diff
@@ -603,6 +628,25 @@ def get_queryset(self, request):
...
         )
 
+    def get_content(self, obj):
+        if obj.id:
+            index = ContentIndex.objects.get(fr_doc=obj)
+            link = reverse("call-extractor", kwargs={"content_id": index.uid})
+            html = '<a class="button" href="{}">Get content</a>'.format(link)
+        else:
+            return "N/A"
+        return format_html(html)
```
peggles2, line 638: since else returns "N/A" does that mean only if obj.id we get format_html(html) so does it make sense to have it in that if statement?

peggles2 (changes requested):
looks good, just a few minor tweeks.

solution/backend/resources/admin.py
thwalker6: We dont need to have it format html in this case since its just a string

solution/backend/file_manager/admin.py
thwalker6: updated

solution/backend/resources/admin.py
thwalker6: corrected

peggles2 (approved):
LGTM 👍

PR #1028, opened 2023-11-02
EREGCSC-2293 -- "Download X" chips on subject pages for uploaded files

PhilR8:
Resolves [EREGCSC-2293](https://jiraent.cms.gov/browse/EREGCSC-2293)

**Description**

Implement file type indications on subject lookup pages for internal docs (uploaded files).

Not for PDF, just for other types of files like XLSX, MSG, DOCX, PPTX, etc.

- Uses the `file_name_string` field for uploaded docs to infer a document's filetype
- Essentially we're checking: 
   - if `file_name_string` is a `string`
   - if is is a string, it has a period (`.`)
   - and that period is not the final character in the string
- If all those conditions are true, we split `file_name_string` on `.` and `pop` the final item in the array.  If that item is a string that is two to four characters long, and if it doesn't contain `pdf`, we display it as a Download button.

1. Visit [experimental deployment](https://asaliew7o2.execute-api.us-east-1.amazonaws.com/dev1028)
2. Look at [Figma](https://www.figma.com/file/IPqcMzYCoqAzKspjOqOsPE/Download-Indicators?type=design&node-id=128-5276&mode=design&t=ZDBAGS4ebNx64mxu-0) for design
3. View files in Policy Repository and make sure PDF files don't show a chip but other filetypes show the expected chip
4. Upload your own files with varied filetypes and filenames to validate display logic


peggles2 (approved):
LGTM 👍

PR #1032, opened 2023-11-03
EREGCSC-2291 -- Subjects Typeahead

PhilR8:
Resolves [EREGCSC-2291](https://jiraent.cms.gov/browse/EREGCSC-2291)

**Description**

We want to offer the equivalent of "jump to" for looking up a subject by enabling a person to start typing the subject name to find the subject they want.

- Adds text input to top of Subject Selector component in the left sidebar of the Policy Repository page
- Implements functionality that filters the subject list based on what is typed into the text input at the top of the Subject Selector
- Styles filtered subjects list based on [Figma design](https://www.figma.com/file/KiWw729Nct6tY9gzmvEkVO/Subject-filter-only?type=design&node-id=3-20&mode=design&t=JekXTsHzWcDuzcmW-0)
- Adds e2e test coverage

1. Make sure all tests pass
2. Visit [experimental deployment](https://i5m9y4e517.execute-api.us-east-1.amazonaws.com/dev1032/policy-repository/)
3. Make sure "Filter the subject list" input exists.  Type in some random stuff and see if the subjects list is filtered as expected
4. Notice that the label for the subject can change if the filter string matches on a field that is not the normal display name.  Search for "21" to see "Cures Act" become "21st Century Cures Act"


solution/ui/regulations/.eslintrc.json
```diff
@@ -18,6 +18,7 @@
         "eqeqeq": "off",
         "no-console": ["error", { "allow": ["warn", "error"] }],
         "no-nested-ternary": "off",
+        "prefer-template": "off",
```
PhilR8, line 21: we should be able to use string concatenation if we want to.

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -15,42 +20,140 @@ const props = defineProps({
 const $router = useRouter();
 const $route = useRoute();
 
+const state = reactive({
+    filter: "",
+    subjects: [],
+});
```
PhilR8, line 26: new local state for subject filter querystring and the array of filtered subjects.  This will not be stored in the URL

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -15,42 +20,140 @@ const props = defineProps({
...
+        }
+
+        state.subjects = props.policyDocSubjects.results;
+    }
+);
+
+const getFilteredSubjects = (filter) => {
+    if (!filter || filter.length < 1) {
+        state.subjects = props.policyDocSubjects.results;
+        return;
```
PhilR8, line 43: we use the passed in list of subjects from props if `filter` reactive object entry is falsy

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -15,42 +20,140 @@ const props = defineProps({
...
+    }
+
+    state.subjects = props.policyDocSubjects.results.reduce((acc, subject) => {
```
PhilR8, line 46: if there is a `filter` querystring, this `reduce` block will create an array of matching subjects with added html markup to highlight the matched substring

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -15,42 +20,140 @@ const props = defineProps({
...
+            };
+
+            return [...acc, newSubject];
+        }
+
+        return acc;
+    }, []);
+};
+
+const debouncedFilter = _debounce(getFilteredSubjects, 100);
```
PhilR8, line 88: we want to debounce this filter method for performance reasons.  Fast typers may see batched filtering but I'm guessing no one will even notice.

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -15,42 +20,140 @@ const props = defineProps({
...
+
+watch(() => state.filter, debouncedFilter);
+
 const subjectClick = (event) => {
     const subjects = $route?.query?.subjects ?? [];
     const subjectsArray = _isArray(subjects) ? subjects : [subjects];
-    const subjectToAdd = event.target.dataset.id;
+    const subjectToAdd = event.currentTarget.dataset.id;
```
PhilR8, line 95: target vs currentTarget is one of those JS gotcha questions that trip up even seasons devs in job interviews

solution/ui/regulations/eregs-vite/src/components/policy-repository/SubjectSelector.vue
```diff
@@ -15,42 +20,140 @@ const props = defineProps({
...
+        <div class="subjects__list-container">
+            <template v-if="props.policyDocSubjects.loading">
+                <div class="subjects__loading">
+                    <SimpleSpinner />
+                </div>
+            </template>
+            <template v-else>
+                <form>
+                    <label for="subjectReduce">Filter the subject list</label>
+                    <input id="subjectReduce" v-model="state.filter" type="text" />
```
PhilR8, line 130: another win for Vue's `v-model`.  Two-way data binding means we don't need to explicitly add an event handler on the input to update local state.

thwalker6 (approved):
lgtm

PR #1041, opened 2023-11-07
hotfix deployexperimental lambda perm fix

thwalker6:
Resolves # HOTFIX lambda resource

**Description-**

The text extractor arn is different on normal environments and needs to be the one for main dev.  Right now its trying to grab one that doesnt exist.

- the permission granted to the lambda is correct.

1. Deploy experimental builds properly


PhilR8 (approved):
lgtm

PR #1047, opened 2023-11-13
[Hotfix] Search page resources external link style regression

PhilR8:
Resolves regression involving external link icon on the Resources results on the combined search page.

See this link: https://eregulations.cms.gov/search/?q=%22Electronic%20Visit%20Verification%22

**Description:**

When styling the policy repository results and adding the Download File chip, some styles leaked into the Combined Search page's Resources column and the external link icon was out of place:

Screenshot 2023-11-10 at 11 17 20 AM

The description of the result item would also become split unexpectedly.

Screenshot 2023-11-10 at 11 18 52 AM

- Tweaks resources page result item style so that it is not affected by Policy Repository style

1. Visit the [experimental deployment's Search Page using the same query string as used in the prod link above](https://p8okdwd1j7.execute-api.us-east-1.amazonaws.com/dev1047/search/?q=%22Electronic%20Visit%20Verification%22).
2. Ensure that the external link icon is not misplaced (compare to the external link icon found in the results on the Resources page)
3. Ensure that item descriptions are not split. (see attached screenshots to find the result item with which to compare)

### DIFF
--- a/solution/ui/regulations/css/scss/partials/_search.scss
+++ b/solution/ui/regulations/css/scss/partials/_search.scss
@@ -54,6 +54,15 @@
 
             .resources-results-content {
                 @include common-results-styles;
+
+                .result {
+                    &__link {
+                        a {
+                            text-decoration: none;
+                            display: block;
+                        }
+                    }
+                }
             }
 
             .search-results-count {


thwalker6 (approved):
lgtm

PR #1050, opened 2023-11-14
Eregcsc 2350- Alphabetize subjects non-case-sensitively

thwalker6:
Resolves # EREGCSC-2022

**Description-**

Subjects were caring about case sensitivity.  This standardizes it.  Also will take numbers in consideration for a true sort.

- Subjects no longer care about case when ordering subjects
- Subjects will order items with numbers appropriately.  
- Subject sorting is switched from frontend to backend.

1. Go to the policy repository page
2. Should appear in alphabetical order regardless of case.  I switched "adult day health care" to lower case and it was correct in ordering.


solution/backend/common/fields.py
```diff
@@ -74,6 +74,16 @@ def naturalize_int_match(match):
         return string
 
 
+#  Allows you to pass in an array of fields instead of just one.  It checks to see if the other fields are blank
+#  To determine sorting, otherwise will do the same as natural sort field.
+class CombinedNaturalSort(NaturalSortField):
+    def pre_save(self, model_instance, add):
+        for field in self.for_field:
```
thwalker6, line 81: we return the first field that has anything in it for our ordering.  So if shortname is empty then it will go to the next one which is abbreviation. The higherarchy is defined in the model.

solution/backend/file_manager/models.py
```diff
@@ -24,6 +24,7 @@ class Meta:
     full_name = models.CharField(max_length=512, null=False, blank=False)
     short_name = models.CharField(max_length=50, null=True, blank=True)
     abbreviation = models.CharField(max_length=10, null=True, blank=True)
+    combined_sort = CombinedNaturalSort(['short_name', 'abbreviation', 'full_name'], null=True)
```
thwalker6, line 27: Sticks with the current hierarchy of ordering in shortname, abbreviation, and full name.  We do this on the front end.

solution/ui/regulations/utilities/filters.js
```diff
@@ -74,25 +74,11 @@ const locationUrl = ({ title, type, part, section_id, subpart_id }, base) => {
...
- */
-const sortSubjects = (a, b) => {
-    const aName = getSubjectName(a);
-    const bName = getSubjectName(b);
-
-    if (aName < bName) return -1;
-    if (aName > bName) return 1;
-
-    return 0;
-};
```
PhilR8, line 89: I approve of all of this

peggles2 (approved):
LGTM 👍

PR #1053, opened 2023-11-15
EREGCSC-2354 -- only allow one subject to be selected at a time

PhilR8:
Resolves [EREGCSC-2354](https://jiraent.cms.gov/browse/EREGCSC-2354)

**Description**

If I'm viewing a subject and choose another subject, replace the current materials with the new materials. (No other changes to persistence of search query, etc.)

- When clicking on a subject in the subject selector, the clicked subject overwrites any existing subject(s) in the URL
- Updates cypress tests to reflect this new reality

1. Visit the experimental deployment
2. Start clicking on subjects using the left sidebar's Subject Selector
3. There the subject you click should always replace the existing subject.  There should never be a time where two subjects are selected at once (unless you manually edit the URL to do so).


peggles2 (approved):
LGTM 👍

PR #1055, opened 2023-11-16
EREGCSC-2350 add to lowercase to subjct sorting on front end

thwalker6:
Resolves # EREGCSC-2350

**Description-**

Add tolowercase to the subject chips sorting.  Was missed on previous PR

- The chips  that are added that are selected had to get a capitalization check.  The ordering is somethign that is only done on the front end.


PhilR8 (approved):
let's get this merged!

PR #1057, opened 2023-11-16
Eregcsc 2264 -DOC, Docx added to text extractor

thwalker6:
Resolves # EREGCSC-2264

**Description-**

Enables the parsing of docx and doc files for our text extractor.  Uses the python library textract(different from aws textract)

- Word documents can now be parsed.
- no longer use Mime types to figure out file type.  Inconsistent with smaller files
- Only one set of credentials needed for invoking the text extractor as long as the IAM role has permissions for both s3 bucket and textract use.
-  Doc files are binary so it does not try to use textract instead does some decoding work using olefile while docx is xml format so textract handles that.  olefile might need to be used for other file types as well, but so far just .doc

1. Log in to the admin panel and upload a .docx file
2. Hit get content after it is uploaded.
3. Text populates for it.
 4.  Doc files have an implementation in place but not a great one.  It has some trouble with certain formatting in .doc files but extracts text from the main body seemingly fine.


solution/text-extractor/backends/s3.py
```diff
@@ -29,10 +31,16 @@ def __init__(self, config: dict):
         except Exception as e:
             raise BackendInitException(f"failed to initialize AWS client: {str(e)}")
 
-    def get_file(self, uri: str) -> bytes:
+    def get_file(self, temp_directory: str, uri: str) -> bytes:
```
thwalker6, line 34: The file is going to be stored in a temporary directory in the docker container as it is running.  When the lambda is done it deletes the file by itself.

solution/text-extractor/backends/s3.py
```diff
@@ -29,10 +31,16 @@ def __init__(self, config: dict):
...
         try:
-            obj = self.client.get_object(Bucket=self.aws_storage_bucket_name, Key=uri)
-            return obj["Body"].read()
+            file_path = os.path.join(temp_directory, uri.split('/')[-1])
```
thwalker6, line 36: We no longer pass around the object, we store it in place.

solution/text-extractor/backends/web.py
```diff
@@ -7,11 +9,14 @@
 class WebBackend(FileBackend):
     backend = "web"
 
-    def get_file(self, uri: str) -> bytes:
+    def get_file(self, temp_directory: str, uri: str) -> bytes:
```
thwalker6, line 12: We store the response content in an object then delete it at the end so same as the way it is done in s3.

solution/text-extractor/extractors/word.py
```diff
@@ -0,0 +1,21 @@
...
+    file_types = (
+                  'docx',
+                  'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+                  'doc'
+                  )
+
+    def extract(self, file_path: str) -> str:
+        text = textract.process(file_path)
+        # cleans up weird characters
+        line = re.sub(r'(\n)+', ' ', text.decode('utf-8').encode('ascii', 'ignore').decode('utf-8'))
```
thwalker6, line 18: Cleans up image characters and unicode stuff.  I used the retro doc and it had all the fun little images which we dont want in a word doc text search

solution/text-extractor/text_extractor.py
```diff
@@ -42,55 +42,53 @@ def handler(event: dict, context: dict) -> dict:
         token = config["token"]
     except KeyError:
         return lambda_response(400, "You must include 'id', 'uri', and 'post_url' in the request body.")
+    with TemporaryDirectory() as temp_dir:
```
thwalker6, line 45: this directory creates a temporary space for the file to be stored.  We use this to pass it around.  It exist in all functions within the with statement.

thwalker6: Code below doesnt really change its just tabbed over.  The  only difference is we pas in the location of the file we create.

solution/text-extractor/text_extractor.py
```diff
@@ -42,55 +42,53 @@ def handler(event: dict, context: dict) -> dict:
...
-        return lambda_response(500, f"Backend unexpectedly failed: {str(e)}")
-
-    # Determine the file's MIME type
-    # Magic docs recommend using first 2048 bytes of the file for most accurate type detection
-    try:
-        file_type = magic.from_buffer(file[:2048], mime=True)
-    except Exception as e:
-        return lambda_response(500, f"Failed to determine file type: {str(e)}")
+        try:
+            file_type = uri.split('.')[-1]
```
thwalker6, line 59: Mime types have a hard time with smaller documents.  We are just going to assume the extension is correct instead.

solution/Dockerfile.template
```diff
@@ -25,9 +25,6 @@ ENV OIDC_OP_JWKS_ENDPOINT=''
 ENV FILE_MANAGER_AWS_ACCESS_KEY_ID=""
 ENV FILE_MANAGER_AWS_SECRET_ACCESS_KEY=""
 ENV FILE_MANAGER_AWS_BUCKET_NAME="" 
-ENV TEXTRACT_ARN=''
-ENV TEXTRACT_KEY_ID = ''
```
thwalker6, line 29: Dont need these anymore

solution/text-extractor/extractors/binary.py
```diff
@@ -0,0 +1,28 @@
...
+                  )
+
+    def extract(self, file_path: str) -> str:
+
+        #  Solution taken from here https://stackoverflow.com/questions/64397811/reading-a-doc-file-in-memory.  Doc files
+        #  This is used for binary types.  So far just .doc but might include .xls.  We shall see.
+        with open(file_path, 'rb') as file:
+            f = ofio(file)
+            data = f.openstream('WordDocument').read()
+            data = data.decode('latin-1', errors='ignore')
```
thwalker6, line 20: worked fine on 9/10 files not an ideal solution but it is one.

solution/text-extractor/extractors/binary.py
```diff
@@ -0,0 +1,28 @@
+import re
+
+from olefile import OleFileIO as ofio
+
+from .extractor import Extractor
+
+
+class BinaryExtractor(Extractor):
```
thwalker6, line 8: Binary extractor is the name for now.  This solution will probably only work for .doc files in the end though.

cgodwin1 (approved):

PR #1067, opened 2023-11-20
[Hotfix] Remove unnecessary npm install command

PhilR8:
Resolves issue preventing deployment

**Description**

A [Snyk-created Pull Request](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1062) illuminated a vestigial and unused `package.json` file.  That file was removed along with its corresponding `package-lock.json` file, but unfortunately this caused the `deploy text extractor lambda` job in `deploy.yml` to fail.

After some investigation, it appears there is an unecessary `npm install` command in the `depoy text extractor lambda` job instructions:

```
      - name: deploy text extractor lambda
        run: |
          pushd solution/text-extractor
          npm install serverless -g
          npm install
          serverless deploy --stage ${{ matrix.environment }}
          popd
```

There is no `package.json` file in `solution/text-extractor`, but there previously had been one in the parent folder, which was removed with the Snyk PR references above.  The theory is that `npm install` was using that parent `package.json` file and not failing, even though that `package.json` file did not contain anything necessary for the text extractor lambda to be deployed and work successfully.

It's likely that the `deploy text extractor lambda` job instructions were copied from another job: `deploy static assets to AWS`.  That job and its associated directory DOES have a `package.json` file with some additional dependencies that need to be installed, which is why `npm install` is needed for that job. 

This theory was tested using the experimental deployment -- we removed `npm install` from the same instructions used to deploy the text extractor lambda for an experimental deployment, and the lambda deployed successfully and the functionality was tested and validated as working as expected.

- removed `npm install` command from `deploy text extractor lambda` job in `deploy.yml`

1. Text Extractor lambda deploys successfully
2. Text extraction functionality works as expected

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -73,48 +73,48 @@ jobs:
           npm install
           serverless deploy --stage dev${PR}
           popd
-  deploy-text-extractor:
-    environment:
-      name: "dev"
-    runs-on: ubuntu-20.04
-    steps:
-      # Checkout the code
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          submodules: true
-      # Find the PR number.  This is not always trivial which is why this uses an existign action
-      - name: Find PR number
-        uses: jwalton/gh-find-current-pr@v1
-        id: findPr
-        with:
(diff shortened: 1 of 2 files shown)


thwalker6 (approved):
lgtm

PR #1073, opened 2023-11-27
EREGCSC-2370 -- Move "Filter the subject list" into input field

PhilR8:
Resolves [EREGCSC-2370](https://jiraent.cms.gov/browse/EREGCSC-2370)

**Description**

Move explicit label into input as a placeholder to better prompt users to use the subject filter. Based on user research session.

- Removes explicit label and its associated styles
- adds label contents into input's `placeholder` attribute
- adds `aria-label` attribute for a11y
- asserts that `placeholder` attribute exists and is correct in e2e test suite

1. Visit policy repository on the experimental deployment
2. Make sure the subject filter input only has a placeholder and not an explicit label (compare to prod)


peggles2 (approved):
LGTM 👍

PR #1074, opened 2023-11-27
EREGCSC-2371 -- Add Document Type labels to Policy Repo search results

PhilR8:
Resolves [EREGCSC-2371](https://jiraent.cms.gov/browse/EREGCSC-2371)

**Description**

Before the Category and SubCategory labels, we need to add a new label that indicates if a result is an Internal document or a Public document.

- adds new `DocTypeLabel` Vue component and related styles
- slightly amends `DOCUMENTS_TYPE_MAP` to exclude the "resources" suffix so that it is more portable

1. Visit policy repository on experimental deployment and add some filters
2. Public documents should have a "Public" label with a "users" icon
3. Internal documents should have an "Internal" label with a "key" icon
4. Make sure it reasonably matches the labels on [this Figma comp](https://www.figma.com/file/D6uem0JreimXr433ttqKwR/Doc-type-filtering?node-id=30%3A20928&mode=dev)


cgodwin1 (approved):
LGTM!

PR #1079, opened 2023-11-28
EREGCSC-2271 - Setup mapping login from EUA

peggles2:
Resolves #EREGCSC-22071-setup mapping login from EUA

**Description-**
Now that we have established the job codes. Map the job codes to the usergroups when the user signs in. 
1. After user signs in via EUA for the first time, create the user and assign the user to correct jobcode.
2. If user signs in after the first time, update the users jobcode incase its different. 
3. Create user group for all the defined jobcodes. "EREGS_READER, EREGS_EDITOR, EREGS_MANAGER, EREGS_ADMIN"
4. Usergroups should be created even if the jobcode has a _V (val) or _P (prod) extention EREGS_READER_V should map to EREGS_READER user group. 
5. If a user signed in and had the jobcode EREGS_READER but the next time they sign in they been upgraded to EREGS_ADMIN they will automatically be in the EREGS_ADMIN group. Same if the user was EREGS_ADMIN then next time they signed in they were EREGS_READER

Currently on production the user groups defined are 
"EREGS_READER, EREGS_EDITOR, EREGS_MANAGER, EREGS_ADMIN"
But the code created the following user groups: "EREGS-READER, EREGS-EDITOR, EREGS-MANAGER, EREGS-ADMIN" 
In this ticket we made it consistent with what prod had originally as it matches better with what the jobcodes are as well. 

expected change 4

1. following files were updated it to use correct jobcodes:

- [solution/backend/cmcs_regulations/fixtures/auth.group.json](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1079/files#diff-84eb48fede3564baaf2d76d310e1fb4348f634d9b7af43bde96be864148e4288)
- [solution/backend/createsu.py](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1079/files#diff-40a651166b9807561475f3302d7c16bc3f46ee43a7739f18f0d544ebabd333aa)

2. solution/backend/regulations/admin.py
( Updated to map the user job code to the user groups when creating/updating the user.)

3. [solution/backend/regulations/tests/test_admin.py](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1079/files#diff-52a86961c008debd983e6b67219d29bc39675f1f8f1ebb112abd37636fb4f1b9)
( Updated tests ) 

1. Visit [admin login](https://h2hzyiip2l.execute-api.us-east-1.amazonaws.com/dev1079/admin)
2. First sign in as an admin and delete your EUA User account if you have one already there. Sign out and go back to the admin age to login. 
3. Sign in with EUA. 
4. Check your user and see that you have been added to EREGS_ADMIN group. 
5. Make sure your user is staff, active, and a super user. 
6. Make sure all tests pass.


solution/backend/createsu.py
```diff
@@ -9,8 +9,10 @@ def handler(self, *args, **options):
...
     from django.contrib.auth.models import Group, User
 
-    e_regs_admin_group, _ = Group.objects.get_or_create(name='EREGS-ADMIN')
-    e_regs_reader_group, _ = Group.objects.get_or_create(name='EREGS-READER')
+    e_regs_admin_group, _ = Group.objects.get_or_create(name='EREGS_ADMIN')
+    e_regs_reader_group, _ = Group.objects.get_or_create(name='EREGS_READER')
+    e_regs_editor_group, _ = Group.objects.get_or_create(name='EREGS_EDITOR')
+    e_regs_manager_group, _ = Group.objects.get_or_create(name='EREGS_MANAGER')
 
     if not User.objects.filter(username=os.environ.get('DJANGO_ADMIN_USERNAME')).exists():
```
thwalker6, line 17: do you want to just set this to check if they are "EREGS_ADMIN".  Since we have the multiple groups and only they should be super user it might make sense to change that now

solution/backend/regulations/views/policy_repository.py
```diff
@@ -12,7 +12,7 @@ def get_context_data(self, **kwargs):
         host = self.request.get_host()
         is_authenticated = self.request.user.is_authenticated
         user_groups = [group.name for group in self.request.user.groups.all()]
-        has_editable_job_code = user_groups and user_groups != ['EREGS-READER']
+        has_editable_job_code = user_groups and user_groups != ['EREGS_READER']
```
thwalker6, line 15: Probably not good logic here.  What if they had a user group that is not eregs related and only eregs reader.  They would then by definition have an editable job code.  I would make the logic to check that they are in a group that can edit, not one that cannot.

peggles2: Why would they have a user group that is not EREGS? These are user groups not jobcodes. We only offer eregs user groups. None eregs job codes will not get assigned to a user group?

thwalker6 (approved):
lgtm

thwalker6 (changes requested):
linter issues

PR #1082, opened 2023-11-30
Eregcsc 2358

thwalker6:
Resolves #

**Description-**

- expected change 1


solution/backend/common/functions.py
```diff
@@ -1,72 +1,11 @@
 import boto3
 from django.conf import settings
-from django.contrib.auth.models import Group, Permission
```
thwalker6, line 3: If we put common functions in the wrong file it causes a circular dependency over the seed data function.  Moved it to its own file to avoid this.

solution/backend/content_search/functions.py
```diff
@@ -22,55 +23,44 @@ def check_index(content):
         return None
```
thwalker6, line 25: this is mostly due to just refactoring duplications in the code.

solution/backend/cmcs_regulations/settings/base.py
```diff
@@ -310,3 +310,6 @@
     "http://*.analytics.google.com",
 ]
 CSP_INCLUDE_NONCE_IN = ["script-src"]
+
+BASIC_SEARCH_FILTER = os.environ.get("BASIC_SEARCH_FILTER", .1)
```
thwalker6, line 314: these values are configured in the aws parameter store in all environments

cgodwin1 (approved):

PR #1085, opened 2023-12-01
EREGCSC-2267 Sanitize filenames

thwalker6:
Resolves # EREGCSC-2267 Sanitize file names

**Description-**

We want to make sure filenames dont have weird characters in it.  Before we only did commas but now we are just adding a slew of them

- Use djangos slugify to clean out characters. https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.text.slugify

1. Log into the admin panel.
2. Upload a file with a bunch of weird characters. Quotes, underscores, hyphens etc.
3. Characters will be cleaned out after saving.  Should be mostly alphanumeric.

### DIFF
--- a/solution/backend/file_manager/admin.py
+++ b/solution/backend/file_manager/admin.py
@@ -59,7 +59,7 @@ class UploadedFileAdmin(BaseAdmin):
         path = form.cleaned_data.get("file_path")
         if path:
             file_name, extension = path._name.split('.')
-            obj.file_name = f"{slugify(file_name).replace('-', ' ')}.{extension}"
+            obj.file_name = f"{slugify(file_name)}.{extension}"
             self.upload_file(path, obj)
         super().save_model(request, obj, form, change)
 
--- a/solution/backend/file_manager/admin.py
+++ b/solution/backend/file_manager/admin.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.contrib import admin
 from django.urls import reverse
 from django.utils.html import format_html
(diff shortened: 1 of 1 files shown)


peggles2 (changes requested):
Nice work. Can you please write a simple test for this as well?

thwalker6:
> Nice work. Can you please write a simple test for this as well?

Slugify if a django library which would already have its own test.  I wouldnt write a unit test for it.

solution/backend/file_manager/admin.py
```diff
@@ -54,16 +55,11 @@ class UploadedFileAdmin(BaseAdmin):
...
-    def clean_file_name(self, name):
-        name = name.replace(',', '')
-        return name
-
     def save_model(self, request, obj, form, change):
         path = form.cleaned_data.get("file_path")
         if path:
-            obj.file_name = self.clean_file_name(path._name)
+            file_name, extension = path._name.split('.')
+            obj.file_name = f"{slugify(file_name).replace('-', ' ')}.{extension}"
```
peggles2, line 62: Although you use slugify you are also replacing the '-' with ' ' so a simple test would be good to have here that we expect the file names to be cleaned up in this way. Thanks

peggles2 (changes requested):

solution/backend/file_manager/admin.py
thwalker6: removed replace to simplify the code.

PR #1086, opened 2023-12-04
Extractor conditonal deploy

thwalker6:
Resolves # None.  An idea I came up with

**Description-**
Its very annoying working with the text extractor.  Commenting and uncommenting out code on deploy is tedious and just awful.  A much simpler solution is to have a conditional on a github action that determines if it should run.  
On line 84 on deploy-experimental it just has if: false. Change it to if: true if you want a text extractor.
Then on line 122 change the needs to be the commented out portion.  Needs cant actually be a conditional. Thanks github.

1. The intieal deployment deploys sucessfully.
2. Change the deploy to true for deploying the text extractor on line 84 and on 122 change it to needs: [deploy-static, deploy-text-extractor]
3. A text extractor will deploy to the experimental.
4. Closing the PR will conditionally delete the text extractor based on if one exist.

Notes.  Once a text extractor is deployed you can no longer revert back without closing the PR.  The way cloud formation looks up the extractor arn, it will find it until the extractor is deleted in aws.

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -80,7 +80,7 @@ jobs:
       deployextractor: false
     runs-on: ubuntu-20.04
     # Change to true if you want to create text extractor for your experimental deploy.
-    if: false
+    if: true
     steps:
       # Checkout the code
       - name: Checkout
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -81,7 +81,7 @@ jobs:
       deployextractor: false
     runs-on: ubuntu-20.04
     # Change to true if you want to create text extractor for your experimental deploy.
-    if: true
+    if: false
     steps:
       # Checkout the code
       - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-experimental.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+
```
peggles2, line 7: We can remove this extra line.

solution/backend/serverless-experimental.yml
```diff
@@ -48,7 +48,7 @@ provider:
     DEPLOY_NUMBER: ${env:RUN_ID}
     EUA_FEATUREFLAG: ${ssm:/eregulations/eua/featureflag}
     # Used for when you want to deploy your own
```
peggles2, line 50: is this comment needed anymore?

peggles2 (changes requested):

.github/workflows/deploy-experimental.yml
thwalker6: removed

solution/backend/serverless-experimental.yml
thwalker6: Yes its good to have a comment for what that specifici arn is for.

peggles2 (approved):

PR #1090, opened 2023-12-05
install findutils that installs xargs

peggles2:
Resolves #textract Dockerfile bug

**Description-**
Currently on prod textract dockerfile is failing with this error:
ERROR: failed to solve: process "/bin/sh -c xargs yum install -y < app/packages.txt" did not complete successfully: exit code: 127

Solution: Install findutils which installs xargs

- expected change 1
solution/text-extractor/Dockerfile

see if it deploys to dev.


thwalker6:
Why is this the solution?

PR #1111, opened 2023-12-15
EREGCSC-2416 - Secret cleanup test

KaushikSamtek:
Resolves #[EREGCSC-2416](https://jiraent.cms.gov/browse/EREGCSC-2416)

**Description-**
Just finishing off the secret cleanup effort by updating the Snyk Workflow to use the same JIRA variables as the SecHub Workflow so that there can be less Jira secrets in all the environments as well as in the general section.

the Snyk workflow to reflect the secret name changes as well as the Snyk ReadMe to do the same.

Verify in the Snyk workflow that the jira token name was changed from JIRA_TOKEN_FOR_SNYK to JIRA_TOKEN.

In the Readme, Verify the same name change for the token and also the HOST secret name was changed to JIRA_HOST. Also the Username references were removed as Username is no longer used.

### DIFF
--- a/.github/workflows/snyk-test.yml
+++ b/.github/workflows/snyk-test.yml
@@ -1,6 +1,8 @@
 name: Snyk Scan and Report
 
 on:
+  push:
+    branches: [ secretCleanupTest ]
   pull_request:
     branches: [ main ]
   schedule:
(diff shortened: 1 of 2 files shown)


peggles2 (approved):
LGTM

PR #1115, opened 2023-12-18
EREGCSC-2252-B Fix attachment extraction for zip, eml, and msg files

cgodwin1:
Resolves #2252

**Description-**

Text extractor was failing for some files with other files contained within them. This PR fixes it.

- Attachment files that fail to extract will not cause the entire extractor lambda to fail.
- Failures are instead logged to Cloudwatch. (using Python logging library)
- Added support for Outlook "msg" files that contain embedded "msg" data. Previously this would cause a failure because this embedded data lacks a filename.
- Lambda return message is also logged to Cloudwatch.

**Follow up work**

- We need more fixture files to find edge cases with these types of files. We can't use policy repo files since they may contain sensitive information including private email addresses etc. So we need someone with Outlook to create test files for us.
- Need to add support for RTF and HTML formatted "eml" files. These _should_ be uploaded currently, but they will not have a message body unless a plain text version is also embedded, which it often is (but not necessarily always!)

1. Verify unit tests pass
2. Try uploading a zip file with 1 supported file type (e.g. docx) and 1 unsupported file type (e.g. png), and verify the docx is still extracted.
3. Do the same with eml and msg, if feasible. You may find examples containing unsupported files in the policy repository, or contact me for help.
4. Find a msg file with embedded msg data and ensure both emails extract properly. Contact me if you need help finding a file to test with.

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -81,7 +81,7 @@ jobs:
       deployextractor: false
     runs-on: ubuntu-20.04
     # Change to true if you want to create text extractor for your experimental deploy.
-    if: false
+    if: true
     steps:
       # Checkout the code
       - name: Checkout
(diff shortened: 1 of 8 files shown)


peggles2 (approved):
LGTM 👍

PR #1120, opened 2023-12-27
EREGCSC-2481 Extract text from TXT files

cgodwin1:
Resolves #2481

**Description-**

This PR adds support for nearly any character encoding for text files. Previously, only UTF-8 would decode properly, which does not represent the majority of text files that users may upload.

- BeautifulSoup dependency added for their automatic encoding detection/extraction software. (Will be used for HTML/XML extraction too.)
- Raw binary is loaded and BeautifulSoup decodes it.
- Tests are updated to test UTF-8, UTF-16, Windows-1252, and ISO 8859-1 encodings. More can be added if edge cases are found. (Impractical to test _every_ possible text file encoding at this time.)

1. Verify unit tests pass.
2. Log into admin panel and go to Uploaded Files.
3. Upload a "txt" file to the panel.
4. Click "Get content" on the newly uploaded file.
5. Refresh the page periodically and wait for extraction to complete.
6. The extracted text should appear on the page under "Index populated".


peggles2 (approved):
LGTM 👍

PR #1122, opened 2023-12-27
EREGCSC-2465 Extract text from RTF files

cgodwin1:
Resolves #2465

**Description-**

We want to extract text from RTF files.

- Add striprtf library for extracting text from RTF files.
- Extract text with "errors=replace" in case of bad encoding/corruption.
- Unit tests updated, checks in case of corruption.

1. Verify unit tests pass.
2. Log into admin panel and go to Uploaded Files.
3. Upload a "rtf" file to the panel.
4. Click "Get content" on the newly uploaded file.
5. Refresh the page periodically and wait for extraction to complete.
6. The extracted text should appear on the page under "Index populated".


peggles2 (approved):
LGTM 👍 as long as checks clear.

PR #1124, opened 2024-01-02
EREGCSC-2345-B Fix 500 error during content search

cgodwin1:
Resolves #2345

**Description-**

The v3 content-search endpoint was failing with a 500 error if an uploaded file was assigned to a category. The cause of this was twofold:

1. Repository categories were not prefetched with `select_subclasses()`, causing the type to be `AbstractRepoCategory`.
2. When passed to the relevant polymorphic serializer, the use of `Serializer(object).data` caused a failure because the parent serializer was returning "Serializer not available" (due to `object` being an `AbstractRepoCategory`), which was not serializable as a _dict_. Instead, using `Serializer().to_representation(object)` allows for arbitrary data types.

- These issues are resolved.

1. Upload a file (with a name or description) to the admin panel of the experimental deploy.
2. Assign a category or sub-category to it.
3. Go to the content search page and search for the name or description you added.
4. Verify the results load properly.
5. Additionally, visit `https://9sqva39it4.execute-api.us-east-1.amazonaws.com/dev1124/v3/content-search/?q=X&location_details=true&category_details=true` and replace `X` with a word from your name/description to see the raw API results and verify that it looks appropriate.

### DIFF
--- a/solution/backend/cmcs_regulations/settings/deploy.py
+++ b/solution/backend/cmcs_regulations/settings/deploy.py
@@ -2,8 +2,6 @@ from .base import * # noqa
 import os
 import re
 
-DEBUG = True
-
 default_database_values = {
     'ENGINE': 'django.db.backends.postgresql',
     'USER': os.environ.get('DB_USER', 'eregs'),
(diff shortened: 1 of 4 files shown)


PhilR8 (approved):
Manually verified on experimental deployment.  LGTM!

PR #1131, opened 2024-01-10
EREGCSC-2465-B Fix rtf extraction failure and add tests

cgodwin1:
Resolves #2465

**Description-**

Extraction was failing for some files because unicode control characters were not being stripped from the output text. This is a valid string in Python, but when uploaded to eRegs it would cause a 500 error because some of these control characters can't be stored in the database (for example `\0`).

- Strips control characters from output text. (Done after text extraction itself is complete, but before uploading.)
- Wrote unit tests for this functionality and also config retrieval.
- Moved some function from `text_extractor.py` to `utils.py` to avoid import errors during unit testing.

1. Ensure unit tests pass.
2. Go to the JIRA ticket and download the 2 RTF files that failed to extract (listed in the ticket's comments).
3. Upload them to the experimental deploy's admin panel ("Uploaded files").
4. "Get Content" for both of these files, the click the back button.
5. Ensure that after some time, "Index populated" will show an extracted text excerpt.

### DIFF
--- a/solution/text-extractor/test_text_extractor.py
+++ b/solution/text-extractor/test_text_extractor.py
@@ -1,0 +1,49 @@
+import unittest
+import json
+
+from text_extractor import clean_output, get_config
+
+
+class TextExtractorTestCase(unittest.TestCase):
+    def test_clean_output(self):
+        text = "Hello   \n  \n\n world \0 \t\r \x00"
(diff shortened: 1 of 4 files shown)


PhilR8 (approved):
Manually validated per validation instructions.  LGTM!

PR #1133, opened 2024-01-11
EREGCSC-2267 Remove filename sanitization

cgodwin1:
Resolves #2267

**Description-**

We determined that filename sanitization is not necessary in the near term. Since the existing code was causing issues, we decided to remove it completely instead of improve it.

Factors:

- S3 allows arbitrary unicode filenames.
- Browsers on Windows clean invalid filenames automatically.
- We want to keep filenames as intact as possible.
- Django's ORM prevents SQL injection for us, so we don't need to sanitize user inputs for that purpose either.

- Filename sanitization is removed.
- Unit test for this feature is removed.
- Experimental text extractor disabled again.

1. Verify unit tests pass.
2. Upload files with strange filenames to eRegs (e.g. "hello<world.txt"). The file should upload with the name intact.
3. Download the file, the file should have the same name on Mac and Linux, and "<" should be stripped out on Windows.


peggles2 (approved):
LGTM 👍

PR #1141, opened 2024-01-19
EUA - test branch

peggles2:
This branch is to test anything with EUA. Push to this branch then test the eua here.

### DIFF
--- a/solution/backend/regulations/logout.py
+++ b/solution/backend/regulations/logout.py
@@ -1,10 +1,9 @@
 from django.conf import settings
 
-
 def eua_logout(request):
     id_token = request.session.get('oidc_id_token')
     # get the domain url from the request and add /login to the end
-    logout_redirect_url = request.build_absolute_uri('/') + settings.STAGE_ENV + '/login'
+    logout_redirect_url = request.build_absolute_uri('/') + settings.LOGIN_REDIRECT_URL
 
     # In the local environment where there is no STAGE_ENV, handle the possibility of //logout
     logout_redirect_url = logout_redirect_url.replace('//login', '/login').replace('/prod', '')
--- a/solution/backend/regulations/logout.py
+++ b/solution/backend/regulations/logout.py
@@ -4,7 +4,7 @@ from django.conf import settings
 def eua_logout(request):
     id_token = request.session.get('oidc_id_token')
     # get the domain url from the request and add /login to the end
-    logout_redirect_url = request.build_absolute_uri('/') + settings.STAGE_ENV + '/admin'
+    logout_redirect_url = request.build_absolute_uri('/') + settings.STAGE_ENV + '/logout'
 
     # In the local environment where there is no STAGE_ENV, handle the possibility of //logout
     logout_redirect_url = logout_redirect_url.replace('//login', '/login').replace('/prod', '')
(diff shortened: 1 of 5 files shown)


solution/ui/e2e/cypress/support/common-commands/login.js
```diff
@@ -4,10 +4,11 @@ export const eregsLogin = ({
...
     landingPage = "/policy-repository/",
 }) => {
-    cy.visit(landingPage);
+    cy.visit('/admin');
     cy.wait(1000);
     cy.get("#id_username").type(username);
     cy.get("#id_password").type(password);
     cy.get("#login-form").submit();
+    cy.visit('/policy-repository/');
 };
```
peggles2, line 14: since we no longer can visit /policy-repository/ and sign in with django admin, to get around it we will visit /admin and sign in that way but it would redirect to /admin so I updated this to then visit /policy-repository.

solution/ui/e2e/cypress/e2e/policy-repository.spec.cy.js
```diff
@@ -42,10 +42,10 @@ Cypress.Commands.add("getPolicyDocs", ({ username, password }) => {
 describe("Policy Repository", () => {
     beforeEach(_beforeEach);
 
-    it("shows the login screen when you visit /policy-repository/ without logging in", () => {
+    it("shows the custom eua login screen when you visit /policy-repository/ without logging in", () => {
```
PhilR8, line 45: if we're going to include all of the header functionality on the custom login page, perhaps we should test that the stuff works. 

You could copy and reuse the Error Page test suite (`error-page.cy.js`) and remove all the error page stuff and replace it with visiting the login page.  It should check for a11y compliance as well as making sure the flash banner, feedback modal and Jump To functionality works as expected.

solution/ui/e2e/cypress/support/common-commands/login.js
PhilR8: Line 12 should still be `cy.visit(landingPage);`.  If you look at like 5, `eregsLogin` takes an object as a parameter, and one of the fields in that object is `landingPage`.  We're providing `/policy-repository/` as a default value for `landingPage`.  So you can just visit `landingPage` if you want to go visit the policy repository.

solution/ui/regulations/css/scss/partials/_login.scss
```diff
@@ -0,0 +1,32 @@
...
+
+@import "../application_settings";
+
+main.login {
+    display: block;
+    margin: $spacer-3 auto $spacer-7;
+    max-width: $text-max-width;
+    max-height: 100%;
+    button.action-btn.default-btn {
+        margin-top:16px;
```
PhilR8, line 13: there needs to be a space after the colon, like so: `margin-top: 16px;`

PhilR8 (changes requested):
I added a few comments.

solution/ui/e2e/cypress/support/common-commands/login.js
peggles2: The issue is since we do not have a way to mock signing in with EUA it wont get redirected back to policy repository. But I see now we also can sign in from the left nav, which then means we should redirect back to the left nav. Best solution is to pass a parameter where it should go back for cypress to know. Since we are backdoor signing in.

PhilR8 (approved):
Changes look good!  LGTM!

PR #1142, opened 2024-01-19
EREGCSC-2494 Extract text from images

cgodwin1:
Resolves #2494

**Description-**

We want to extract text from images to ensure feature-completeness while extracting emails, zip files, and webpages. 

- AWS Textract functionality of PdfExtractor is moved to a generic TextractExtractor class which accepts jpeg, png, and tiff files.
   - These file types are selected because they are the types that Textract supports. Other types need to be converted to one of these to extract.
- PdfExtractor class only handles the conversion of pdf pages to images, then calls TextractExtractor.
- ImageExtractor handles all other image files that Textract doesn't natively support: gif (first frame only), jpeg 2000, bmp, tga, and webp to start as more common image file formats.
- Improved mocking of Textract for unit tests.
- Added the concept of "collections" to extractor unit tests. Collections permit handling multiple sample files which have the same expected output.

1. Verify unit tests pass.
2. In the experimental deploy's admin panel, visit "Uploaded files".
3. Upload several file types with embedded text, at least one each of: pdf, png, jpg, and bmp. This will test multiple extractor's functionalities.
4. For each file, click "Get content", then the back button.
5. Refresh the page until "Index populated" shows an excerpt of the text contained in the file.


peggles2 (approved):
LGTM 👍

PR #1149, opened 2024-01-25
EREGCSC-2502 Extract text from webpages

cgodwin1:
Resolves #2502

**Description-**

Enables proper extraction from HTML/HTM/XML/XHTML webpages contained in eRegs' "Resources" Django app.

- "Get content" on links will call the web backend on the text extractor.
- Text extractor supports parsing robots.txt.
- Text extractor will respect 429 responses and delay for 30 seconds or, if specified in the response headers, the `Retry-After` parameter.
- Text extractor will retry after timeouts.
- Retries are unlimited because Lambda imposes a maximum time limit of 15 minutes.

1. Run unit tests.
2. Go to the experimental deploy's admin panel.
3. Create a new "Supplemental Content" object that has a URL pointing to an arbitrary HTML page where robots.txt allows crawling. Click "Get content" and refresh until "Index populated" shows an excerpt.
4. Create a new "Supplemental Content" object that has a URL pointing to a page where robots.txt disallows crawling. Click "Get content" and observe that the index never populates.


PhilR8 (approved):
LGTM!

PR #1173, opened 2024-02-16
EREGCSC-2538 -- Replace Search.gov results with public `content-search` results on Combined Search

PhilR8:
Resolves [EREGCSC-2538](https://jiraent.cms.gov/browse/EREGCSC-2538)

**Description**

Removes Search.gov from front end of application.

As a result, the Resources column (right column) on the Search page will have identical results as the Subjects page.  Same number of results displayed the same way.

- No longer makes requests to `resources/search` or `resources/` on the Search page
- Instead, makes requests to `content-search`, just like the Subjects page
- Displays results on the Search page using same styles/components as they are displayed on the Subjects page using the `PolicyResults` component
- Removes almost all front end methods/tests/fixtures/artifacts related to Search.gov functionality
   - Notable exception: Schema Validation methods used with Cypress will remain.  These methods are used to validate the schema of responses from third party APIs, of which Search.gov results were the only thing being tested.  For now, we'll keep them in the codebase.
- Updates end to end tests to add coverage for new functionality

1. Visit [experimental deployment](https://5wumozc4z0.execute-api.us-east-1.amazonaws.com/dev1173)
2. Search for anything using the search box in the top right of the header (i.e. the normal way to get to the search page)
   - ex: [/search/q=MAT](https://5wumozc4z0.execute-api.us-east-1.amazonaws.com/dev1173/search/?q=MAT)
4. Ensure that the Resources column on the right works as expected
5. In another browser tab, search for the same query on the Subjects page and ensure that the results here look identical to the results on the Search page
   - ex: [/subjects/q=MAT](https://5wumozc4z0.execute-api.us-east-1.amazonaws.com/dev1173/subjects/?q=MAT)
6. Log in and ensure that both Public and Internal documents are being displayed on both the Search and Subjects page


solution/ui/e2e/cypress/e2e/api.spec.cy.js
```diff
@@ -85,18 +83,3 @@ describe("Synonyms endpoint special character testing", () => {
...
-// https://ahmed-alsaab.medium.com/cypress-api-schema-validation-with-ajv-ec61348374f7
-describe("Schema validation", () => {
-    describe("Search.gov Resources", () => {
-        it("gets successful response and expected schema from v3/resources/search", () => {
-            cy.getSearchGovResources(SEARCH_TERM).then((response) => {
-                expect(response.status).to.eq(200);
-                expect(response.isOkStatusCode).to.eq(true);
-                expect(response.statusText).to.eq("OK");
-
-                cy.validateSchema(getSearchGovResourcesSchema, response.body);
```
PhilR8, line 98: We no longer call out to any external APIs for data, so no more need to validate the schema of any API response.  We're still going to keep around the `validateSchema` method in case we need it later.

solution/ui/e2e/cypress/e2e/search.spec.cy.js
```diff
@@ -84,81 +89,12 @@ describe("Search flow", () => {
...
-        ).should("be.visible");
-        cy.get(
-            ".reg-results-content .reg-results-container .result:nth-child(1) .result__link"
-        ).should("be.visible");
-        cy.get(
-            ".reg-results-content .reg-results-container .result:nth-child(1) .result__link a"
-        ).should("have.attr", "href");
+    it("should have the correct labels for public and internal documents", () => {
+        cy.checkPolicyDocs({
+            username,
```
PhilR8, line 94: Since the Resources column (the right column) on the search page now requests and displays data the exact same way as the Subjects page, I've moved the assertions for those results to a utility method that can be called in both the Search and Subjects test suites.

solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
```diff
@@ -608,57 +608,11 @@ describe("Find by Subjects", () => {
     });
 
     it("should have the correct labels for public and internal documents", () => {
-        cy.intercept("**/v3/content-search/?q=test**", {
-            fixture: "policy-docs.json",
-        }).as("queriedFiles");
-        cy.viewport("macbook-15");
-        cy.eregsLogin({
+        cy.checkPolicyDocs({
             username,
```
PhilR8, line 612: Since the Resources column (the right column) on the Search page now requests and displays data the exact same way as the Subjects page, I've moved the assertions for those results to a utility method that can be called in both the Search and Subjects test suites.

solution/ui/e2e/cypress/schemas/example-schema.js
```diff
@@ -1,3 +1,5 @@
+// EXAMPLE SCHEMA
+
```
PhilR8, line 2: We're no longer using this file -- we're not validating any response schema anymore, since we only needed to do that for third party responses like Search.gov which we're no longer doing.  However, I can see us making calls to third party services in the future and I'd hate to have to search through the git logs to reproduce this work.  As such, I'm leaving it here for now as an example of how to create a schema to validate.

solution/ui/e2e/cypress/support/common-commands/policyDocs.js
```diff
@@ -0,0 +1,49 @@
+export const checkPolicyDocs = ({ username, password, landingPage }) => {
+    cy.intercept("**/v3/content-search/?q=test**", {
```
PhilR8, line 2: this is the utility method that will be used to verify that Internal and External policy docs are retrieved and displayed as expected.  This is currently used in the Search and Subjects test suites.

peggles2 (approved):
LGTM 👍

PR #1182, opened 2024-02-21
EREGCSC-2528 Extract text from webpages with no file extension

cgodwin1:
Resolves #2528

**Description-**

We want to extract text from files and web pages with no file extension: for example the type of the file "https://google.com" is HTML, but it has no ".html" extension so relying on extension will not work.

- Switch back to using MIME types for file type detection.
- Using Google's Magika ML library for file type detection instead of python-magic (https://github.com/google/magika)
- Fix a unit test error due to a new release of `moto`.
- Re-export image fixtures due to all of them accidentally being saved as PNG files with other extensions.

1. In the admin panel, go to "Uploaded Files" and upload a supported file. Then click "Get content" on that file, click back and refresh until "index populated" is populated. This makes sure regular file uploading still works.
2. Go to supplemental content and create new supplemental content. Add a URL pointing to a webpage that robots.txt allows (for example `http://example.com`). Then "Get content" and refresh on this until "Index populated" is populated.


solution/text-extractor/extractors/tests/fixtures/textract/sample.jpeg
cgodwin1, line 1: Image files were edited because somehow they were all saved as png files the first time.

peggles2 (approved):
LGTM 👍

PR #1184, opened 2024-02-26
EREGCSC-2520 -- Update navigation header to link to subject page

PhilR8:
Resolves [EREGCSC-2520](https://jiraent.cms.gov/browse/EREGCSC-2520)

**Description**

Remove Resources and About links from header; add Find by Subject link

- Updates URLs that are being passes into Vue Single Page Apps from Django templates
   - Removes `about` and `resources` URLs
   - Adds `subjects` URL
- Updates header to only include Statutes and Find by Subjects links
- Updates end to end tests

1. Visit [experimental deployment](https://4v3t3q94zf.execute-api.us-east-1.amazonaws.com/dev1184)
2. Notice that the Header only contains two links: 
   - Statutes
   - Find by Subject
1. Visit both of those links from several pages (reader view, Search page, Statutes page, About page, etc)


solution/ui/regulations/eregs-vite/src/views/Subjects.vue
```diff
@@ -351,8 +351,6 @@ getDocSubjects();
                 </template>
                 <template #links>
                     <HeaderLinks
-                        :about-url="aboutUrl"
-                        :resources-url="resourcesUrl"
```
peggles2, line 355: is subjectsUrl not needed here?

peggles2 (changes requested):
Looks good! Just had one question.

solution/ui/regulations/eregs-vite/src/views/Subjects.vue
PhilR8: Nope, not needed.  This is the Subjects page -- we don't need to link to it because we're already there.  It's the same pattern in the Statutes page: we pass in the subjects URL but not the statutes URL.

peggles2 (approved):
LGTM 👍

PR #1186, opened 2024-02-27
EREGCSC-2542 Don't extract extra large PDFs

cgodwin1:
Resolves #2542

**Description-**

We want to be able to limit the maximum size that the text extractor will extract, and offer an override functionality.

- The `Extractor` class now has an attribute, `max_size`, which is disabled (`-1`) by default. If set to a positive integer, the extractor will refuse to extract any file larger than `max_size` in megabytes.
    - This is configurable per-extractor because some extractors are slower than others.
    - Callers of the text extractor lambda may pass in `ignore_max_size: true` in the config dict to override this behavior.
    - Additionally, all extractor's `extract` methods are renamed to `_extract` to support the template method of function overriding. (`extract()` checks the size first, and then calls the child class's `_extract()`.)
- PdfExtractor's `max_size` is set to 20, per this PR's AC.
- Also, moved MIME type detection to the Extractor class as a classmethod to eliminate code duplication.
- Updated the README file.

1. Go to the admin panel and visit Uploaded Files.
2. Upload a small PDF (less than 20MB) and verify that the extractor will extract text successfully. (Click "Get content" -> go back -> refresh the page until "Index populated" is populated.)
3. Upload a large PDF (greater than 20MB) and verify that the extractor will NOT extract text anymore. (Same process as before, but this time, wait a long time to make sure "Index populated" is never populated.)

### DIFF
--- a/solution/text-extractor/README.md
+++ b/solution/text-extractor/README.md
@@ -164,7 +164,7 @@ from .extractor import Extractor
 class SampleExtractor(Extractor):
     file_types = ("filetype1", "filetype2", ...)
 
-    def extract(self, file: bytes) -> str:
+    def _extract(self, file: bytes) -> str:
         return ...extract text here...
 ```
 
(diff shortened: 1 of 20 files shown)


PhilR8 (approved):
LGTM!

PR #1195, opened 2024-03-05
[Hotfix] Increase prod lambda timeout to 20 seconds

cgodwin1:
Resolves #n/a

**Description-**

Content search is failing due to a timeout. This is a bandaid fix to hopefully avoid that.

- Prod's lambda timeout is increased to 20 seconds.

1. Verify that `/v3/content-search?q=test` no longer times out. You should refresh it several times to make sure.


PhilR8 (approved):
LGTM!

PR #1196, opened 2024-03-05
[Hotfix] Re-enable client side caching for `getCombinedContent` where used

PhilR8:
**Description**

`content-search` endpoint is consistently failing due to a timeout. This will re-enable client side caching for this endpoint wherever it is used.

The first request will still take ~10 seconds to recieve a response, but all subsequence requests for that endpoint with the same query parameters will be retrieved from the browser cache immediately with no waiting.

- Removes `cacheResponse: false` prop when calling `getCombinedContent`
   - the default parameter sets `cacheResponse` to `true`

1. Visit Subjects page or Search page on the [experimental deployment](https://6dsqjcks44.execute-api.us-east-1.amazonaws.com/dev1196).
2. Search for `test`
3. Measure the response time for the first request using the Network tab in DevTools
4. Reload the page and note that `content-search` no longer appears in the Network Tab's list of requests. 
5. This is because it has been served from the browser cache and should load immediately.

### DIFF
--- a/solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
+++ b/solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
@@ -7,6 +7,7 @@ const readerUsername = Cypress.env("READER_USERNAME");
 const readerPassword = Cypress.env("READER_PASSWORD");
 
 const _beforeEach = () => {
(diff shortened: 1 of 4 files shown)


PhilR8:
Further testing reveals that this is not an effective solution.  Closed.

PR #1201, opened 2024-03-07
EREGCSC-2349 - remove redirect serverless from experimental, update t…

peggles2:
…o the stage to be dev/val/prod

Resolves: Remove of deployment of redirect api/lambda function on experimental. 

**Description-**
redirect api is not needed on experimental. As this is to solve the problem on production so that when users sign in to regulations-pilot.cms.gov to be redirected to eregulations.cms.gov. It was added to experimental just so we can test the concept out. Now that its been approved and deployed, removing it. 
Also updated the api-gateway name to make it less redundent. 

- expected change 1

Visit https://github.com/Enterprise-CMCS/cmcs-eregulations/actions/runs/8191295614/job/22400196867?pr=1201 
Notice experimental did not deploy the redirect api.

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -156,7 +156,6 @@ jobs:
           pushd solution/backend
           npm install serverless -g
           npm install
-          serverless deploy --config ./serverless-redirect.yml --stage redirect${PR} | tee output.log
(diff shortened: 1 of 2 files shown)


cgodwin1 (approved):
LGTM!

PR #1203, opened 2024-03-08
EREGCSC-2552 Sort out scraping of FR docs

cgodwin1:
Resolves #2552

**Description-**

FR docs were failing to scrape because federalregister.gov is blocking scraping of their site, only allowing the usage of their API.

- FR parser now pulls "raw text url" from the FR API and includes that in the model.
- Bandaid for parser timeout issue: for now, "Get content" button fetches the raw text url and stores it in the model for old FR docs.
- FR docs instruct text extractor to ignore robots.txt for federalregister.gov, as it is also blocked. (We have permission to scrape the API, so robots.txt is not needed for this site.)
- Fixed bug in serializer causing FR parser to fail.

1. Log into the admin panel and go to a Federal Register Document that has "raw text URL" that is blank (located at the bottom of the page, for example: https://pn56tpa79f.execute-api.us-east-1.amazonaws.com/dev1203/admin/resources/federalregisterdocument/2643/change/)
2. Click "Get content", then back and refresh until "Index populated" is populated.

Note that some "raw text" urls are encoded as plain text but include _some_ HTML tags for reasons unknown. These are filtered out during search and shouldn't be of any concern.


PhilR8 (approved):
LGTM!

PR #1206, opened 2024-03-11
update code to redirect dynamically

peggles2:
Resolves # 

**Description-**

We want to redirect to eregulations but also the redirection should include the path. 
for example if we visit 
regulations-pilot.cms.gov/43/430 it should redirect to eregulations.cms.gov/42/430 not eregulations.cms.gov

expected change 1
- Update the lambda function to retain the path. 

1. Visit https://8t3teirzae.execute-api.us-east-1.amazonaws.com/dev1206/42/430
2. Notice it redirects to http://eregulations.cms.gov/dev1206/42/430
3. This does not impact our api gateway django code as it has its own api. So if you visit 
https://y0hrkjjdek.execute-api.us-east-1.amazonaws.com/dev1206/42/430/
it stays the same.

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -156,6 +156,7 @@ jobs:
           pushd solution/backend
           npm install serverless -g
           npm install
+          serverless deploy --config ./serverless-redirect.yml --stage dev${PR} | tee output.log
           serverless deploy --config ./serverless-experimental.yml --stage dev${PR} | tee output.log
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-experimental.yml
```diff
@@ -156,6 +156,7 @@ jobs:
           pushd solution/backend
           npm install serverless -g
           npm install
+          serverless deploy --config ./serverless-redirect.yml --stage dev${PR} | tee output.log
```
peggles2, line 159: Will remove this before merging to main.

cgodwin1 (approved):
LGTM!

PR #1209, opened 2024-03-14
EREGCSC-2561 Update Python to unblock Django upgrade

cgodwin1:
Resolves #2561

**Description-**

We are currently on Django 3.2 which supports up to Python 3.10. In order to unblock migrating to the latest versions of Django we can upgrade to Python 3.10 which unblocks up to Django 5.0.

Once we have upgraded Django, we can revisit upgrading again to an even newer version of Python.

- Update Python to 3.10 locally and on deployment.
- In one case, downgrade from 3.11 to 3.10 for consistency.

1. Green check marks on this deployment (unit tests good, no cypress errors, no seed data errors, no migration errors, etc.)
2. Visit the experimental deploy and click around to make sure the site is working as expected.
3. Create a new instance locally (rebuild, either with `docker system prune` or use `docker compose up -d --build regulations`).
4. Make sure the build completes successfully, and visit the site to make sure it's working as expected locally.
5. Run unit tests (`make test.pytest`).

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -40,7 +40,7 @@ jobs:
       - uses: actions/setup-python@v4
         if: success() && steps.findPr.outputs.number
         with:
-          python-version: "3.9"
+          python-version: "3.10"
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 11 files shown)


PhilR8 (approved):
Rebuilt locally and everything works as expected.  Poked around the experimental deployment and everything works as expected.  Python tests pass locally.  LGTM!

PR #1210, opened 2024-03-14
EREGCSC-2563 -- Omit admins from analytics

PhilR8:
Resolves [EREGCSC-2563](https://jiraent.cms.gov/browse/EREGCSC-2563)

**Description**

In order to create meaningful reports about user behavior, we don't want to count logged-in people with admin roles (job codes) as part of our Google Analytics.

- Creates `IS_ADMIN_USER` variable that can be used in all templates
- Adds logic to base template that will only render Google Tag Manager/Analytics `script` tags in the page `head` if `IS_ADMIN_USER` is `False`

1. Visit [experimental deployment](https://sascih35oc.execute-api.us-east-1.amazonaws.com/dev1210)
2. Ensure you are NOT logged in
3. Open DevTools and inspect the `head` element of any page.  
   - You should see the following comment: `<!-- Global site tag (gtag.js) - Google Analytics -->`
   - Underneath the comment, you should see two `script` tags.  One with a `src` that includes `googletagmanager.com`, and a second that contains some `gtag` logic
4.  Log in with an account that is not an admin
5. You should still see the two script tags described above
6. Log out and log in with an account that IS an admin
7. You should not see the GTag comment or the script tags in `head`

### DIFF
--- a/solution/backend/cmcs_regulations/context_processors.py
+++ b/solution/backend/cmcs_regulations/context_processors.py
@@ -8,6 +8,12 @@ def google_analytics(request):
     }
 
 
+def is_admin_user(request):
+    return {
+        "IS_ADMIN_USER": request.user.groups.filter(name='EREGS_ADMIN').exists()
+    }
+
+
 def custom_url(request):
     custom_url = settings.CUSTOM_URL
     host = request.get_host()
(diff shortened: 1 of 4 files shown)


peggles2 (approved):
LGTM 👍

PR #1222, opened 2024-03-27
[Hotfix] In test, add user search query that is guaranteed to succeed

PhilR8:
Resolves failing test on `dev` environment due to data being synced with `prod`

**Description**

We're syncing up database data for `dev`, `val`, and `prod` environments.  So far, we've taken a dump of `prod` data and used it to populated `dev`.  Now, `dev` behaves like `prod` -- which, in our case, means that certain search queries will now fail.

This is causing a test to fail on `dev`, which is blocking deployment to `prod`.

For future discussion: this failing test is one of the few tests that does not intercept a request and use fixtures for the results. Instead, it waits for the actual response from the server.

- Replaces the search query `test`, which fails, which search query `streamlined modular certification`, which passes.

1. This PR can be successfully deployed to `prod`.

### DIFF
--- a/solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
+++ b/solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
@@ -514,18 +514,18 @@ describe("Find by Subjects", () => {
     });
 
     it("should display and fetch the correct search query on load if it is included in URL", () => {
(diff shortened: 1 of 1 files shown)


peggles2 (approved):
LGTM 👍

PR #1226, opened 2024-04-02
EREGCSC-2598 Investigate speeding up SearchHeadline

cgodwin1:
Resolves #2598

**Description-**

Our search is extremely slow and timing out because Postgres' `ts_headline()` function can process text at a rate of about 4-5 MB/s, and some of the documents we are generating headlines for store multiple megabytes of text in `ContentIndex`'s `content` field. Furthermore, `ts_headline()` can't make use of indexing or vectors, so cannot be optimized that way.

After experimenting, we determined the best way moving forward is to force `ts_headline()` to only generate headlines for the first 50,000 (configurable) characters of the `content` field of `ContentIndex`. This should consistently produce query times of around 10 milliseconds per resource or less.

Combining this with efficient pagination will produce fast search results. Note that setting large page sizes or disabling pagination can still cause excessive query times because `ts_headline()` will run on _all_ results on the current page.

- Created environment variables for fragments, min/max words, max headline text size.
- Using `Substr`, annotated a field `content_short` that contains the first `N` characters of `content` where `N` is configured via environment variable.
- Instructed `SearchHeadline` to use `content_short` instead of `content` for generating headlines. Search still occurs on the `content` field, but its query time is negligible in comparison.

1. Green check marks and unit tests pass.
2. Go to the experimental deploy and go to `/subjects`. Enter a search term and make sure results come back. Note that results on the experimental deploy are _not_ loaded with extracted text. This test is to ensure search works and is not broken.
3. Approve and deploy this PR. On dev and val, where the searches are currently slow, make sure searches come back fast and reliably assuming pagination is enabled and `page_size` is a reasonable number (100 or less).


PhilR8 (approved):
LGTM!

PR #1231, opened 2024-04-03
[hotfix] EREGCSC-2598-B Paginate before generating headlines

cgodwin1:
Resolves #2598

**Description-**

New search code failed on deployment because it was still generating headlines for all results instead of only the current page. 

- Force pagination of the queryset _before_ generating headlines

1. Green check marks.
2. Make sure `/v3/content-search/?q=test` and `/v3/content-search/?q=test&paginate=false` both do not fail.
3. Check if searching for "test" works properly on dev and val before deploying to prod.


PhilR8 (approved):
LGTM!

PR #1232, opened 2024-04-04
add ssl restrictions for the cloudfront logs bucket

peggles2:
Resolves #

**Description-**

- expected change 1

### DIFF
--- a/solution/static-assets/serverless.yml
+++ b/solution/static-assets/serverless.yml
@@ -71,6 +71,7 @@ resources:
         VersioningConfiguration:
           Status: Enabled
         AccessControl: LogDeliveryWrite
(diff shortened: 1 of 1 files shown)


solution/static-assets/serverless.yml
```diff
@@ -116,10 +143,6 @@ resources:
                   - - 'arn:aws:s3:::'
                     - !Ref AssetsBucket
                     - /*
-                - !Join
-                  - ''
-                  - - 'arn:aws:s3:::'
-                    - !Ref AssetsBucket
```
peggles2, line 122: This was redundant. I don't see a reason why we have this here twice.

PR #1233, opened 2024-04-04
[hotfix] EREGCSC-2598-C Fix search query configuration while generating headlines

cgodwin1:
Resolves #2598

**Description-**

The `search_query` variable was being defined as two different things in search: first as a string containing the raw text of the query, and then as a Django `SearchQuery` object. This led to me forgetting to include it while generating headlines.

- Made a method in the ContentIndex queryset to create the `SearchQuery` object.
- Used a different name for this variable to avoid future confusion.
- SearchHeadline now uses this object instead of the raw string, as intended.

Test highlighting on a search query and ensure all instances are highlighted. [See here](https://4k5cpzavkb.execute-api.us-east-1.amazonaws.com/dev1233/v3/content-search/?html_api&q=affordable+care+act). Before, only "care" and "act" in the `document_name_headline` field of the first result were highlighted. Now all three words are.

### DIFF
--- a/solution/backend/content_search/models.py
+++ b/solution/backend/content_search/models.py
@@ -27,28 +27,29 @@ class ContentIndexQuerySet(models.QuerySet):
     max_fragments = int(settings.SEARCH_HEADLINE_MAX_FRAGMENTS) or None
 
     def search_configuration(self, query):
-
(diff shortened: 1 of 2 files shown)


PhilR8 (approved):
LGTM!

PR #1261, opened 2024-04-16
EREGCSC-2604 -- Remove unsupported browser message

PhilR8:
Resolves [EREGCSC-2604](https://jiraent.cms.gov/browse/EREGCSC-2604)

**Description**

When waiting for a page to load, people sometimes see a Firefox logo that covers much of the page. 

This happens because the CSS styles for the "browser not supported" message take a moment to load if a user is visiting the site for the first time or if the stylesheets are no longer cached in their browser.  Once the CSS stylesheets load, the message disappears.

This message was added in 2021 and was meant to only target Internet Explore 10 and 11 users.

Since Internet Explorer reached End of Life in June 2022, we can safely assume very few users are still using it.  As such, the simplest way to resolve this is to remove the "browser not supported" message.

- Removes Browser Not Supported Django template, removes its inclusion from the base template, and deletes the stylesheet associated with it.

1. Green check marks
2. Visiting the [experimental deployment](https://2k9j13c0q2.execute-api.us-east-1.amazonaws.com/dev1261) for the first time or after clearing cache will never show the giant Firefox logo
3. If you have access to Internet Explorer, visit the [experimental deployment](https://2k9j13c0q2.execute-api.us-east-1.amazonaws.com/dev1261) in IE 11 and ensure that the message no longer appears.


peggles2 (approved):
LGTM 👍

PR #1266, opened 2024-04-17
Vite Upgrade

PhilR8:
Resolves several [Dependabot issues](https://github.com/Enterprise-CMCS/cmcs-eregulations/security/dependabot)

**Description**

Vite's version was not defined specific enough to ensure the latest version is always installed.  This PR pegs the vite versions to the latest patch (5.2.9) in all places that vite is used.

- Updates vite version from `^5.0` to `^5.2.9` in all three places where it is used
- Fixes syntax error that is now enforced as a breaking error instead of an ignorable warning

1. Green check markes
2. Experimental Deployment works as expected
3. Site builds and can be watched locally as expected
4. Once merged to `main`, Vite-specific Dependabot alerts go away

### DIFF
--- a/solution/ui/regulations/eregs-component-lib/package-lock.json
+++ b/solution/ui/regulations/eregs-component-lib/package-lock.json
@@ -8,7 +8,7 @@
       "name": "eregs-component-lib",
       "version": "0.0.0",
       "dependencies": {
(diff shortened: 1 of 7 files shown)


peggles2 (approved):
LGTM 👍

PR #1275, opened 2024-04-25
EREGCSC-2560 -- Simplify reg page sidebars to prep for subjects

PhilR8:
Resolves [EREGCSC-2560](https://jiraent.cms.gov/browse/EREGCSC-2560)

**Description**

The right sidebar in the Reader View needs to be simplified so users can better see Public/Internal documents. 

Currently, these documents are being pushed down the page and below the screen by the "View Past Versions" and Print buttons at the top of the right sidebar.

We're not convinced that these buttons are useful or used at all.  As such, this PR will remove the "View Past Versions" and Print buttons so that the Public/Internal document categories have better visibility for users.

- Removes Print and View Past Versions buttons from right sidebar Django template
- Removes View Past Versions button template fragment
- Removes `PrintBtn` Vue component
- Removes related/unneeded styles
- Style tweaks based on AC
- Fixes print view style bugs/side effects that were introduced in the Revised Header story

1. Green check marks
2. Visit [Experimental Deployment](https://encc9nob0d.execute-api.us-east-1.amazonaws.com/dev1275) and go to the Reader View for any subpart/section
   - View Past Versions button should be gone
   - Print button should be gone
3. Press the keyboard shortcut for print view (CMD+P or Ctrl+P) to make sure Print view still exists and looks as expected
4. Visit a previous version of the same Section ([example](https://encc9nob0d.execute-api.us-east-1.amazonaws.com/dev1275/42/433/Subpart-A/2020-12-31/)\) to make sure the green banner at the top of the page still appears
   - We want links to previous versions of the regs to still work, communicate that this is a past version, and allow the user to return to the current version


cgodwin1 (approved):
LGTM!

PR #1277, opened 2024-04-25
EREGCSC-2555 Fix for "Potential race conditions in deploy sequence"

cgodwin1:
Resolves #2555

**Description-**

Sometimes, after a deployment, some people would see old static assets on the site and other people would see newly updated assets. This issue appeared to be geographic in nature as it would affect, for example, everyone on the west coast but not the east.

I believe the issue was that we were doing invalidation as part of our site deployment action and not our static-assets bucket deployment action. Why? I don't know. But the consequence is that the site would be deployed, an invalidation would be triggered, and then static-assets would be updated.

My thinking is that if someone accessed prod between site lambda deployment and static-assets deployment, it would rebuild the cache at whatever edge location they were closest to, but not necessarily for others. That rebuilt cache would contain the old static-assets since they hadn't fully deployed yet.

- Moved `serverless-cloudfront-invalidate` to `static-assets/serverless.yml` and out of the `backend` files.
- Updated static-assets' version of cloudfront-invalidate. Was pinned to 1.3 from the old days. We are choosing not yet to fix the vm2 vulnerability with cloudfront-invalidate. It it out of the scope of this ticket.
- Removed duplicate static-assets deployment in both `serverless.yml` and `serverless-experimental.yml`. Static-assets was being deployed before the reg site and then again after building Vue assets.

There's _probably_ no good way to be 100% certain of this except to merge and deploy to prod and see if the issue crops up again in the future. Sort of like this bugfix: https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1262 How can we truly validate something when we aren't totally certain about what's really causing it?

Otherwise, validate in the standard way:

1. Green check marks.
2. Code review.
3. When deployed, keep an eye on it before deploying to prod since the `deploy.yml` has changed.


brittag:
Noting that we had a "failing" Snyk vulnerability check on this one, and I chose to skip/ignore it because this PR does not actually introduce any new Snyk findings. We continue to use an outdated dependency that we’ve already assessed as a non-critical issue, and we have it in our backlog to fix ([EREGCSC-2478](https://jiraent.cms.gov/browse/EREGCSC-2478)).

PR #1278, opened 2024-04-29
EREGCSC-2602 If SearchHeadline can't generate a highlighted headline, show nothing

cgodwin1:
Resolves #2602

**Description-**

By default, `SearchHeadline` will return the beginning of the text of the field of interest if the search term is not contained within that field. Although this can be useful behavior, in our case we want to display nothing instead. This PR solves that by adding a boolean option to the `HeadlineField` field we created, `blank_when_no_highlight`. When set to `True` the field will check for the existence of the `<span>` in the headline results, and if it doesn't exist, it will set the resulting field to `None`.

- Added option `blank_when_no_highlight` to `HeadlineField`.
- `blank_when_no_highlight` is set to `True` for the `content_headline` field.
- Added a test for both conditions (when `blank_when_no_highlight` is true and false, 4 tests total).
- Fixed a local issue where we had multiple usernames for the database: `eregs` in some places and `eregsuser` in others. It is now `eregsuser` everywhere.
- Reverted changes to Github Actions deploy yml files. Turns out we were deploying static assets twice for a reason: a circular dependency between reg site and static assets. Vite needs to know Django's URL to build, and Django needs to know the static assets URL. So we were building static assets sans vite first, then deploying it to have a valid URL, then deploying Django, then building vite and deploying static assets again.

In the experimental deploy, I have created a sample uploaded file with the name "This is a test file", summary "Test file", and content of the text file is "Hello world".

1. Go to https://unkbr9td91.execute-api.us-east-1.amazonaws.com/dev1278/v3/content-search/?q=test%20file and you should see my test file show up with an empty "content_headline" field. Note that this file is at the bottom of the first page of results.
2. Go to https://unkbr9td91.execute-api.us-east-1.amazonaws.com/dev1278/search/?q=test+file and note that my file shows up at the bottom of the search results, showing no excerpt from the document's text.
3. Go to https://unkbr9td91.execute-api.us-east-1.amazonaws.com/dev1278/subjects?q=test+file and note that my file shows up at the bottom of the search results, showing no excerpt from the document's text.

### DIFF
--- a/.github/workflows/deploy-experimental.yml
+++ b/.github/workflows/deploy-experimental.yml
@@ -13,6 +13,67 @@ permissions:
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
+  deploy-static:
+    environment:
+      name: "dev"
+      url: ${{ steps.deploy-regulations-site-server.outputs.url }}
+    runs-on: ubuntu-22.04
+    steps:
+      # Checkout the code
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+      # Find the PR number.  This is not always trivial which is why this uses an existign action
+      - name: Find PR number
+        uses: jwalton/gh-find-current-pr@v1
+        id: findPr
+        with:
+          # Can be "open", "closed", or "all".  Defaults to "open".
+          state: open
+      # should build first and save the artifact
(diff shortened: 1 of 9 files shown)


peggles2 (approved):
LGTM 👍

PR #1287, opened 2024-05-08
EREGCSC-2667 -- Reloading a subject page causes it to have the wrong <title> value

PhilR8:
Resolves [EREGCSC-2667](https://jiraent.cms.gov/browse/EREGCSC-2667)

**Description**

Document title is sometimes incorrect on load when visiting the subject page with a subject included in the URL.

The issue: 

- Fixes bug where only the first digit of a subject ID was used on page load to set the Document Title (ex: 25 Dental would be interpreted as 2 ABP)
- Adds Cypress end to end test to validate this change
- Fixes issue with Subject Selector highlighting/bolding where multiple subjects could be bolded

1. Green check marks
2. Visit the [Subjects page on the experimental deployment](https://tuncmlts43.execute-api.us-east-1.amazonaws.com/dev1287/subjects/)
   - select and deselect a few subjects from both the Table of Contents (the right side of the page) and the Subject Selector (in the left sidebar) and make sure the Document Title updates correctly each time
   - Make sure the correct subject is highlighted/bolded in the Subject Selector in the left sidebar
3. Visit the Subjects page with a subject already in the URL. Example: [25 Dental](https://tuncmlts43.execute-api.us-east-1.amazonaws.com/dev1287/subjects/?subjects=25)
   - Ensure that the Document Title is correct (Dental)
   - Ensure that only Dental is highlighted/Bold in the Subject Selector in the left sidebar

### DIFF
--- a/solution/ui/e2e/cypress/e2e/title-tags.spec.cy.js
+++ b/solution/ui/e2e/cypress/e2e/title-tags.spec.cy.js
@@ -55,5 +55,17 @@ describe("Updated HTML Title Tags", { scrollBehavior: "center" }, () => {
             "eq",
             "Find by Subject | Medicaid & CHIP eRegulations"
         );
+
+        cy.visit("/subjects/?subjects=1");
+        cy.title().should(
+            "eq",
+            "Cures Act | Find by Subject | Medicaid & CHIP eRegulations"
+        );
+
+        cy.visit("/subjects/?subjects=18");
+        cy.title().should(
+            "eq",
+            "Collateral Contacts | Find by Subject | Medicaid & CHIP eRegulations"
+        );
     });
 });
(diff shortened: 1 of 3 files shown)


solution/ui/e2e/cypress/e2e/title-tags.spec.cy.js
```diff
@@ -55,5 +55,17 @@ describe("Updated HTML Title Tags", { scrollBehavior: "center" }, () => {
             "eq",
             "Find by Subject | Medicaid & CHIP eRegulations"
         );
+
+        cy.visit("/subjects/?subjects=1");
+        cy.title().should(
+            "eq",
+            "Cures Act | Find by Subject | Medicaid & CHIP eRegulations"
+        );
```
PhilR8, line 63: Note: we are using a fixture file (subjects.json) to mock the response to `/v3/file-manager/subjects`.  So these category IDs (1 Cures Act, 18 Collateral Contacts) will not be affected by any changes to the actual subject names and IDs; 1 will stay Cures Act and 18 will stay Collateral Contacts until we update the fixture file.

cgodwin1 (approved):
LGTM!

PR #1304, opened 2024-05-17
EREGCSC-2648 - Add department as field for user

peggles2:
Resolves #EREGCSC-2648

**Description-**
We want people to be able to log into eRegs and automatically see the right items according to their group/division membership.

To prepare to automate the association of a group/division with a user, we need to associate department info with a user.

Updates the code to have a User app that contains a Profile model. 
(In the future we will also have Group and Division model in this app as well)
Create migration files to have a Profile that has the department field for now, and it belongs to AbstractUser. 
Update Admin panel so that it displays the department
Added tests to test everything out. 
Screen Shot 2024-05-22 at 9 50 42 AM

1. Sign into the admin panel.
2. click on [users](https://tktbrw35k1.execute-api.us-east-1.amazonaws.com/dev1304/admin/auth/user/)
3. select your user and see if your department has been filled in. 
4. Click on the user. Notice the department is read-only


cgodwin1 (approved):
LGTM!

PR #1309, opened 2024-05-22
[Hotfix] Remove `resources/search` endpoint from Cypress API test suite

PhilR8:
Resolves issue where our deployments depend on Search.gov having 100% uptime

**Description**

In order to return a successful response, one of our unused and soon-to-be-deleted endpoints, `v3/resources/search`, needs a valid response from Search.gov.

Unfortunately, right now Search.gov is having some issues and requests are timing out. Our `v3/resources/search` endpoint is not equipped to gracefully handle a Search.gov timeout, and as a result this endpoint is returning a 500 error that is causing a Cypress test to fail.  This failing test is preventing us from deploying anything.

- Removes test for `v3/resources/search` endpoint

1. Green check marks
2. Successful deployments all the way to `prod`


cgodwin1 (approved):
LGTM!

PR #1323, opened 2024-05-31
EREGCSC-2641 Delete the file-manager app

cgodwin1:
Resolves #2641

**Description-**

As part of the V4 refactor, we have copied all the data over from file-manager to the new resource models: `InternalFile`, `Subject`, etc. This PR is to delete the old file-manager models and code.

Currently, the app itself remains but only contains the migrations to avoid dependency issues while the refactor is completed. Once all refactor code is fully deployed to prod, we should be able to squash the migrations in `resources` and `content_search`, remove any references to `file_manager` in the resulting squashed migrations, and then fully delete all remnants of `file_manager`. At that time, `file_manager` will be fully merged with `resources`.

- All models, views, serializers, urls, admin panel views for `file_manager` are deleted.
- All old v3 `resources` models are deleted.
- The search view is temporarily disabled while this transition is in progress.
- New `v4` models are renamed via migration to their final naming scheme. Note the complex (but documented) migration in resources, `0040_rename_v4_models.py`, which performs this task.
- Modified v4 resources refactor to increase the local statement timeout.
- Modified `migrate.py` to call just "migrate". This was [originally changed in this commit](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/194/commits/8c143bdb4718bc0274750035f68ff92362ce3913) due to renaming an app, but is no longer needed and is causing dependency issues.

Experimental deploy is here: https://9ibb466jw3.execute-api.us-east-1.amazonaws.com/dev1323/admin/
No comment is created by GitHub for this PR because of failing tests. This is normal, see below.

1. Verify that the admin panel still works even with the bulk of `file_manager` code deleted.
2. No `file_manager` models show in the admin panel anymore.
3. Verify all migrations succeeded (full deploy will fail either way, but the `deploy-django` step will fail if migrations fail.)

Please note that if approved, this branch will be merged into `2639-refactor-resources-v4`, **_not_** `main`. At this stage it is non-critical, and in fact likely impossible, to get full green check marks for this experimental deploy. This is not a cause for concern at this point in time and should not be used as a factor in approval decisions.


PhilR8 (approved):
LGTM!

PR #1325, opened 2024-06-04
[Hotfix] EREGCSC-2352 -- Fix issue with subjects not always being an array (again); page numbers

PhilR8:
Resolves [EREGCSC-2352](https://jiraent.cms.gov/browse/EREGCSC-2352)

**Description-**

- expected change 1

### DIFF
--- a/solution/ui/regulations/utilities/filters.js
+++ b/solution/ui/regulations/utilities/filters.js
@@ -60,7 +60,7 @@ const locationUrl = ({ title, type, part, section_id, subpart_id }, base) => {
  * getSubjectName({ short_name: "Federal Regulations", abbreviation: "CFR", full_name: "Code of Federal Regulations" }) // "Federal Regulations"
  */
 const getSubjectName = (subject) =>
-    subject.short_name || subject.abbreviation || subject.full_name;
+    subject?.short_name || subject?.abbreviation || subject?.full_name;
 
 /**
  * @param subject {Object} - a subject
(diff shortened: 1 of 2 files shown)


solution/ui/regulations/utilities/filters.js
```diff
@@ -60,7 +60,7 @@ const locationUrl = ({ title, type, part, section_id, subpart_id }, base) => {
  * getSubjectName({ short_name: "Federal Regulations", abbreviation: "CFR", full_name: "Code of Federal Regulations" }) // "Federal Regulations"
  */
 const getSubjectName = (subject) =>
-    subject.short_name || subject.abbreviation || subject.full_name;
+    subject?.short_name || subject?.abbreviation || subject?.full_name;
```
PhilR8, line 63: Don't crash site if `subject` is undefined, which it can be sometimes.

solution/ui/regulations/eregs-vite/src/components/dropdowns/Categories.vue
```diff
@@ -45,8 +45,10 @@ onBeforeMount(() => {
     const { categories, intcategories } = $route.query;
 
     if (categories) {
+        silentReset.value = true;
         selectedId.value = `${categories}-categories`;
     } else if (intcategories) {
+        silentReset.value = true;
         selectedId.value = `${intcategories}-intcategories`;
```
PhilR8, line 52: `selectedId` is being `watch`ed later in this file; changing it triggers a route update UNLESS you set `silentReset` to `true`.  When we have a category in the URL on load, we need to set the Categories dropdown to have that value selected; this was triggering a route update, which was scrubbing `page` from the route if it existed.  So here we set `silentReset` to `true` so URL params aren't scrubbed out of the route on load.

peggles2 (approved):
LGTM

PR #1329, opened 2024-06-06
EREGCSC-2711 -- Update Vue Router navigation guard to ensure query params are not an array (except `type`)

PhilR8:
Resolves [EREGCSC-2711](https://jiraent.cms.gov/browse/EREGCSC-2711)

**Description**

We've had some issues where some query parameters from the Vue Router `route` are arrays sometimes and strings other times.  Specifically, this has been happening to `subjects`.

The reason for this issue turns out to be: we are explicitly setting `subjects` as an array when we programmatically set it via `router-link` or `$router.push`.  Whoops!

This was likely left over from the initial implementation of the Subjects page, where we briefly allows users to select more than one subject to view results from all subjects.  This was quickly changed to the current implementation, which only allows one subject to be selected at a time.

As such, we've removed the instances where we are programmatically setting `subjects` as an array -- it is now set as a string once again.

However.

If you have more than one query parameter in the URL -- `?subjects=1&subjects=2`, for example -- then Vue Router will still return `subjects` as an array (`subjects: ["1", "2"]`).  So we still need to handle the possibility that a user may somehow edit the URL and add another query parameter to one that already exists, if only to make sure the site doesn't keep breaking.

In this case, we're just going to return the first item in the resulting array -- so `subjects: ["1", "2"]` will be changed to to `subjects: "1"`.  The URL will still contain the multiple `subjects` query params but the site will load assuming that only the first one matters.  This is an extreme edge case that would be the result of the user mucking around in the URL, but it's a low lift to guard against this unlikely scenario.

So, now `subjects` and all query params that are not `type` (which is used in the Document Type Selector to filter by Internal or External documents) can safely be assumed to be a string.  (`type` will probably get refactors a tiny bit in a coming story so it may also join the other query params in being consistent.)

- When programmatically setting the `subjects` query param via `router-link` or `$router.push`, do not wrap the value in an array
- In the `router.beforeEach` navigation guard, loop through query params and ensure they are not arrays.  If they are arrays, only return the first item in that array.
   - except `type`, as noted above
- Clean up unused imports
- Fix `removeEventListener` to remove the correct event listener

1. Green check marks
2. Site works as expected on [experimental deployment](https://dd3cd7n5ue.execute-api.us-east-1.amazonaws.com/dev1329)
3. Visit a complex filter state and make sure it loads properly ([example](https://dd3cd7n5ue.execute-api.us-east-1.amazonaws.com/dev1329/subjects/?categories=11&page=2&q=long-term))
4. No errors in the console

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/components/dropdowns/Categories.vue
+++ b/solution/ui/regulations/eregs-vite/src/components/dropdowns/Categories.vue
@@ -152,7 +152,7 @@ onMounted(() => {
     window.addEventListener("popstate", onPopState);
 });
 
-onUnmounted(() => window.removeEventListener("resize", onPopState));
+onUnmounted(() => window.removeEventListener("popstate", onPopState));
 </script>
 
 <template>
--- a/solution/ui/regulations/eregs-vite/src/components/subjects/SubjectChip.vue
+++ b/solution/ui/regulations/eregs-vite/src/components/subjects/SubjectChip.vue
@@ -38,7 +38,7 @@ defineProps({
         :to="{
             name: 'subjects',
             query: {
-                subjects: [subjectId.toString()],
+                subjects: subjectId.toString(),
                 type: ['all'],
             },
             params: { subjectName },
--- a/solution/ui/regulations/eregs-vite/src/main.js
+++ b/solution/ui/regulations/eregs-vite/src/main.js
@@ -6,6 +6,8 @@ import vuetify from "./plugins/vuetify";
 import App from "./App.vue";
 import vueRouter from "./router";
 
+import _isArray from "lodash/isArray";
+
 const mountEl = document.querySelector("#vite-app");
 const { customUrl, host } = mountEl.dataset;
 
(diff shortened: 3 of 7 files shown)


solution/ui/regulations/eregs-vite/src/components/dropdowns/Categories.vue
```diff
@@ -152,7 +152,7 @@ onMounted(() => {
     window.addEventListener("popstate", onPopState);
 });
 
-onUnmounted(() => window.removeEventListener("resize", onPopState));
+onUnmounted(() => window.removeEventListener("popstate", onPopState));
```
PhilR8, line 155: this is a bug from the dropdown story -- we were removing the wrong event listener, and not removing the one we needed to remove in order to properly clean up after ourselves

solution/ui/regulations/eregs-vite/src/components/subjects/SubjectChip.vue
```diff
@@ -38,7 +38,7 @@ defineProps({
         :to="{
             name: 'subjects',
             query: {
-                subjects: [subjectId.toString()],
+                subjects: subjectId.toString(),
```
PhilR8, line 41: I had forgotten that we were explicitly setting the subjects query param as an array.  This is left over from when we were going to allow multiple subjects to be selected at the same time.

solution/ui/regulations/eregs-vite/src/main.js
```diff
@@ -28,8 +28,8 @@ router.beforeEach((to) => {
         }
 
         if (!isAuthenticated && to.query?.type) {
-            const { type, ...query } = to.query;
-            return { name: "subjects", query };
+            const { type, ...typelessQuery } = to.query;
+            return { name: "subjects", query: typelessQuery };
```
PhilR8, line 32: a little more verbose/explicit about what we're doing here

cgodwin1 (approved):

PR #1330, opened 2024-06-06
EREGCSC-2709 - Turn off frontend caching for resources

peggles2:
Turn off frontend caching for resources
Resolves #2709

**Description-**
Modified the code to disable frontend caching by introducing a constant, DEFAULT_CACHE_RESPONSE, set to false. This approach centralizes the caching configuration, allowing us to easily adjust the default caching behavior in one place in the future without needing to update every instance in the code.

Modifies the api.js file to set the default cache response to false.

**Steps to view and verify change**
update a resource item in the admin panel, 
Notice the changes right away on the front end of the site, even if previously loaded that page recently

### DIFF
--- a/solution/ui/regulations/composables/fetch.js
+++ b/solution/ui/regulations/composables/fetch.js
@@ -3,7 +3,7 @@ import { ref } from "vue";
 export default function useFetch({
     method,
     apiUrl,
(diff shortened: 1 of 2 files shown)


PhilR8 (approved):
LGTM!

PR #1340, opened 2024-06-12
EREGCSC-2713 -- Encode hash marks and other characters so `content-search` endpoint returns successfully

PhilR8:
Resolves [EREGCSC-2713](https://jiraent.cms.gov/browse/EREGCSC-2713)

**Description**

If you search for something with a # in the query (from the combined search page or from the subjects page), you get a few unexpected effects:

- In the results, the “related regulation citations” show as “none”
- In the results, the category and subcategory labels are missing ("Subregulatory Guidance" etc)
- Anything after # in the query seems to be ignored
- If you're on the subjects page and run this kind of query, the filters don't work as expected. For example, if I uncheck "public", public items are still displayed. 

- Encodes `q` query parameter before it is included `content-search` API call
   - We're creating the `content-search` query string two different wants in Search and Subjects pages, so we are encoding `q` in two different places
   - This will be addressed in a future story (2657) -- Search page will be refactored to share most of Subjects page logic, which will reduce the number of times we need to encode `q` to one.
- Adds unit test

1. Green check marks
2. Visit [experimental deployment](https://g065mi4m29.execute-api.us-east-1.amazonaws.com/dev1340/) and start adding special characters to search page and subjects page queries
   - Special characters shouldn't break site and they should be encoded in the query string parameters of the API call
   - Example: [SMDL #12-002](https://g065mi4m29.execute-api.us-east-1.amazonaws.com/dev1340/search/?q=SMDL+%2312-002) (Search)
   - Example: [SMDL #12-002](https://g065mi4m29.execute-api.us-east-1.amazonaws.com/dev1340/subjects?q=SMDL+%2312-002) (Subjects)

### DIFF
--- a/solution/ui/regulations/eregs-vite/src/views/Search.vue
+++ b/solution/ui/regulations/eregs-vite/src/views/Search.vue
@@ -455,16 +455,24 @@ export default {
                 return;
             }
 
+            const encodedQuery = encodeURIComponent(query);
+
             this.regsLoading = true;
             this.resourcesLoading = true;
 
-            this.retrieveResourcesResults({ query, page, pageSize }).then(
-                () => {
-                    this.resourcesLoading = false;
-                }
-            );
+            this.retrieveResourcesResults({
+                query: encodedQuery,
+                page,
+                pageSize,
+            }).then(() => {
+                this.resourcesLoading = false;
+            });
 
-            this.retrieveRegResults({ query, page, pageSize }).then(() => {
+            this.retrieveRegResults({
+                query,
+                page,
+                pageSize,
+            }).then(() => {
                 this.regsLoading = false;
             });
         },
--- a/solution/ui/regulations/utilities/utils.js
+++ b/solution/ui/regulations/utilities/utils.js
@@ -1065,6 +1065,7 @@ export {
     isFloat,
     mapToArray,
     niceDate,
+    PARAM_ENCODE_DICT,
     PARAM_VALIDATION_DICT,
     parseError,
     removeFragmentParams,
--- a/solution/ui/regulations/utilities/utils.js
+++ b/solution/ui/regulations/utilities/utils.js
@@ -75,6 +75,21 @@ const PARAM_VALIDATION_DICT = {
 };
 
 /**
+ * Dictionary of query parameters to encode before sending to the API.
+ *
+ * * @type {Object}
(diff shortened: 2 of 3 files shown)


solution/ui/regulations/eregs-vite/src/main.js
```diff
@@ -31,6 +31,11 @@ router.beforeEach((to) => {
         if (_isArray(value) && key != "type") {
             to.query[key] = value[0];
         }
+
+        // Encode the `q` query param to handle special characters in API calls etc
+        if (key == "q") {
+            to.query[key] = encodeURIComponent(value);
```
PhilR8, line 37: this keeps the URL clean and encoded as normal, i.e. spaces are still `+` marks.  This will help existing links to still work as expected.

solution/ui/regulations/utilities/utils.js
```diff
@@ -146,7 +161,12 @@ const getRequestParams = (query) => {
             );
 
             return filteredValues
-                .map((v) => `${PARAM_MAP[key]}=${v}`)
+                .map(
+                    (v) =>
+                        `${PARAM_MAP[key]}=${
+                            PARAM_ENCODE_DICT[key] ? encodeURIComponent(v) : v
```
PhilR8, line 167: encode query string when assembling query params to be sent to `content-search` endpoint

solution/ui/regulations/eregs-vite/src/views/Search.vue
```diff
@@ -455,16 +455,24 @@ export default {
...
+            }).then(() => {
+                this.resourcesLoading = false;
+            });
 
-            this.retrieveRegResults({ query, page, pageSize }).then(() => {
+            this.retrieveRegResults({
+                query,
+                page,
+                pageSize,
+            }).then(() => {
```
PhilR8, line 475: prettier formatting

solution/ui/regulations/utilities/utils.js
PhilR8: This is how we are creating the querystring for the `content-search` request on the Subjects page, which is a little different than how we do it on the Search page; however, the Search page is set to be refactored very soon and it will start using this same utility method then.

cgodwin1 (approved):
LGTM!

PR #1341, opened 2024-06-13
EREGCSC-2679 -- Shift internal/public filters into center area

PhilR8:
Resolves [EREGCSC-2679](https://jiraent.cms.gov/browse/EREGCSC-2679)

**Description**

When a researcher is looking at items on subject pages, they may want to filter by choosing whether to see internal documents, public documents, or all documents.

We currently have these filters in the sidebar, but we want to move it to the center - including so that they can be reused for the future combined search page.

- Repositions and restyles Document Type selector per [Figma doc](https://www.figma.com/design/j3ttFC9FtexKXBrV4yb3m7/Search-Everything?node-id=1-50&m=dev)
- Changes Document Type checkbox behavior to match story AC
   - Checkboxes are not visible on page load, only when a subject is selected or text search is executed
   - Checkboxes default state is unchecked
   - Selecting a category will select the related Document Type checkbox
      - Ex: selecting an Public doc category like "Related Statutes" will check the "Public Resources" checkbox and uncheck the "Internal Resources" checkbox if checked
- Tweak/fix popstate/forward/back behavior
- Updates Cypress tests

1. Green check marks
2. Visit [experimental deployment](https://6dhoro4jia.execute-api.us-east-1.amazonaws.com/dev1341)
3. Select a subject and a category without logging in
4. Ensure the document type selector checkboxes are never visible if you aren't logged in, even if you follow a [link with a `type` parameter in the URL](https://6dhoro4jia.execute-api.us-east-1.amazonaws.com/dev1341/subjects?categories=5&type=external&subjects=2)
5. Log in
6. Ensure that document type selector becomes visible in the top center of the page once you select a subject
7. Click around, change document types and categories and subjects etc and make sure everything behaves as expected
8. Use the back and forward buttons and make sure you can step back and forward through your session with everything loading and displaying as expected
9. When you select a category from the drop down, ensure that the appropriate checkbox is checked

### DIFF
--- a/solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
+++ b/solution/ui/e2e/cypress/e2e/subjects.spec.cy.js
@@ -136,7 +136,7 @@ describe("Find by Subjects", () => {
         cy.get(".subj-toc__list li[data-testid=subject-toc-li-63] a")
             .should("have.text", "Managed Care")
             .click({ force: true });
-        cy.url().should("include", "/subjects?subjects=63");
+        cy.get(".doc-type__toggle fieldset").should("not.exist");
         cy.get(".subject__heading")
             .should("exist")
             .and("have.text", "Managed Care");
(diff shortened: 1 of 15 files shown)


peggles2 (approved):
LGTM 👍

PR #1351, opened 2024-06-24
EREGCSC-2701 -- V4 frontend cleanup

PhilR8:
Resolves [EREGCSC-2701](https://jiraent.cms.gov/browse/EREGCSC-2701)

**Description**

Make necessary changes to the front end and associated end to end tests so that the site works as expected.

- Provide proper parameters to API endpoints
- Handle updated API response
- Updates Cypress tests and fixtures

** This pull request does NOT change:**

- Vitest unit tests (will be addresses in follow-up ticket)
- JSDoc comments and type definitions (will be addresses in follow-up ticket)

1. Cypress tests pass in `deploy-experimental` GitHub Action
2. Visit [experimental deployment](https://typvl84r9b.execute-api.us-east-1.amazonaws.com/dev1351/)
3. Validate that site works as expected without any regressions


.github/workflows/ruff-action.yml
```diff
@@ -4,8 +4,8 @@ jobs:
   ruff:
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: chartboost/ruff-action@v1
         with:
           src: "./solution"
-          args: --preview
```
PhilR8, line 11: this issue has already been fixed on `main`

.github/workflows/deploy-experimental.yml
```diff
@@ -351,7 +351,8 @@ jobs:
     environment:
       name: "dev"
     runs-on: ubuntu-22.04
-    needs: [deploy-go, deploy-django, build-and-deploy-vue]
+    needs: [deploy-django, build-and-deploy-vue]
+    #needs: [deploy-go, deploy-django, build-and-deploy-vue]
```
PhilR8, line 355: This is a temporary change until `deploy-go` passes.  This will be removed before merging to `main`.

solution/ui/regulations/eregs-vite/src/components/subjects/PolicyResults.vue
```diff
@@ -27,33 +27,71 @@ const addSurroundingEllipses = (str) => {
     return str;
 };
 
+const getFieldVal = ({ item, fieldName }) => {
+    if (item?.resource) {
+        return item.resource[fieldName];
+    } else {
+        return item[fieldName];
+    }
+};
```
PhilR8, line 36: This method is necessary (for now) because we are using two endpoints to retrieve data on the Subjects page: 
- `content-search` is there is a search query (`q`)
- `resources` / `resources/public` / `resources/internal` (depending on document type selections

These two endpoints return results in slightly different ways; `content-search` has a `resource` field that contains fields common to different document/file/link types.  The `resources` endpoint does not nest these common values in a `resource` field; they're just at the top level of the result item object.  Since we need to consume responses from both endpoints, we have to use this `getFieldVal` helper method to extract the proper value no matter which endpoint is used.

solution/ui/regulations/eregs-vite/src/views/Subjects.vue
```diff
@@ -200,18 +210,30 @@ const policyDocList = ref({
...
+                requestParams,
+                docType,
+            });
+        } else {
+            contentList = await getContentWithoutQuery({
+                apiUrl: props.apiUrl,
+                requestParams,
+                docType,
+            });
+        }
```
PhilR8, line 235: Since `content-search` requires a query string in order to return results, we have to conditionally use a different endpoint if no query string is present (`/resources` if no doc type is specified; `/resources/public` if only Public documents are requested; `resources/internal` if only Internal documents are requested

Ex: user searches for "FMAP" in category "3".  A request is made to `/content-search`
Ex: user searches for category "3" only.  A request is made to `/resources`

solution/ui/regulations/eregs-vite/src/views/Subjects.vue
```diff
@@ -496,7 +525,10 @@ getDocSubjects();
                         </div>
                         <div class="subject__filters--row">
                             <DocumentTypeSelector v-if="isAuthenticated" />
-                            <FetchCategoriesContainer v-slot="slotProps">
+                            <FetchCategoriesContainer
+                                v-slot="slotProps"
+                                :categories-capture-function="setCategories"
```
PhilR8, line 530: If a `content-search` response item is assigned to a subcategory, the included subcategory information (`id`, `name`, etc) no longer includes the parent category name.  It now only includes the parent category's id in the `parent` field.

Ex: `"parent": 4,`"

However, we need the name of this parent category to be displayed in the UI alongside the subcategory name.  As such, we need to fetch the categories ourselves from the API in order to use the `parent` id to find the parent's `name`.

We're already getting categories in this `FetchCategoriesContainer` component to be used in the categories dropdown menu.  We can use these fetched categories instead of making a second, separate API call; we just have to pass in a closure down to `FetchCategoriesContainer` that will update the parent scope when the child fetches the categories.

Admittedly, passing a function as a prop is considered an anti-pattern in Vue, but for this specific instance it can be tolerated.  We may also update the backend to return the parent category name instead of the id, but for now we must accept this implementation.

cgodwin1 (approved):
LGTM!

PR #1359, opened 2024-07-02
[Security Hub] Remove vue-loader from project

PhilR8:
Resolves [Dependabot Alert #96](https://github.com/Enterprise-CMCS/cmcs-eregulations/security/dependabot/96)

**Description**

The [vue-loader](https://vue-loader.vuejs.org/) npm package is a major version out of date and is being flagged as a security risk by Dependabot.  This security risk needs to be mitigated by either updating or removing vue-loader.  Since vue-loader is no longer used or necessary, it will be removed.

**What even is Vue Loader?**

It is described as a "[webpack loader for Vue Single-File Components](https://github.com/vuejs/vue-loader)".  However, we don't use webpack in this project -- we use Vite, a different tool that does similar things (JavaScript module bundling).

**Why was Vue Loader explicitly added to the project, then?**

Vue Loader has been explicitly defined as a project dependency for three years now, back to before this repository existed.  My guess is that it was required to load Vue components into Storybook, since [Webpack is the default builder for Storybook](https://storybook.js.org/docs/builders/webpack).

We've stopped using Storybook and removed it from the project in a previous story that was meant to remove Storybook and all of its dependencies.  It's likely that Vue Loader should have also been removed as part of that story but wasn't.

- Removes `vue-loader` from the `package.json` file where it was defined
- Updates related `package-lock.json` file

1. Green check marks
2. [Experimental deployment](https://c8cxmgka7e.execute-api.us-east-1.amazonaws.com/dev1359) works as expected


peggles2 (approved):
LGTM

PR #1381, opened 2024-07-23
EREGCSC-2707 Remove the file-manager app

cgodwin1:
Resolves #2070

**Important before deploying!**

Before deploying to prod, make sure a backup of prod is on hand. This should be a fairly safe migration as long as the steps for verifying below are followed, but this does overwrite our previous migration history so it is more risky than a normal migration. Extended validation is warranted here.

**Description-**

This PR is another step in completing the v4 API refactor. In the previous batch of work, we removed database and code dependencies on the file-manager app. However, we couldn't remove the app entirely because we needed the migration that removes file-manager's models to run, and resources/content-search both had migrations that depended on file-manager's migrations.

- Removed file-manager completely (file-manager directory, and `INSTALLED_APPS`).
- Manually squashed migrations for resources and content-search.
    - This is the easiest way to remove migration-based circular dependencies that have been resolved in code since the v4 refactor.
    - Automatic squashing via `squashmigrations` wasn't feasible because the resulting migrations had circular dependency errors and were overly complex.
    - By deleting all migrations and running `makemigrations`, we are able to recreate the current state of the database perfectly. Then instructing those new "initial" migrations to act as squashed migrations with the `replaces =` attribute.
    - All old RunPython-based data migrations are removed since they already did their jobs in prod.
    - In content-search, the RunSQL to create the search vector is manually reproduced.
    - In resources, a single RunPython migration is added for new databases to populate the FR Link Category config.
- Removed extra management commands that were causing ruff errors (not used anymore).
- Fixed a failing test.
- Removed the old "supplementary content -> supplemental content -> resources" renaming migrations from `regulations/migrations`.

It is best to try this on your local environment as well. To do so:
1. Download a local copy of prod.
2. Use the restore_db.sh script in the repo to restore the backup to Postgres at `localhost:5432`.
3. Switch to this branch.
4. Run `docker-compose exec regulations python manage.py showmigrations`, verify the output for `resources` and `content-search` shows that there is an unapplied squash.
5. Run `make python.migrate`, then the `docker-compose exec regulations python manage.py showmigrations` command again to verify that the squash is recorded correctly.
6. Verify that the site still works, particularly admin panel, search, sidebar etc., places where resources are used.
7. Run `make local.clean; make local.docker` to verify that the migrations work properly on a new database too.

Once that's done, validate this deploy:
1. Green check marks indicate successful migration.
2. Check the site in the same way as step 6 of the previous section.


peggles2 (approved):
LGTM 👍

PR #1384, opened 2024-07-24
EREGCSC-2772 Fix for "Can't delete internal files"

cgodwin1:
Resolves #2772

**Description-**

On prod, internal files are failing to delete. This is because the old code used `get_key()` and the new code uses `key` as a computed property. Note that this code, along with all upload/download code, will be refactored in a near-future ticket, so even though this isn't the most elegant or resilient code it will change soon. This PR only fixes the immediate issue.

- Fix this naming error.
- Remove redundant exception.

1. Go to the admin panel of the experimental deploy.
2. Go to internal files, and upload a _new_ file (don't use an existing file for this test, since they don't actually exist in the S3 bucket for this deploy).
3. Save the new file, then go back to it and hit Delete and confirm deletion.
4. Make sure the file is deleted, and no 500 error occurs.

### DIFF
--- a/solution/backend/resources/admin/internal_resources.py
+++ b/solution/backend/resources/admin/internal_resources.py
@@ -110,12 +110,7 @@ class InternalFileAdmin(AbstractInternalResourceAdmin):
 
     def del_file(self, obj):
         s3_client = establish_client("s3")
-        key = obj.get_key()
-
-        try:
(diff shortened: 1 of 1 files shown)


PhilR8 (approved):
New file uploaded to experimental deployment and deleted successfully.  LGTM!

PR #1395, opened 2024-07-30
EREGCSC-2726 Release branch for text extraction refactor

cgodwin1:
Resolves #2726

**Description-**

This release branch incorporates several pieces of work, including:
- Refactoring the text extractor to use SQS for batching extraction operations to allow for a more friendly, speedy UI.
- Continuing to decouple search from resources to avoid circular dependencies.
- Automatic extraction of resource text on upload or change.

This branch:
- Creates the SQS queue and loads the environment variables into eRegs.
- Separated the success/fail mixed case message for text extraction into two separate messages for clarity.

From https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1396:
- Removed unused and incorrectly named "textract" parameters from settings files.
- ~~Added base managers for ContentIndex and Resource types to defer loading of often-large `content` fields on all queries (including via reverse many-to-many relationships) by default.~~ (this didn't work, broke saving due to proxy object being returned after defer, replaced with separate model containing only content text linked via one-to-one)
- Added admin action for extracting text, with custom success/error message that handles pluralization.
- Removed unused "extract" endpoint in favor of admin action.
- Added search by URL to public resources on the admin panel.
- Copied the `content` field from ContentIndex to AbstractResource, see migration 0004 in resources.
    - This helps to avoid a circular dependency between search and resources apps from being created again.
    - Text extraction is now a function of the Resources app only. Other apps can hook into this functionality as needed.
    - The tradeoff to this is a duplicated `content` field, but the upside is no circular dependencies.
- Created `aws_utils.py` in `utils` directory of resources app. This file handles all connections from the resources app to AWS, including calling the text extractor.
- If configured, text extraction now occurs through SQS.
- Text extraction requests are batched in groups of 10 (the max available) and sent to SQS, allowing the operation to be very fast on the admin frontend (1-2 seconds for all 200 resources on the page, for example). Before batching, this operation took up to 15 seconds (and sometimes timed out) for all 200 resources on the page.
- Removed direct text extractor lambda access from serverless.yml and serverless-experimental.yml.
- Created "host.docker.internal:host-gateway" access for Linux Docker hosts.
- Added SQS event handling to the text extractor (note that followup work includes automated retries by throwing an exception instead of returning a 500 status code response if SQS is enabled).

From https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1402:
- Fixed the bug where async requests to the local text-extractor docker container would crash it.
    - Lambda-proxy now queues requests and only sends one to the underlying lambda at a time.
    - This was needed to prevent a deadlock locally. Django Admin locks the object that the text-extraction endpoint is trying to modify synchronously.
    - (Added benefit that it more closely mocks the deployed environment now that SQS is enabled.)
- The content field on resources is replaced by a one-to-one to a model containing only extracted text.
    - This is necessary because universally deferring the sometimes-large `content` field to prevent blowing up memory usage on resources endpoints caused saving any resource model to fail.
    - Note that this is because a defer operation does not return a real instance of the model, but a proxy. Saving the proxy doesn't fail outright but it does nothing silently.
- Added `auto_extract` field to `ResourcesConfiguration`, defaults to false.
- Resources will call the text extractor from the admin panel on save if the object is new, or if an existing object has its URL modified.

From https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1403:
- For all non-FR resources, the `extract_url` field is blanked out for now.
- Admin panel save_model will try to retrieve `extract_url` from the FR API if `url` or `document_number` has changed.
    - If it fails, a warning message is shown to the user and the event is logged, but saving continues unhindered.
- If `extract_url` is updated and `auto_extract` is true, the text extractor will be invoked even if `url` hasn't changed.
- Implemented `auto_extract` checking anywhere the text extractor is automatically called.
- Two new utility methods: `get_support_link` and `field_changed` in general purpose utils.
- Updated FR parser endpoint to utilize `transaction.atomic` that was already present (i.e. no need to check exceptions and delete the newly-created FR link).

This list of changes boils down to enabling extraction of text via admin panel action and on Resource object save if it's new, or if URL or (for FR docs) document number has changed. To validate:

1. On the experimental deploy, go a resource (say FR links).
2. Select multiple (or "all", if you feel like it) and run the "Extract text" action from the actions dropdown.
3. You should see "Text extraction started successfully for N resources" message, unless one or more links have an invalid URL.
4. Go to an individual FR link's change page.
5. Edit the document number by adding random numbers at the end, to cause it to be invalid.
6. Save, and verify an error message shows up but saving continues anyway.
7. Remove the random numbers and save again, and see that the object saves.
8. Do steps 5 through 7 again, but before, go to Resources Configuration and check "Auto Extract" first.
9. After step 7, you should see an additional message stating that text extraction started.

You can also validate local async text extraction like so:
1. Locally, rebuild the text extractor with `docker-compose stop text-extractor; docker-compose up -d --build text-extractor`.
2. Run `docker-compose logs -f --since "0m" text-extractor` and keep this open.
3. Go to the admin panel locally and go to FR links.
4. Select several links and run the extract action. Verify a success message appears.
5. Go back to your terminal and watch as the text extractor runs each extraction in order. This verifies async extraction works locally.

Also helpful to validate because the FR upload endpoint changed _slightly_:
1. Run `make frdocs.local` to run the FR parser locally.
2. Verify this completes without error.


PhilR8 (approved):
LGTM!

PR #1396, opened 2024-07-31
EREGCSC-2255 Action for indexing resources ("get content")

cgodwin1:
Resolves #2255

**Description-**

This PR adds an admin panel dropdown action for extracting content from resources. This replaces the old "Get content" button on the change page, which was due for refactor anyway because it redirected the user to a JSON page rather than indicating success or failure.

- Removed unused and incorrectly named "textract" parameters from settings files.
- Added base managers for ContentIndex and Resource types to defer loading of often-large `content` fields on all queries (including via reverse many-to-many relationships) by default.
- Added admin action for extracting text, with custom success/error message that handles pluralization.
- Removed unused "extract" endpoint in favor of admin action.
- Added search by URL to public resources on the admin panel.
- Copied the `content` field from ContentIndex to AbstractResource, see migration 0004 in resources.
    - This helps to avoid a circular dependency between search and resources apps from being created again.
    - Text extraction is now a function of the Resources app only. Other apps can hook into this functionality as needed.
    - The tradeoff to this is a duplicated `content` field, but the upside is no circular dependencies.
- Created `aws_utils.py` in `utils` directory of resources app. This file handles all connections from the resources app to AWS, including calling the text extractor.
- If configured, text extraction now occurs through SQS.
- Text extraction requests are batched in groups of 10 (the max available) and sent to SQS, allowing the operation to be very fast on the admin frontend (1-2 seconds for all 200 resources on the page, for example). Before batching, this operation took up to 15 seconds (and sometimes timed out) for all 200 resources on the page.
- Removed direct text extractor lambda access from serverless.yml and serverless-experimental.yml.
- Created "host.docker.internal:host-gateway" access for Linux Docker hosts.
- Added SQS event handling to the text extractor (note that followup work includes automated retries by throwing an exception instead of returning a 500 status code response if SQS is enabled).

Please note that there are some limitations still, such as being unable to directly view the `content` field on deployed environments. So validation can be split into a local component and an experimental deploy component.

Experimental deploy component (validates SQS is working):
1. Log into AWS and visit the SQS page. You should see a list of queues (there is only one), keep this page open.
2. Log into the admin panel on the experimental deploy in a different tab, and go to "Federal Register Links".
3. Perform a "Select all", then select the Actions dropdown and select "Extract text from selected resources".
4. Click "Go". After a moment you should see a success message indicating "Text extraction successfully started on 200 resources."
5. Immediately switch to the SQS tab, and repeatedly click the reload button.
6. You should see the number of messages available/in flight changing rapidly as the queue processes the 200 new messages. Most will complete quickly but some will remain, these are usually PDFs that take longer to process.
7. Go to the Cloudwatch logs for the 1396 text extractor, and click "Search all logs".
8. Click on the "30m" selector to limit to only the most recent runs, and enter into the filter "Function exited normally".
9. You should see a large number of these as the queue continues to process all the 200 requests. Note that it is currently possible for at least one to fail due to an unhandled text extractor issue where very large files cause a crash. This will be handled in a follow-up ticket.

Local deploy component (validates success/error messages and uploading to eregs):
1. Open a docker shell (`docker-compose exec regulations python manage.py shell`), and import FR Links: `from resources.models import FederalRegisterLink`.
2. Delete all content fields: `FederalRegisterLink.objects.all().update(content="")`
3. Go to the admin panel locally, and visit the Federal Register Link changelist.
4. Select some documents (note that PDFs will not extract unless you have local Textract keys available). I recommend sorting by recent and finding links that are not PDFs.
5. Select several, then run the Extract action as before. This will take a few moments because they are processing sequentially locally. If it times out, this is normal locally. Select fewer links and try again until you see a green success message.
6. Note the ID number of at least one of the links you just extracted, and switch back to your terminal.
7. Run `FederalRegisterLink.objects.get(pk=X).content` where X is the ID from the previous step. You should see extracted text as the output. This confirms the content upload endpoint is working.
8. With no local AWS keys in your environment, attempt to run the extraction on a link that goes to a PDF. After a moment, you should see an error message displayed.
9. Select an HTML link and a PDF link to see the error message adapts to the output (1 success and 1 failure). This confirms that the admin panel messaging system is working.


PhilR8 (approved):
LGTM!

PR #1402, opened 2024-08-12
EREGCSC-2624 Implement auto extract functionality

cgodwin1:
Resolves #2624

**Description-**

When we add a public link, FR link (whether it's added by the FR parser or by hand), internal link, or internal file, the content should become searchable if possible. This PR enables that by allowing for automatic extraction of resources on save.

- Fixed the bug where async requests to the local text-extractor docker container would crash it.
    - Lambda-proxy now queues requests and only sends one to the underlying lambda at a time.
    - This was needed to prevent a deadlock locally. Django Admin locks the object that the text-extraction endpoint is trying to modify synchronously.
    - (Added benefit that it more closely mocks the deployed environment now that SQS is enabled.)
- The content field on resources is replaced by a one-to-one to a model containing only extracted text.
    - This is necessary because universally deferring the sometimes-large `content` field to prevent blowing up memory usage on resources endpoints caused saving any resource model to fail.
    - Note that this is because a defer operation does not return a real instance of the model, but a proxy. Saving the proxy doesn't fail outright but it does nothing silently.
- Added `auto_extract` field to `ResourcesConfiguration`, defaults to false.
- Resources will call the text extractor from the admin panel on save if the object is new, or if an existing object has its URL modified.

Auto-extraction has already been validated via https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1403, no need to re-validate this behavior. Use this PR only to validate that the lambda_proxy changes are working properly on your machine.

1. Rebuild the text extractor with `docker-compose stop text-extractor; docker-compose up -d --build text-extractor`. This enables the new async functionality. 
2. Run `docker-compose logs -f --since "0m" text-extractor` to follow the most recent text extractor logs.
3. Locally, go to the admin panel and go to FR Links.
4. Select several FR Links and run the "Extract" action.
5. After a moment, you should see that all come back successful (except in cases where the URL is invalid etc.)
6. Check the logs in your terminal and verify the lambda didn't crash and no errors appeared.


PhilR8 (approved):
LGTM!

PR #1404, opened 2024-08-13
EREGCSC-2779 -- Vitest fixes

PhilR8:
Resolves [EREGCSC-2779](https://jiraent.cms.gov/browse/EREGCSC-2779)

**Description**

The V4 refactor broke some front end tests and made some others obsolete.  This PR will remove the unneeded tests and fix the remaining broken tests.

**This pull request changes:**

- Adds new Makefile command to run Vitest UI locally
   - `make test.vitest.ui`
- Updates test fixtures to reflect updated API responses where necessary
- Removes unneeded assertions and test suites
- Removes failing snapshot tests and associated snapshot files
   - Keeps passing snapshot tests for now
- excludes MSW directory from coverage calculations

**Steps to manually verify this change:**

1. Green check marks
2. Maybe run the Vitest tests locally and see how it looks? 
   - `make test.vitest.ui`

PR #1044, opened 2023-11-08
[Hotfix] Revert CMS Design System back to v2.5

PhilR8:
Resolves regression introduced with updated version of CMS Design System

**Description**
[Snyk created a Pull Request](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1042) that bumped CMS Design System's NPM package from v2.5.0 to 2.13.0.  This was merged into `main` before it was discovered that it introduced a regression where the focus styles (outline and background) were changed and did not always look good.

**This pull request changes:**

- Reverts CMS Design System package version to 2.5.0

**Steps to manually verify this change:**

1. Tab around the homepage and make sure the focus styles aren't pink outlines and sometimes white backgrounds

PR #1408, opened 2024-08-21
[Hotfix] Ensure CMS Design System npm package version is 2.5.0

PhilR8:
**Description**

[EREGCSC-2779](https://jiraent.cms.gov/browse/EREGCSC-2779) (Vitest fixes) introduced a regression where the CMS Design System npm package was updated to a newer minor version.  Our targeted version is `2.5.0`, but the [pull request for 2779](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1404) rebuilt the `package-lock.json` file to install `2.13.0`.

We don't want `2.13.0` because it includes unwanted style changes.  Specifically, element `focus` styles are pink instead of blue.

We've previously attempted to upgrade to `2.13.0` based on a Snyk suggestion, but that upgrade [was reverted](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1044) back to `2.5.0` for this same reason.

Except we didn't actually lock the version to `2.5.0`.  Note this line from `package.json`:

```
"@cmsgov/design-system": "^2.5.0",
```

The caret (`^`) before the version indicates that minor (`0.X`) and patch (`0.0.X`) versions can be updated if available and compatible.  See [NPM documentation here](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004).  This means that when rebuilding the `package-lock` file, the most recent `2.X.X` version will be installed.  In our case, `2.13.0` is the most recent version and was installed.

This issue is happening now because the `package-lock.json` file was recently rebuilt as part of EREGCSC-2779.

- Removes caret from CMS Design System version number

1. Green check marks
2. Use `tab` to navigate around the site via the keyboard.  Ensure that you don't see any pink focus styles around any elements.

### DIFF
--- a/solution/ui/regulations/package-lock.json
+++ b/solution/ui/regulations/package-lock.json
@@ -7,7 +7,7 @@
       "name": "regulations",
       "license": "CC",
       "dependencies": {
-        "@cmsgov/design-system": "^2.5.0",
+        "@cmsgov/design-system": "2.5.0",
         "@fortawesome/fontawesome-free": "^5.15.4",
         "@vitejs/plugin-vue": "^4.6.2",
         "@vue/test-utils": "^2.4.6",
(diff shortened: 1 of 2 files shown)


peggles2 (approved):
LGTM 👍

PR #1414, opened 2024-09-03
EREGCSC-2823 Fix for unapproved resources showing in search

cgodwin1:
Resolves #2823

**Description-**

The `content-search` endpoint is not filtering out unapproved resources, resulting in unapproved resources showing in search results if indexed.

- Unapproved resources are filtered out from content-search results.
- Added a test to verify that this is the case.
- Removed an irrelevant content-search test that was testing the resources app.

1. Go to the search page and search for any term. Make note of the first result's ID.
2. Go to the admin panel and find the instance corresponding to the first result on the search page.
3. Mark that resource as unapproved and save it.
4. Refresh the search page. That result should now be missing, as expected.


peggles2 (approved):
LGTM 👍

PR #1417, opened 2024-09-10
EREGCSC-2825 Enable granular counts for types of search results

cgodwin1:
Resolves #2825

**Description-**

When a researcher is looking at items in search results, they may want to filter by choosing whether to see internal documents, public documents, or regulations.

To help a person decide whether to click these filter options, the filter options should display counts of how many items are available for that kind of document within the visible results.

We want these numbers to dynamically update if a person applies a filter, such as a category filter.

- ViewSetPagination class moved from common dir to `cmcs_regulations/utils/pagination.py`. (Goal is to eventually remove the common directory.)
- Added the ability to override ViewSetPagination and apply additional attributes (by overriding `get_additional_attributes()`) and additional counts (by overriding `get_count_map()`) to the top-level of the paginated response (alongside `next`, `previous`, etc.)
- Added `public_count`, `internal_count`, and `reg_text_count` to the paginated content-search endpoint response.
- Added `public_count` and `internal_count` to the paginated resources endpoints response.

1. Visit `/v3/content-search?q=test` and verify that `public_count`, `internal_count`, and `reg_text_count` all show up and the numbers in those fields add up to the number in the `count` field of the response.
2. Verify that if logged out, the same response will show `internal_count: 0` where it would otherwise show a number if logged in.
3. Do steps 1 and 2 for `/v3/resources`.
4. Do the same for at least one of these: `/v3/resources/public/links`, `/v3/resources/public/federal_register_links`, `/v3/resources/internal/files`, `/v3/resources/internal/links`. (Testing all isn't necessary as they all inherit from the same base class.)


PhilR8 (approved):
LGTM!

PR #1418, opened 2024-09-11
EREGCSC-2657 -- One Column Search (take two)

PhilR8:
Resolves [EREGCSC-2657](https://jiraent.cms.gov/browse/EREGCSC-2657)

**Description**

Updates the search page to be a Single Column design that uses the updated `/content-search` endpoint to display Regulations and Resources mixed together.

- creates new `Search.vue` front end view to use for the `/search` route; uses Vue 3 Composition API and borrows heavily from `Subjects.vue`
- creates `useSearchResults` composable to encapsulate all logic related to fetching from `/content-search` and setting/updating Vue refs related to the response.  This removes duplicated logic from `Subjects.vue` and the new `Search.vue`
- creates `_eregs_design_system.scss` stylesheet to contain eRegs Design System constants to use moving forward
   - starts using CSS variables in anticipation of upcoming CMS Design System upgrade
- styles `/search` page to match [Figma comp](https://www.figma.com/design/j3ttFC9FtexKXBrV4yb3m7/Search-Everything?node-id=926-83599&node-type=FRAME&m=dev)
- Updates and creates Cypress end to end tests and Vitest unit tests where appropriate and needed

1. Green check marks
2. Visit [experimental deployment](https://8e0xj3xnrl.execute-api.us-east-1.amazonaws.com/dev1410) and try some searches
3. Ensure that regulations items in the search results have PUBLIC and REGULATIONS labels
4. Clicking on a Regulations result item should not open a new tab, it should take you to the Reader View and should highlight any search terms that entered by user and/or returned by the server
5. Ensure that when logged out, you can only see the Regulations and Public checkboxes
6. Ensure that when only Regulations checkbox is checked, the Categories dropdown disappears and is not visible
7. When logged in, ensure that all three checkboxes are visible and usable (Regulations, Public, Internal)
8. Ensure that counts look correct
9. Ensure that request params are correct based on selected checkboxes, query strings, and category selections


cgodwin1 (approved):
LGTM!

PR #1423, opened 2024-09-20
EREGCSC-2850 Enable querying for item counts when document type is not selected

cgodwin1:
Resolves #2850

**Description-**

Counts were not properly displayed when one of the show_* flags was passed to the content-search endpoint, or from the subjects page if internal or public were unselected. This is because all filtering was done first before counting. To solve this problem (and reduce endpoint overhead), we decided to move counting to a separate endpoint dedicated to only this task (`/v3/content-search/counts`).

- New pagination function removed (`get_additional_attributes()` is retained).
- New endpoint `/v3/content-search/counts` exists and returns counts of public, internal, and reg text for given filters.
- Endpoint is linked to from the pagination header of the `/v3/content-search` and `/v3/resources` endpoints via `count_url`.
    - For `/v3/resources`, this is done in a way that avoids a permanent dependency (by catching `ImportError` if `ContentCountViewSet` does not exist).
- Tests reflect that total count must equal all combined counts.
- Updated OpenAPI schema for pagination to reflect `count_url` parameter.

1. Visit `/v3/content-search` and apply several filters (for example `?q=test`, `?q=test&citations=42.433.10`, `?q=test&show_public=false`).
2. Note the existence of `count_url` and visit that URL to see the broken down counts for each test search. Ensure that this is an accurate cont.
3. Do steps 1-2 for `/v3/resources`, omitting the `?q=test` parameter.

### DIFF
--- a/solution/backend/cmcs_regulations/utils/pagination.py
+++ b/solution/backend/cmcs_regulations/utils/pagination.py
@@ -1,4 +1,3 @@
-from django.db.models import Count
 from drf_spectacular.utils import OpenApiParameter
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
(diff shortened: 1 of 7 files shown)


PhilR8 (approved):
LGTM!

PR #1433, opened 2024-10-02
EREGCSC-2805-B Remove jira-host setting

cgodwin1:
Resolves #2805

**Description-**

Update docs say: "You don't need to specify the Jira host if it's 'jiraent.cms.gov'. So you can remove this". Including this key causes an error message in the action, so better to remove it.

- Remove jira-host setting from jira sync github action

1. Deploy this to prod.
2. Wait until after 6:25am the following day.
3. Check the securityhub action and the action is successful with no errors.


PhilR8 (approved):
LGTM!

PR #1454, opened 2024-10-25
Enable Dependabot to check for GitHub Actions updates

brittag:
No associated Jira ticket.

**Description**

This pull request enables Dependabot to suggest updates to GitHub Actions. See the explanation here: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

You could check the linked documentation to make sure this looks right. Otherwise we'll know if this is working if we start getting some PRs to update our actions.


cgodwin1 (approved):
LGTM!

PR #1467, opened 2024-11-04
[Hotfix] Ensure intercept is calling fixture for homepage test suite

PhilR8:
Resolves failing Homepage end to end test suite test on `prod`

**Description**

The Cypress homepage test suite has a failing test on the `prod` environment.  The test is supposed to be using an `intercept` to replace the server response with a fixture file.

The issue: the intercept is incorrectly specifying the path to be matched. As a result, the test is being run against an actual response from the server, which is returning data that is not expected by the test.

- Updates the intercept path to add a glob (`**`) to the end of the intercept path so that the correct fixture file is returned by the API call instead of real data from the server
- Updates the fixture file to include a custom Subject name that does not and will never exist in prod to verify that the fixture is being returned by the API call

1. Green check marks
2. Deploy to `prod` and ensure that the test passes


cgodwin1 (approved):
LGTM

PR #1483, opened 2024-11-14
EREGCSC-2900 Prevent text extractor from infinitely retrying, add max pages for PDF extraction

cgodwin1:
Resolves # 2900

**Description-**

We saw an increase in Textract charges this month and tracked it back using Lambda monitor charts to a period between 11/6-11/8 where there was a consistent number of extractor errors. Checking CloudWatch we saw the extractor continually attempting to process an object and timing out after 900 seconds (15 mins).

We were able to reasonably guess that it was the result of the extractor attempting to extract a 500 page PDF that was only 6 MB. Given that the max file size was 20 MB, it didn't exceed that threshold, but since each page is sent to Textract sequentially, we could guess that it would easily exceed 900 seconds. (Max processing time per page is 900/500 = 1.8 sec/page, which we are unlikely to beat.)

Additionally, since the Lambda orchestrator is what is causing the timeout, the Lambda itself never successfully consumes the queue message and it is returned to the queue an indefinite number of times. We can fix this with a dead-letter queue configured to receive failed messages after a certain number of attempts.

- PDFs are limited by _page_ now instead of size. Extractor will only process up to 50 pages.
- If 50 pages is exceeded, return the first 50 pages worth of text and exit.
- Added a dead-letter queue for the text extractor. If an unhandled error occurs 5 times, send the message to the DLQ and do not reprocess.

**Followups for the future:**

- Handle _all_ errors using the queue instead of retrying internally. (Web extractor retries, etc., should be removed.)
- Find a way to extract PDFs in parallel. Currently, the PDF extractor converts the PDF to a set of images and sends them one-by-one to Textract. What if we instead sent each image/page to the queue to be processed whenever possible?
    - Textract extractor could easily save each image/page to an S3 bucket, which is already a supported backend for the extractor. Then add the option to _delete_ an object from S3 if configured and permitted, and when consumed.
    - We could modify eRegs' backend to permit sending numbered chunks of text for a given Resource object instead of only allowing the entire text to be sent.
    - Could add two fields to the HTTP patch request JSON: "chunk" representing the current chunk number, and "chunks" representing the total number.
    - If the above are non-zero, the update method would put placeholders in the text field until the corresponding chunk comes in, then replace the placeholder with the chunk's value.
    - Estimated points for this work: Maybe a 3. Would be worthwhile in our quest to eliminate timeouts due to Lambda limitations.

1. Tests pass (added a test for max page processing).
2. Attempt to extract the 500 page, 6 MB PDF again and verify that it does not fail and does extract some text. (Use the "Index populated" filter on the admin panel, and/or see if it comes up in search results.)
3. Go to AWS for dev -> Simple Queue Service -> dev1483-text-extractor-queue.
4. Notice under "Details", "Dead-letter queue" is set to "Enabled".
5. Go to the "Dead-letter queue" tab.
6. Notice the queue is set to the correct dead-letter queue ARN, and max retries (called "max receives" in SQS lingo) is set to 5.

### DIFF
--- a/solution/text-extractor/extractors/pdf.py
+++ b/solution/text-extractor/extractors/pdf.py
@@ -11,11 +11,14 @@ logger = logging.getLogger(__name__)
 
 class PdfExtractor(Extractor):
     file_types = ("pdf",)
-    max_size = 20
+    max_pages = 50
+    output_file_type = "jpeg"
+    
 
-    def __init__(self, file_type: str, config: dict):
+    def __init__(self, file_type: str, config: dict, *args, **kwargs):
         super().__init__(file_type, config)
-        self.extractor = Extractor.get_extractor("jpeg", config)
+        self.output_file_type = kwargs.pop("output_file_type", self.output_file_type)
+        self.extractor = Extractor.get_extractor(self.output_file_type, config)
 
     def _convert_to_images(self, file: bytes, temp_dir: str) -> [str]:
         logger.debug("Converting PDF file to images stored in a temporary directory.")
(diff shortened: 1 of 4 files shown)


PhilR8 (approved):
LGTM!

PR #1487, opened 2024-11-19
EREGCSC-2862 -- Update sidebar for subjects pages

PhilR8:
Resolves [EREGCSC-2862](https://jiraent.cms.gov/browse/EREGCSC-2682)

**Description**

Updates left sidebar on `/subjects` page to match [Figma design](https://www.figma.com/design/LfXWcXJI49UaxMfFTsQHZO/Browse-without-Search?node-id=1-18968&node-type=frame&t=WZzWcJZqfFDePuVR-0).

- Uses `position: sticky` to fix left sidebar in place so it does not scroll with the rest of the page
   - Left sidebar can now scroll independently
- Subject Selector in left sidebar is restyled to match Figma design
- Subject Selector behavior changes:
   - Selected subject is displayed underneath text input in left sidebar
   - Selected subject and text input stick to the top of the page when left sidebar is scrolled -- the are always visible
   - Selected subject is removed from subject list
   - When subject is unselected, it is added back to subject list

1. Green check marks
2. [Visit experimental deployment](https://i52gwakusd.execute-api.us-east-1.amazonaws.com/dev1487)
3. Compare experimental deployment to [Figma comp](https://www.figma.com/design/LfXWcXJI49UaxMfFTsQHZO/Browse-without-Search?node-id=1-18971&node-type=frame&m=dev) to see if they match
4. Select a subject and make sure it is removed from the subject list and re-added to the list when unselected
5. Arrive at page with a selected subject and ensure everything works as expected
6. Visit `/search` and use the Subjects dropdown and ensure there have not been any changes or regressions


solution/ui/regulations/eregs-vite/src/components/subjects/PolicySelections.vue
```diff
@@ -29,48 +35,20 @@ const removeClick = () => {
...
-            .sort((a, b) => {
-                // sort by label and then by name
-                if (a.label < b.label) return -1;
-                if (a.label > b.label) return 1;
-                if (a.name.toLowerCase() < b.name.toLowerCase()) return -1;
-                if (a.name.toLowerCase() > b.name.toLowerCase()) return 1;
-                return 0;
-            });
-    }
-);
```
PhilR8, line 55: Originally, the Subjects page allows multiple subjects to be selected.  It was also a possibility that other filters (such as Category) would also be listed as chips alongside subjects at the top of the left sidebar.

However, this page evolved to allow only a single Subject to be selected at a time, and other filters have their own elements for selection and display.  (For example, the Categories dropdown at the top of the results pane allows selection and displays the selected category.)

Since only single subjects can be selected simultaneously, the sorting logic above is no longer necessary.  Additionally, we no longer have to watch for a change in the selectedParams object: we are now able to pass in the selected subject as a prop, and the component re-renders as that prop changes.

cgodwin1 (approved):
LGTM!

PR #1488, opened 2024-11-19
EREGCSC-2894 Update CDK bootstrap for latest version

cgodwin1:
Resolves #2894

**Description-**

In order for our CDK scripts to reliably deploy/create AWS resources, we need to keep our CDK stacks up to date. Sometimes AWS releases updates, and we have to apply them, and it's happened already!

We decided we wanted to do this automatically via cron'd Github Action, but found it wasn't possible to do this using standard bootstrap procedure because we are relying on a CMS-specific `template.yaml`.  By diffing the CMS `template.yaml` with the [default one from CDK](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml), we found it was feasible to create a script (`update_template.py`) that attempts to automatically apply those needed changes to the default template.

From there, we can automate the bootstrap update procedure.

- New Github Action, `update-cdk-bootstrap.yml`, that runs weekly on Monday and runs the commands needed to execute a CDK bootstrap update.
- New script, `update_template.py`, which takes as CLI arguments a JSON file containing a list of roles to update, the default CDK template to process, and an output filename.
    - JSON file and optional arguments allow customization, see `./update_template.py -h` for more info.
- Github Action downloads the default CDK template, runs `update_template.py` to generate a usable one, and then performs the bootstrap on an empty "temp" project directory.
- Because prod (and currently val) have protection rules, only the dev (and eventually val) bootstrap updates will run automatically. This means a dev will need to approve bootstrap updates for prod, which while still requiring user action, is probably the appropriate thing to do anyway. We can work around this later if we want, using a set of custom environments specific to this action.

You can't directly verify this without merging to prod, because this action runs on a schedule (see the cron section of the action file). Instead, [check out this test run](https://github.com/Enterprise-CMCS/cmcs-eregulations/actions/runs/12019176455). You can go to each environment's run and see that with the updated template.yaml file that the script generates, the bootstrapping process completed successfully for dev, val, and prod.


PhilR8 (approved):
LGTM!

PR #1501, opened 2024-12-13
EREGCSC-2902 - Fix CDK remove command

addis-samtek:
Removed the --exclusively flag from CDK destroy commands as it was causing execution issues during PR-specific stack cleanup.


PhilR8 (approved):
LGTM

cgodwin1 (approved):
LGTM as well

PR #1506, opened 2024-12-30
EREGCSC-2363-B Fix escape vulnerability in text extraction user message on admin panel

cgodwin1:
Resolves #2363

**Description-**

Pen-testing revealed a vulnerability in the admin panel where it is possible to execute arbitrary code by inserting the code into the title of a resource and saving it. This triggers a text extraction user message to be displayed with the title of the resource in the message.

- Use Django's `escape()` function around user-entered strings that are displayed in an HTML formatted user message.

1. In the experimental deploy, go to the admin panel and sign in.
2. Go to a public link and arbitrarily add junk to the end of the URL to trigger a text extraction event (e.g. if URL is http://google.com, add "/test" to the end).
3. Replace the `title` field with the following: `CCIC">`
4. Save the resource.
5. You should see a message on the top of the screen like this: `Text extraction requested for Public Link "2000-12-15 CCIC">".`
6. You should NOT see a popup prompt. If you see no text extraction message at all, go to "Resources Configuration" on the left sidebar and ensure "Auto Extract" is checked.

### DIFF
--- a/solution/backend/resources/admin/resources.py
+++ b/solution/backend/resources/admin/resources.py
@@ -67,12 +67,12 @@ class AbstractResourceAdmin(CustomAdminMixin, admin.ModelAdmin):
             url = f"<a target=\"_blank\" href=\"{reverse('edit', args=[obj.pk])}\">{escape(str(obj))}</a>"
             if fail:
                 logger.error("Failed to invoke text extractor for resource with ID %i: %s", obj.pk, fail[0]["reason"])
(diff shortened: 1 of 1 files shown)


PhilR8 (approved):
Alert currently appears on `dev` website admin panel after editing, saving, and automatic extraction.  

Validated on experimental deployment that alert does NOT appear after editing, saving, and auto extraction.  LGTM!

PR #1525, opened 2025-01-22
[Hotfix] EREGCSC-2734 -- Hide Get Account Access link when logged in on Subjects/Statutes page

PhilR8:
Fixes defect found after deploying [EREGCCS-2734](https://jiraent.cms.gov/browse/EREGCSC-2734) to `prod`

**Description**

The "Get Account Access" link in the top left corner of the header should only be visible if the user is not logged in.  Once the user logs in, the "Get Account Access" link should not exist.

Currently, the Subjects and Statutes pages are still showing the "Get Account Access" link in the header when logged in.

- Adds `v-if` directive to `AccessLink` component when used on Subjects and Statutes pages.
   - The `v-if` directive conditionally renders the component if the directive's expression is truthy
   - The `isAuthenticated` prop is being used in the directive's expression: `!isAuthenticated` must be truthy for the component to render.
   - This expression was present in the Search page, but missing from the Subjects and Statutes page.
- Adds Cypress end to end tests to ensure "Get Account Access" link exists or does not exist where expected

1. Green check marks
2. Before logging in, explore around the site and note that the "Get Account Access" link is visible in the header everywhere you go
3. Log in
4. Explore around the site to ensure that "Get Account Access" link in header is not visible when visiting all areas of the experimental deployment while logged
5. Pay special attention to `/subjects` and `/statutes` -- this is where the defect occurs. When logged in, you should not see the "Get Account Access" link on these pages.

### DIFF
--- a/solution/ui/e2e/cypress/e2e/get-access-page.cy.js
+++ b/solution/ui/e2e/cypress/e2e/get-access-page.cy.js
@@ -1,6 +1,12 @@
 const username = Cypress.env("TEST_USERNAME");
 const password = Cypress.env("TEST_PASSWORD");
 
+Cypress.Commands.add("checkHeaderLink", ({ shouldBeVisible = false }) => {
+    const desiredState = shouldBeVisible ? "exist" : "not.exist";
+    cy.get("a[data-testid='get-account-access-wide']").should(desiredState);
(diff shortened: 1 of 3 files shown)


cgodwin1 (approved):
LGTM!

PR #1552, opened 2025-02-11
EREGCSC-2859-H Split CDK deploy/remove Github Actions into own files

cgodwin1:
Resolves #2859

**Description-**

This PR should complete the AC for 2859 after the previous PR (#1550) was deployed. Additionally it fixes a bug in the ephemeral environment where it was using the `eregs` database instead of its own ephemeral copy of the `eregs` database. Finally, it fixes an issue where the AWS region was set as a repo secret, causing the API Gateway URL output from the lambda deploy step to be blank (this is needed for the notify step).

- CDK deploy and remove steps are split out into their own files separate from Serverless (deploy-experimental-cdk.yml and remove-experimental-cdk.yml respectively).
- API CDK stack is configured with environment variable `DB_NAME` as `eph_$PR_NUMBER` for ephemeral or `eregs` otherwise.
- On deploy, AWS CLI is used to invoke the createdb, migrate, and createsu lambdas, respectively.
- On remove, AWS CLI is used to invoke the dropdb lambda.
- Switch `AWS_DEFAULT_REGION` from a repo secret to a repo environment variable.

1. Verify ephemeral deploy (regular and CDK actions) succeeds. (Cypress test for CDK will fail due to the region secret error- merge this PR and I will remove the secret to fix.)
2. Go to the ephemeral CDK site and verify it loads and does not show a 500. (This tests that the database is being created and migrated properly.)


PhilR8 (approved):
LGTM!

PR #1559, opened 2025-02-19
EREGCSC-2872 -- Make page widths consistent and usable

PhilR8:
Resolves [EREGCSC-2872](https://jiraent.cms.gov/browse/EREGCSC-2872)

**Description**

Our site is somewhat hard to read and use in wide browser windows - for example, search results are pinned to the left, while the regulations are centered but can have a line width that is wider than recommended for easy reading.

To give our site a more consistent experience at large browser widths, we want to center content for all pages, and have a consistent max width of 1024px for search results and informational pages such as About, Sign In, and Get Account Access.

- Adds CSS variable `--content-max-width` and sets it to 1024px
- Adds reusable SCSS mixins 
   - `eds-max-width`: encapsulates properties to set `max-width` to a container and center that container when screen width is wider than max width
   - `eds-width-with-gutters`) encapsulates properties to responsively set `width` if `eds-gutters` mixin is used
- Refactors Vue component logic where needed
   - Remove Statutes banner left margin logic
- Various tweaks to HTML markup to add/remove classes where necessary

1. Green check marks
2. Visit [experimental deployment](https://n9v5v93z8k.execute-api.us-east-1.amazonaws.com/eph-1559)
3. Navigate around all areas of site at various screen widths
   - Note that the site should be largely unchanged at screen widths below 1024px, except the Statutes page that now makes better use of the entire width of the screen
   - Note that Statutes and Subjects pages should be centered and content should expand to fit the entire width of the page at all screen sizes with gutters that match Header gutters
   - Ensure that Search results block is centered, with equal margins on both sides and a max width of 1024px
   - Ensure that About, Login, and Get Account Access page content is centered, with equal margins on both sides and a max width of 1024px


cgodwin1 (approved):
LGTM!

PR #1565, opened 2025-02-20
EREGCSC-2922-B Fix workflow order for CDK deploy

cgodwin1:
Resolves #2922

**Description-**

Matrix strategy attempted to run each job for all environments before moving to the next job. We want the opposite: run all jobs for one environment before moving onto all jobs for the next environment. To utilize a multi-job strategy we must use a "reusable workflow" implementation.

- Two actions: deploy-cdk-manager.yml and deploy-cdk.yml
- The manager action uses a matrix to call the deploy-cdk action for each environment
- `max-parallel: 1` ensures only one deploy at a time

1. Approve and deploy
2. Check out both new actions and ensure _all_ of the dev deploy happens before val attempts to start

### DIFF
--- a/.github/workflows/deploy-cdk-manager.yml
+++ b/.github/workflows/deploy-cdk-manager.yml
@@ -1,0 +1,26 @@
+name: Deploy CDK (Manager)
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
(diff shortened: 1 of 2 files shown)


PhilR8 (approved):
LGTM!

PR #1564, opened 2025-02-20
EREGCSC-2922 Advance CDK deploy to val account

cgodwin1:
Resolves #2922

**Description-**

CDK is validated in dev, we want to advance to val for further testing and validation before deploying to prod.

**This pull request changes...**

- Rename action "Deploy CDK to Dev" to "Deploy CDK"
- Change matrix in action to permit Github Actions to deploy the CDK version of the site to val
- Semi-related fix: syntax for aws lambda invoke for CDK would silently fail if AWS CLI returned an error before invocation occured

**Steps to manually verify this change...**

1. Approve and merge this PR
2. Check the "Deploy CDK" action for success deploying to dev
3. On success, approve deploy to val
4. Validate successful deploy in val (check AWS console, check the site, etc.)

PR #1567, opened 2025-02-21
EREGCSC-2922-C Convert deploy-cdk action to one big job for matrix purposes

cgodwin1:
Resolves #2922

**Description-**

The previous PR (#1565) was worse than (#1564) because the use of reusable workflows ruined the flow diagram and required approval for each job instead of one per environment.

- Convert deploy-cdk.yml into a single large job with many steps, much like our current serverless deploy.yml

1. Approve and merge
2. Observe that the "Deploy CDK" action deploys successfully to dev, then val if approved, and behaves very similarly to our current "Deploy" action


PhilR8 (approved):
LGTM!

PR #1582, opened 2025-03-03
EREGCSC-2934 -- Adjust regulation history display

PhilR8:
Resolves [EREGCSC-2934](https://jiraent.cms.gov/browse/EREGCSC-2934)

**Description**

In order to help policy researchers who need to view past versions of a regulation, each regulation section provides the option to "view regulation history", with a list of years linked to past versions.

We need this list to read left to right starting at the top right of the list.

We also need the popup to dynamically size itself better.  Specifically: the list should be as tall as it needs to be without any list items overflowing outside of the tooltip container.

- Updates CSS style rules to make tooltip list work as described above
- Refactors tooltip Vue components to use Composition API (?)
- TBD

1. Green check marks
2. TBD

### DIFF
--- a/solution/ui/regulations/css/scss/partials/_buttons_and_tooltip.scss
+++ b/solution/ui/regulations/css/scss/partials/_buttons_and_tooltip.scss
@@ -224,9 +224,9 @@
                         40px
                 );
 
-                @include screen-sm {
+                @include screen-md {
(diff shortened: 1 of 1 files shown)


cgodwin1 (approved):
LGTM!

PR #1573, opened 2025-02-25
EREGCSC-2932 -- Refactor homepage Vue components to Composition API

PhilR8:
Resolves [EREGCSC-2932](https://jiraent.cms.gov/browse/EREGCSC-2932)

**Description**

TBD

**This pull request changes:**

- TBD

**Steps to manually verify this change:**

1. Green check marks
2. TBD

PR #1586, opened 2025-03-05
[Hotfix] Reader Sidebar rules lost their pipe separator

PhilR8:
Resolves iteration bug caused by [EREGCSC-2932](https://jiraent.cms.gov/browse/EREGCSC-2932) ([pull request](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1573))

**Description**

Fixes a regression where Related Rules in the Reader view sidebar lost the pipe that separates date and Document ID.  This regression was caused by EREGCSC-2932, which updated the pipe separator styles for the homepage items but did not properly account for side effects elsewhere in the app.

- Creates SCSS mixin to encapsulate `recent-date--bar` styles to be reused throughout the app
- Uses this mixin for Related Rules on the homepage and in the Reader View sidebar

1. Green check marks
2. Reader view sidebar rules have a pipe separator that matches results items everywhere else on the site

### DIFF
--- a/solution/ui/regulations/css/scss/_application_settings.scss
+++ b/solution/ui/regulations/css/scss/_application_settings.scss
@@ -595,6 +595,12 @@ select {
     @include common-chip-colors;
 }
 
+@mixin recent-date-pipe {
+    padding-right: var(--spacer-1);
+    margin-right: var(--spacer-1);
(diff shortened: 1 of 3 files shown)


cgodwin1 (approved):
LGTM!

PR #1590, opened 2025-03-10
EREGCSC-1593, EREGCSC-2688 Dynamic path determination for API Gateway

cgodwin1:
Resolves #1593, 2688

**Description-**

Previously we were relying on an environment variable to tell the site where the site is hosted, due to differences between API Gateway (with `/stage` root path) and a custom domain (with `/` root path). Django can determine this for us by relying on the fact that _all_ API Gateway URL's have a stage.

Additionally, the `createdb` lambda was occasionally failing in dev with a 3 minute timeout. The time it takes this lambda to run is expected to grow linearly with the size of the database.

- `CUSTOM_URL` used only for allowed hosts and redirect lambda.
- `CUSTOM_URL` is not included in non-prod environments. (This allows us to delete the parameter from SSM in dev, val.)
- `SITE_ROOT` variable loaded into Django's template context with a context processor.
- `SITE_ROOT` determined from Django request object, _not_ the `CUSTOM_URL` variable.
- Increase `createdb` lambda timeout to 15 minutes, which should prevent timeouts for the foreseeable future.
- Removed unused database fixtures that previously were used to for local development and testing.

1. If the SPA pages (search, subjects, statutes) work locally after rebuilding (`make local`) then we can be reasonably assured that the site will work with a `/` root path (prod).
2. If those same pages work on the experimental deployment, then we know for a fact that the site works with a `/stage` root path.
3. Otherwise, green checks.


PhilR8 (approved):
LGTM!

PR #1596, opened 2025-03-17
[hotfix] Include custom_url param in redirect lambda environment

cgodwin1:
Resolves #n/a

**Description-**

CUSTOM_URL param is missing from the redirect lambda's environment, causing it to fail with a 500 error.

- For prod only, add the CUSTOM_URL parameter to the lambda's environment.

1. Approve and deploy to dev, val, and prod
2. Go to the redirect lambda URL and verify it redirects, no 500 error


PhilR8 (approved):
LGTM!

PR #1597, opened 2025-03-17
[hotfix 2] Include schema in redirect lambda's URL

cgodwin1:
Resolves #n/a

**Description-**

Calling Python's `urljoin()` without a schema (i.e. `https://`) will result in the domain part being ignored entirely. In the redirect lambda, this results in a redirect to `/`, creating an infinite redirect.

- Add `https://` before the domain

1. Deploy this to prod
2. Go to the redirect lambda and verify it redirects successfully


PhilR8 (approved):
LGTM!

PR #1598, opened 2025-03-18
EREGCSC-2915 Remove Serverless code and AWS resources

cgodwin1:
Resolves #2915

**Description-**

Now that we have migrated to CDK, we need to remove our old Serverless stacks and deploy code.

- All serverless.yml files removed.
- `requirements.txt` removed from `static-assets`; `requirements_tmp.txt` in `backend` renamed.
- Serverless Github Actions removed.
- `*-cdk.yml` actions renamed.
- Other actions renamed for consistency.
- Cypress test added to prod CDK deploy.
- Django unit tests moved to separate action invoked automatically via push or merge.
- Added Github Action to automatically run unit tests for the text extractor (**note** this is currently failing, but not our concern right at this moment).
- Added new badges to README.
- CDK version pinned:
    - New versions of CDK have a tendency to break things as we have observed several times now.
    - The CDK bootstrap update action will now only run manually via workflow dispatch.
    - CDK CLI and library versions pinned via package.json in `cdk-eregs` dir. Patch updates allowed with `~` syntax.
    - Removed all `npm install -g aws-cdk@latest....` commands in action files, replaced with only `npm install`.
    - All `cdk *` CLI commands updated to `npx cdk *` in action files.

1. Green checks (**except** for the text extractor unit test.. this will be a follow-up!)
    - Validate Django unit tests, linting of Python, JS, TS, Go all running as expected.
2. Observe that the `Deploy Experimental` action now only deploys the CDK experimental site.
3. Deploy this to prod. Observe that the `Deploy` action now only deploys the CDK site.


PhilR8 (approved):
LGTM!

PR #1602, opened 2025-03-20
EREGCSC-2947 Pull credentials at runtime instead of deployment time

cgodwin1:
Resolves #2947

**Description-**

We have been storing certain credentials as environment variables in the deployed environment, mimicking our local development environment. We need to change this, so we are choosing to pull credentials from AWS Secret Manager at runtime and to only store the ID of each secret, not their contents, in our environment variables.

Local deploy will not be affected by this change at this time.

- Added 2 new `auth` config types, `basic-secretsmanager` and `basic-secretsmanager-env`.
    - See the [text extractor readme](https://github.com/Enterprise-CMCS/cmcs-eregulations/tree/2947-pull-credentials-at-runtime/solution/text-extractor#configuring-authentication) for more info.
- The deployed instance of eRegs uses `basic-secretsmanager-env` auth type when calling the text extractor. (Local is unchanged.)
- The Text Extractor CDK stack no longer imports credentials directly as environment variables, only IDs of secrets.
- Add a "parser launcher" Python Lambda function that retrieves credentials from AWS secrets manager and invokes the eCFR and FR parsers on a cron schedule. This replaces the previous independent cron schedules that the eCFR and FR parsers already had.
- eCFR and FR parsers pull authentication info from the Lambda event object if they are invoked as a Lambda function. Behavior is unchanged locally.
- The eCFR & FR parser CDK stacks no longer import credentials directly as environment variables, only IDs of secrets.
- eRegs backend contains `secret_manager` module that provides several helper functions for accessing secrets via AWS, and falling back to environment variables for local instances.
    - Provides `get_username()`, `get_password()`, `get_secret_field()`, and `get_secret()`.
    - For typical use, only the first two functions should be directly needed.
- The eRegs site stack no longer imports credentials directly as environment variables, only IDs of secrets.
- The Cypress action now pulls credentials from Secrets Manager, not SSM.
- Unrelated quick fix: Pin Magika version to 0.5.1 (for now) to eliminate unit test failures.
- Unrelated quick fix: Use lowercase for log level in database construct -> pgaudit params.

1. Verify all green check marks, including the text extractor unit tests.
2. Log into the AWS Console and go to the Lambda page. Search for 1602 and open each `*-eph-1602-*` Lambda function.
3. Go to "Configuration", "Environment Variables", and browse each one for instances of stored secrets. There will be none.
4. Go to the CloudFormation page, open each `*-eph-1602-*` stack, and browse the template. You will see zero instances of secrets.
5. Validate the text extractor works:
    a. Log into Django Admin on ephemeral site.
    b. Upload a test doc to "Internal Files". (Best is a simple "txt" file.)
    c. Click "Save". The notice stating that text extraction is requested should show.
    d. Filter by "Index Populated = yes", and verify that the new resource shows here after a few moments.
6. Validate the eCFR & FR parsers work:
    a. In the AWS Console, go to Lambda functions, search for "1602".
    b. Click on the "parser-launcher" Lambda.
    c. Click the button to create a "Test" event, and confirm the event (no changes needed to the JSON here.)
    d. Verify the run succeeds, then go to the Lambda functions for the eCFR & FR parsers for 1602.
    e. For each Lambda, click "Monitor" -> "View CloudWatch logs" to go to to the CloudWatch page.
    f. Verify a new log entry starting at the time you clicked "Test" shows up.
    e. Click the log entry and verify there are no errors listed.
7. Verify the site still works for you locally after switching to this branch.


PhilR8 (approved):
LGTM!

PR #1608, opened 2025-03-26
EREGCSC-2909 and EREGCSC-2010 -- Search sorting by new/old

PhilR8:
Resolves [EREGCSC-2909](https://jiraent.cms.gov/browse/EREGCSC-2909) and [EREGCSC-2910](https://jiraent.cms.gov/browse/EREGCSC-2910)

**Description**

Adds ability to sort search results on `/search` page by Newest or Oldest.

This Pull Request satisfies both EREGCSC-2909 (backend work) and EREGCSC-2910 (frontend work).

- Creates `<Sort /> dropdown Vue component that updates `route` with `sort` param when changed
   - `sort=-date` to sort newest first
   - `sort=date` to sort oldest first
   - no `sort` param to sort by relevance
- Styles dropdown based on [Figma comp](https://www.figma.com/design/QTeJIYHaipHux8tx88A4jh/Search-Everything?node-id=1-49916&p=f&t=s6WNa00OSZDxAdHl-0)
   - Makes some assumptions about narrow/mobile widths, but ensures that Sort dropdown responsively renders gracefully
- Disabled Newest and Oldest sort options in Dropdown if only Regulations document type is selected
- Updates filter side effect logic so that sort method is not reset when user changes subject, category, or search term
- Adds `sort` param to API calls when requesting data from back end
- Updates `content_search` Django view to receive `sort` param from API call
- Updates `content_search` `ContentIndexQuerySet` model to update search method to use `sort` param to sort results by date
   - Takes care to sort dated resources first, then add undated resources, and finally add regulations no matter the sorting direction
- Creates migration to add `date` field to all resources on which the sort will depend
- Adds python unit tests and Cypress end to end tests to validate that AC criteria has been met

1. Green check marks
2. Visit [ephemeral deployment](https://crk4wdi7o5.execute-api.us-east-1.amazonaws.com/eph-1608)
3. Go to Search page and perform as many searches as you'd like to validate AC from 2909 and 2910.  Namely:
   - Dated resources should always be at the top of the results list, sorted by selected sort type
   - Undated resources should be listed after sorted dated resources
   - Regulations results should always be listed last no matter the sort direction
   - Results sorted by Relevance should be unchanged compared to `prod`
4. Select Regulations checkbox only.  Note that sort method will be reset and Newest/Oldest sort methods will be disabled in the Sort dropdwn.
5. Perform a search, choose a sort method, and then choose a category or subject.  Note that the sort method will NOT be reset: it will remain.
6. Perform a search, choose a sort method, and then perform a new search.  Note that the sort method will NOT be reset: it will remain.
7. If there are no results, the sort method will be reset and the sort dropdown will not be visible.


solution/ui/regulations/eregs-component-lib/jsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
```
PhilR8, line 2: These .jsconfig files have been added for better LSP performance -- they allow editor LSP integrations know where the aliases reference

cgodwin1 (approved):
LGTM!

PR #1622, opened 2025-04-08
EREGCSC-2910-C Fix data migration using multiple atomic transactions

cgodwin1:
Resolves #2910

**Description-**

Previous migration still failed even with the local statement timeout set to 10 mins. Using a different approach to fix.

- Created a separate non-atomic data migration that operates in blocks of 100 ContentIndex objects per transaction, looping until all dates have been populated properly. Each block of 100 should update in well under the 10 second statement limit.
- Also updated the `backup_db.sh` and `restore_db.sh` scripts to use AWS Secrets Manager as well as environment variables, and automatically retrieve all relevant info.
- Also removed artificial limitation that causes the restore DB script to exit early if it detects a prod database. Opting instead to confirm with the user before proceeding. Since the script already takes a backup of the DB before restoring, this is safe. (There are valid uses for restoring the prod DB from a backup file, we have had to do it a couple of times already.)

1. This PR is green. Double check the data migration step of the deployment.
2. Approve and deploy.
3. Dev, val, and prod all show successful migrations.


PhilR8 (approved):
LGTM!

PR #1625, opened 2025-04-10
EREGCSC-2880-B Add boto3 install instructions for MacOS users to use resource removal scripts

cgodwin1:
Resolves #2880

**Description-**

If Python is installed via Homebrew, running `pip3 install boto3` will fail and suggest users to use Homebrew to install boto, which is not possible. Instead the user should create a virtualenv.

- Add instructions on creating a virtualenv for boto3 for MacOS users.
- Unrelated fix: text extractor tests were failing due to outdated Ubuntu repos, added `apt-get update` to fix.

n/a

### DIFF
--- a/.github/workflows/test-text-extractor.yml
+++ b/.github/workflows/test-text-extractor.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Install dependencies
         if: success()
         run: |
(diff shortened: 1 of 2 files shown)


PhilR8 (approved):
LGTM!

PR #1639, opened 2025-04-24
EREGCSC-2810 -- Adjust file type indicator for internal files

PhilR8:
Resolves [EREGCSC-2810](https://jiraent.cms.gov/browse/EREGCSC-2810)

**Description**

Policy people find it helpful to know what kind of file something is before they click it. We already have a "download x" indicator for some internal files, but we want to make it more consistent around the site. 

- Refactors `getFileNameSuffix` utility method
   - Allows PDF file types to be displayed
   - Labels MSG file types as Outlook file types via an included dictionary
- Refactors `getFileTypeBotton` utility method
   - Removes "Download" from `span`
- Updates SCSS stylesheets to ensure that file type chip does not wrap to another line by itself; it now wraps with the final word in preceding string
- Updates and extends Vitest unit tests
- Adds Cypress end to end tests
- Adds file type chip to internal files in the Reader View sidebar hidden by the "Show More" button (fixes `prod` bug)
   - Adds Cypress end to end test to assert that this is fixed

**This pull request does NOT change:**

- File types are not taken from Magika format information, since [EREGCSC-2722](https://jiraent.cms.gov/browse/EREGCSC-2722) has yet to be implemented
   - File type info is determined the same as before: by the file extension at the end of the file name in the `file_name` field

1. Green check marks
2. Visit [ephemeral deployment](https://qrjlx97zja.execute-api.us-east-1.amazonaws.com/eph-1639) and log in
3. Perform a search and filter by Internal Resources
4. Ensure that the filetype chips no longer say "Download" and the MSG filetypes now say "Outlook"
5. Narrow your browser window to ensure that filetype chips do not wrap by themselves to the next line -- they should wrap with with final word of the string
6. Follow a "Related Citations" link to the Reader View to ensure all of the above situations also occur for internal documents listed in the right sidebar ([example](https://qrjlx97zja.execute-api.us-east-1.amazonaws.com/eph-1639/42/440/Subpart-A/2025-01-01/#440-60))
   - Make sure to click "Show More" to ensure that hidden documents also have the file type chip


cgodwin1 (approved):
LGTM

PR #1658, opened 2025-05-06
EREGCSC-2828 Fix for "Lists of CFR citations are out of order" bug

cgodwin1:
Resolves #2828

**Description-**

A long-standing bug is that citations in lists (admin panel, API responses, etc.) are not in the correct order when referring to generalized citation objects (i.e. not specifically subpart or section objects).

This is because sorting is defined as `["title", "part", "section_id"]` for the section model and `["title", "part", "subpart_id"]` for the subpart model. As a consequence of this inheritance structure, the abstract citation model has no defined order and the assumption that the child models being sorted would trickle up to the parent model was incorrect.

- Add a `child_id` CharField on the AbstractCitation model which will never be blank or null.
- On subpart save, assign `child_id` to `subpart_id`.
- On section save, assign `child_id` to `section_id` cast as a string.
- In the section case, the output is padded with zeros up to 12 characters in order to normalize sections, and prevent a case where e.g. section 1015 comes before section 110.
    - (Why 12? It's arbitrary, the max size of the subpart field has always been 12 so that's the number I chose.)
- Non-atomic migration with atomic batches of 100 populates `child_id` for all existing citations.

1. Go to [the subjects page](https://f3epo0q880.execute-api.us-east-1.amazonaws.com/eph-1658/subjects/?subjects=240).
2. Click "Show Related Citations" on the first result and verify the citations are in the correct numeric ascending order (title, part, then section).
3. Log into [the admin panel](https://f3epo0q880.execute-api.us-east-1.amazonaws.com/eph-1658/admin).
4. Go to an existing Resource or add a new one.
5. Scroll through citations in the "related citations" box, and see that they are also in the correct numeric ascending order.


PhilR8 (approved):
LGTM!

PR #1664, opened 2025-05-09
EREGCSC-2247 Enable providing resource statute citation info to frontend

cgodwin1:
Resolves #2247

**Description-**

In order to enable related (non-CFR) citations to show on the frontend, we need to provide that in the API response wherever resources are serialized. In addition, the citations need to be sorted alphabetically.

- Add dependency on `natsort` to provide the `natsorted()` function which replaces Python's built-in `sorted()` function.
    - This is needed because alphabetical sorting is incorrect when inconsistent-length numbers are in the strings. For example, `sorted(["1", "2", "3", "10"])` would become `["1", "10", "2", "3"]`. Natsort resolves this.
- Add natural sorting (and whitespace removal) to the pre-save hook for custom reference fields (+unit test).
- Add Resources app migration to re-save all resources to utilize the new pre-save hook behavior.
- Add serializers for USC and Act citations.
- Use new serializers in the Resource serializer to enable these citations to show in `/v3/resources` and `/v3/content-search`.
- Unrelated quick fix for resource category prefetching: add prefetching for prefetched subcategory parents.

1. Verify green checks (deploy & migrate succeeded.)
2. Check `/v3/resources` API endpoint and search for "usc_citations" and "act_citations", see that the response includes these fields and some items have one or both of these fields populated (and for those that do, that they are in alphabetical order).
3. Do the same as step 2 for `/v3/content-search/?q=test`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,12 @@ You can run `make` to learn about other available tasks. For example:
 
 `make local.clean` will remove the local environment completely, useful when you want to start fresh.
 
+You may also need [Docker compose commands](https://docs.docker.com/compose/). For example:
+
+If you add or update Python dependencies, you may need to rebuild the site: `docker compose up -d --build regulations`
+
+In general, `make local` will run any migrations that need to run, but if you need to run migrations on their own: `docker compose exec regulations python manage.py migrate`
+
 ## Run tests
 
 #### Testing setup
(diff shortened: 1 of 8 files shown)


PhilR8 (approved):
LGTM!

PR #1669, opened 2025-05-13
[bugfix] Fix bug on category admin list view: invalid prefetch

cgodwin1:
Resolves #n/a

**Description-**

A 500 error is occurring on the public and internal subcategory admin panel list views. It's because we're prefetching parent categories for the list display, but we were specifying the models to prefetch as `InternalCategory` and `PublicCategory` instead of the abstract class.

- `InternalCategory.objects.all()` and `PublicCategory.objects.all()` in prefetch calls replaced with `AbstractCategory.objects.select_subclasses()`.

1. Go to the admin panel and click on public subcategory, make sure the page loads and parent categories are displaying properly.
2. Go to internal subcategory, make sure the page loads and parent categories are displaying properly.

### DIFF
--- a/solution/backend/resources/admin/categories.py
+++ b/solution/backend/resources/admin/categories.py
@@ -3,6 +3,7 @@ from django.db.models import Prefetch
 
 from common.admin import CustomAdminMixin
 from resources.models import (
+    AbstractCategory,
     InternalCategory,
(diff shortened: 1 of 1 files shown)


PhilR8 (approved):
LGTM!

PR #1695, opened 2025-06-05
EREGCSC-2976 Fix vulnerability in redirect lambda

cgodwin1:
Remove ability to redirect to non-root paths which introduced an open redirect vulnerability. Now the lambda will _always_ redirect to the homepage. This will force users to update bookmarks, but the pilot domain has been deprecated for some time already.

Resolves #

**Description-**

- expected change 1


PhilR8 (approved):
LGTM!

PR #1715, opened 2025-06-23
EREGCSC-2974-B Fix extractor queue and ensure polite delay 

cgodwin1:
Resolves #2974

**Important**

This PR depends on #1717 to be deployed all the way to prod before deploying this. Merging this before #1717 will fail because we can't convert the text extractor's SQS queue from standard to FIFO while the deployed site depends on it.

**Description-**

We need to re-extract a large quantity of documents, but we don't want to overload the websites we work with. Let's ensure we delay between requests.

Additionally, while working on this we identified that the queue wasn't working correctly. We reduced the number of concurrent executions to 1 and saw that, for example, if 10 messages are sent to the queue, one will be processed by the lambda and the remaining 9 will be marked "in flight", and made unavailable for at least 15 minutes.

This is because we were trying to go below a limit of 10 concurrent executions on a standard queue, and a FIFO queue is really what is needed for that purpose. Additionally, using a FIFO queue allows us to group messages to allow parallel processing of sequentially processed subgroups of messages.

- Limit max concurrent extractions to 1 per server, with a configurable minimum delay between each download request.
    - This will drastically increase the amount of time it takes to run a batch extraction of many docs from the same source, but should not make much difference in average day-to-day operations.
- Add retries managed by SQS by forcing extractor to raise exception on failure instead of returning a Lambda response object.
- Remove custom retries on file download behavior, rely on SQS retries.
- Add `retrieval_delay` parameter to extractor request config.
- Add "Extraction Delay Time" field to Resources Configuration (default 180 seconds, managed in admin panel).
- Pass delay field to text extractor via config object in request. This is currently for web backend only.
- Switch text extractor SQS queue from standard to FIFO.
    - This guarantees sequential processing.
    - By taking advantage of message group IDs, we can enable parallel processing of multiple resources from different domains but enforce sequential processing for resources from the same domain.
    - This relies on extracting the domain name from each resource URL and assigning that as the message group ID.
- Increase Lambda memory to 1,769MB. This assigns exactly 1 vCPU and should permit the extractor to run at max speed.
- Split Dockerfile into local and deployed version.
    - This is needed because the lambda-proxy software we use locally causes an init timeout in the deployed instance.
    - This costs us valuable time and money, so let's just remove it since it's not needed in the deployed version anyway.

To validate new behavior:

1. Go to the SQS queue for this PR in AWS.
2. In another tab, go to the admin panel and select a small batch of public links from the same domain (e.g. `medicaid.gov`), then go to the action dropdown, "Extract text...", submit.
4. Refresh the SQS page and see all the messages in the queue.
5. Observe that the queue decreases slowly, at roughly 3 minute intervals. You can also view this in CloudWatch -> log group for 1715 text extractor -> search for "Waiting for" and see the wait info message.
6. Repeat steps 1-5 but select a batch of links that includes some from a different domain (e.g. 2 from `medicaid.gov` and 2 from `gao.gov`).
7. Observe that the resources from separate domains process at the same time, but within each domain group, they are processed sequentially with a 3 minute delay between them.

To validate admin panel changes:

1. Go to admin panel -> resources configuration.
2. Verify "Extraction Delay Time" field exists and is populated with 180 seconds.


PhilR8 (approved):
LGTM!

PR #1717, opened 2025-06-24
EREGCSC-2974-A Temporarily disable the text extractor

cgodwin1:
Resolves part of #2974

**Description-**

In order to convert the text extractor's SQS queue from standard to FIFO, we can't have the eRegs site depending on the queue via exported URL or ARN. This is because in order to do the switch, AWS must delete the old queue and create a new one in its place. Attempting to do this currently will fail as the deploy process won't allow the deletion of a resource that is depended on by a downstream stack.

- Temporarily disable the text extractor by removing imports in the CDK API stack.
- Modify the text extractor calling code to properly display an admin panel error in the case where the text extractor's destination URL or ARN is not configured.

1. Go to the admin panel -> public links.
2. Select one or more resources and use the action dropdown to attempt to extract text.
3. Observe the red error banner at the top of the page that results. This is expected until the next 2974 PR is deployed.

### DIFF
--- a/cdk-eregs/lib/stacks/api-stack.ts
+++ b/cdk-eregs/lib/stacks/api-stack.ts
@@ -209,10 +209,10 @@ export class BackendStack extends cdk.Stack {
         // ================================
         // SQS QUEUE
         // ================================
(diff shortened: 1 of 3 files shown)


PhilR8 (approved):
LGTM!

PR #1721, opened 2025-06-30
EREGCSC-2625 Omit text from scraped HTML that is irrelevant to search

cgodwin1:
Resolves #2625

**Description-**

Currently, the markup extractor will extract all visible text on a webpage. This includes text such as nav bars, headers, and footers. This reduces the effectiveness of full-text search if a query is only present in one of these non-content parts of the page.

Unfortunately, not all sites use tags such as `<main>`, `<section>`, and `<article>` in the appropriate way, and many older sites don't make use of that at all. So we want a way to add rule exceptions on a per-site basis while providing a sane default set of behaviors to fall back on.

- The default behavior will now be to look for content in a `<main>` tag.
- If that doesn't exist, it will attempt to find content in one or more `<article>` and/or `<section>` tags.
- If they don't exist, then it will fall-back to using the entire `<body>`.
- The following tags are removed entirely: `<script>`, `<style>`, `<nav>`, `<header>`, `<footer>`, `<aside>`, `<form>`, and `<noscript>`.
- Custom rules are defined by method name. For example, to write custom rules for "test.example.com":
    ```python
    def _extract_from_test_example_com(self, soup: BeautifulSoup) -> str:
        return soup.do_something_here()
    ```
    (Note that a top-level exception such as "_extract_from_example_com()" will automatically apply to both "example.com" and "www.example.com".)

1. Check out the `test_markup.py` changes. The fixture files there cover all currently implemented rules and exceptions.
2. Re-extract some of the links found in these searches:
    - Search for CCTAG
    - Search for 'training & education'
3. After re-extracting, reload those pages on the experimental deploy and see that the results are more relevant, i.e. are not showing up because the query text is in a header/footer.


PhilR8 (approved):
LGTM!

PR #1761, opened 2025-07-28
Make "extract_url" field visible and editable in the admin panel

cgodwin1:
Resolves #n/a

**Description-**

We have had a field on the resource model, `extract_url`, for a long time. It's purpose so far has been limited to Federal Register links where the user-visible page doesn't extract as well as a non-user-friendly link. However, we can apply this to other resources as well. For example: show a user a friendly web page but index a large PDF that is linked to from that page.

- `extract_url` visible and editable on the admin panel.
- If `extract_url` field changes, it will trigger extraction if auto extract is enabled.

1. Go to a public link and expand the "Search indexing" fieldset.
2. Enter a new URL in to the "Text extraction URL" field and save.
3. After some time, the indexed text should change to reflect the new URL.

### DIFF
--- a/solution/backend/resources/admin/internal_resources.py
+++ b/solution/backend/resources/admin/internal_resources.py
@@ -45,7 +45,7 @@ class InternalLinkAdmin(AbstractInternalResourceAdmin):
         }),
         ("Search indexing", {
             "classes": ("collapse",),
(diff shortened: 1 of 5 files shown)


PhilR8 (approved):
LGTM!

PR #1784, opened 2025-08-14
EREGCSC-2992 — Upgrade Node to latest LTS

PhilR8:
Resolves [EREGCSC-2992](https://jiraent.cms.gov/browse/EREGCSC-2992)

**Description**

eRegs is using Node 20 wherever Node is required.  Node 20 is currently in "Maintenance LTS (Long Term Support)" mode and will be [supported through May 2026](https://nodejs.org/en/about/previous-releases).  We should upgrade to a newer version of Node before that time.

- Updates node versions to the latest current LTS version, which is Node 22.  This will be supported through May 2027.

1. Green check marks
2. All team members install and use Node 22 LTS ([Jod](https://nodejs.org/en/blog/release/v22.11.0)) locally
   - If using Node Version Manager ([nvm](https://github.com/nvm-sh/nvm)), this command will install and use Node 22: `nvm install 22`
   - User may also desire to set nvm's default version to 22: `nvm alias default 22`
4. No issues with local development, Github Actions runs, or deployments to hosted environments using Node 22

### DIFF
--- a/.github/workflows/deploy-to-env.yml
+++ b/.github/workflows/deploy-to-env.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
 
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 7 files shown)


cgodwin1 (approved):
LGTM!

PR #1785, opened 2025-08-14
EREGCSC-2998 — Fix glob/inflight deprecation warning

PhilR8:
Resolves [EREGCSC-2998](https://jiraent.cms.gov/browse/EREGCSC-2998)

**Description**

When we deploy the static assets, the following warning is thrown:

> npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
> npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported

This warning is being caused by `jest`, which depends on `glob@7.2.3`.  

Snyk also shows that `shelljs` in `eregs-component-lib` depends on `glob@7.2.3` and contains the same issue.

Upgrading `shelljs` to the latest version (0.10.0) fixes the issue in `eregs-component-lib`.

Removing `jest` from the `cdk-eregs` project silences the warning seen in the [`deploy-static-assets` step](https://github.com/Enterprise-CMCS/cmcs-eregulations/actions/runs/16970109448/job/48104480452) of the `deploy-experimental` Github action.

- Updates [shelljs](https://github.com/shelljs/shelljs) to latest version (0.10.0)
- Removes `jest` and all related dependencies from `cdk-eregs` because:
   1. Jest depends on the same version of `glob` that is causing the warning
   2. Jest has been throwing this warning for [over a year without a fix](https://github.com/jestjs/jest/issues/15173)
   3. eRegs isn't even using Jest because there are no tests in `cdk-eregs` or the entire cdk project
   4. If/when it is decided to add tests to `cdk-eregs`, Vitest would be used instead of Jest because there are already extensive Vitest test suites running in other parts of the project; Vitest would be preferred here in `cdk-eregs` for the sake of consistency and maintainability
- Removes `--debug` flag in vite `watch` command to make local development less noisy in terminal

1. Green check marks
6. Local development builds and reloads as expected
7. Deprecation warning is gone when deploying static assets


cgodwin1 (approved):
LGTM!

PR #1527, opened 2025-01-24
EREGCSC-2905-B -- Add list of directories for npm ecosystem package.json locations in Dependabot config

PhilR8:
Resolves [EREGCSC-2905](https://jiraent.cms.gov/browse/EREGCSC-2905)

**Description**

Further refinement of Dependabot configuration work.

Since we began updating the `dependabot.yml` file, we've been seeing some [failures of the Dependabot Github Action](https://github.com/Enterprise-CMCS/cmcs-eregulations/actions/workflows/dependabot/dependabot-updates).  The [error](https://github.com/Enterprise-CMCS/cmcs-eregulations/network/updates/952232478) is:

```
Dependabot couldn't find a package.json
Dependabot requires a package.json to evaluate your JavaScript dependencies. It had expected to find one at the path: /package.json.
```

To attempt to fix this error, `dependabot.yml` is being updated to use the `directories` option to specify locations of multiple `package.json` files.

See:
- [Controlling which dependencies are updated by Dependabot / Defining multiple locations for manifest files](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files) (Github Docs)
- [Existing `dependabot.yml` example with multiple `directories`](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/.github/dependabot.yml)
- [Long Github Issues discussion about `groups` and `directories` options](https://github.com/dependabot/dependabot-core/issues/2178)

**This pull request changes:**

- Changes `directory: "/"` to `directories: <list/of/directories>` for `npm` package ecosystem
- Updates Cypress test in `part` test suite to mock async call to `/v3/title/{title}/part/{part}/history/section/{section}` endpoint
   - Adds related fixture file

**Steps to manually verify this change:**

1. Green check marks
2. Dependabot action does not fail

PR #1523, opened 2025-01-21
EREGCSC-2905-A -- Increment open Pull Request limit to 2

PhilR8:
Resolves [EREGCSC-2905](https://jiraent.cms.gov/browse/EREGCSC-2905)

**Description**

Further refinement of Dependabot configuration work.  The expectation was that each defined `package-ecosystem` would have a maximum of one PR open; with two `package-ecosystems` defined, the expected number of PRs open at any given time would be a maximum of two.

Instead, we're only seeing one pull request.  We'd like to tweak this setting to get two PRs open concurrently.

**This pull request changes:**

- updates `open-pull-request-limit` to `2` for each package ecosystem
- Adds `groups` option to group certain dependencies together into a single pull request

**Steps to manually verify this change:**

1. Green check marks
2. Merge to `main` and see if a second dependabot PR joins the currently open one

PR #1517, opened 2025-01-15
EREGCSC-2905 -- Automate prompts for major updates for key components

PhilR8:
Resolves [EREGCSC-2905](https://jiraent.cms.gov/browse/EREGCSC-2905)

**Description**

Make Dependabot a bit more aggressive when it comes to upgrading our `npm` dependencies.  

To do this, we will allow dependabot to create pull requests every weekday, but will limit the number of open pull requests to one.  This way, we can move as fast as we want with these PRs -- we can have a new one every day if needed, or we can allow an open PR to sit for several days before being reviewed without worrying about new PRs being created.

**This pull request changes:**

- Adds section to `dependabot.yml` for `npm` ecosystem
- Sets pull request creation interval to `daily` (weekdays)
- Limits open PRs for `npm` ecosystem to `1` (one)
- Increase minimum required version in `package.json` to the recently updated version

**Steps to manually verify this change:**

1. Green check marks
2. `npm` ecosystem pull requests are created more often than we're used to seeing them

PR #1822, opened 2025-09-11
EREGCSC-3008 — Pin all npm dependencies

PhilR8:
Resolves [EREGCSC-3008](https://jiraent.cms.gov/browse/EREGCSC-3008)

**Description**

To improve consistency in deploys and minimize risk of bringing in uninspected NPM dependencies, we want to ensure we're pinning patch versions.

**Background**

NPM allows for [several types of versioning for dependencies listed in our `package.json` file](https://thetshaped.dev/p/pin-your-dependencies-in-packagejson).  Some types that we're currently using in eRegs:
   - prefixing a version number with a tilde (ex: `~2.1.3`) will automatically install the newest patch version of `2.1`
   - prefixing a version number with a caret (ex: `^2.1.3`) will automatically install the newest minor version of `2`)
   - prefixing a version number with a greater than/equals operator (ex: `>=2.1.3`) will automatically install the newest major version of the library that is `2` or greater

These are known as [semver (semantic versioning) ranges](https://docs.npmjs.com/cli/v6/using-npm/semver#ranges).

The ability to use these semver ranges to install newer versions of dependencies than what is explicitly listed has some benefits: chief among them is that it [allows the project to "automatically" get updated releases, which may even include security fixes](https://docs.renovatebot.com/dependency-pinning/#why-use-ranges).

However, the inverse can be also be true: it can allow compromised packages to be installed without explicit knowledge of the team.  To wit: a [recent incident occured where 18 popular npm packages were compromised](https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised).

None of our libraries or their many dependencies were affected by the aforementioned incident, but the incident has led to conversations among the eRegs team regarding our own dependency pinning habits.  Our current pattern of creating ephemeral deployments for all pull requests, as well as being able to perform daily zero downtime deployments, means we are deploying early and often.  Is the benefit of "automatically" getting updated packages worth the risk of accidentally and unknowingly shipping a compromised package to `prod`?

The answer is no, mainly because we have [spent](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1517) [much](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1523) [time](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1527) [and](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1533) [effort](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1755) [this past year](https://github.com/Enterprise-CMCS/cmcs-eregulations/pull/1773) configuring and allowing dependabot to create pull requests that manage and update npm packages on an aggressive yet realistic pace.  

That is to say: we can expect Github's dependabot and code scanning functionality to create daily pull requests to keep the project's npm dependencies up to date and secure.  Along with periodic explicit audits of our dependencies, we can pin every npm dependency to an exact version without fear that the project will be missing out on security updates found in newer package versions.  That is what this pull request does.

- Removes all caret and tilde semver ranges from all `package.json` files
- Updates associated `package-lock.json` files

1. Green check marks
2. Visit ephemeral deployment and validate that site works as expected

### DIFF
--- a/cdk-eregs/package.json
+++ b/cdk-eregs/package.json
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-ssm": "3.888.0",
-    "aws-cdk-lib": "~.214.0",
+    "aws-cdk-lib": "2.214.0",
     "constructs": "10.4.1",
     "fs-extra": "11.3.1",
     "path": "0.12.7",


cgodwin1 (approved):
LGTM

PR #2240, opened 2026-09-04
EREGCSC-3256 Update dependency chain including Starlette to address issue

cgodwin1:
Resolves [EREGCSC-3256](https://jiraent.cms.gov/browse/EREGCSC-3256)

**Description:**

FastMCP has a dependency on Starlette which needs to be updated. This PR updates FastMCP to the latest stable version which is 4.0.2. Also checked existing MCP server code to ensure nothing breaks from v3 -> v4.

- Update FastMCP to 4.0.2 (which updates Starlette to 1.6.0).

1. Green check marks


PhilR8 (approved):
LGTM!

